PR #21, opened 2016-06-08
Fix logic such that -DHIP_USE_SHARED_LIBRARY is honored

kknox:
The old code clobbered -DHIP_USE_SHARED_LIBRARY with environment variable, forcing users to use environment variable which can be difficult with cmake ExternalProject_Add().

My opinion:
HIP should honor the cmake built-in 'BUILD_SHARED_LIBS' setting instead of rolling it's own solution.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,10 +80,12 @@ else()
 endif()
 
 # Set if we need to build shared or static library
-if(NOT DEFINED ENV{HIP_USE_SHARED_LIBRARY})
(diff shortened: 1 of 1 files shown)


mangupta:
HIP on the developer-preview branch now uses the flag -DHIP_LIB_TYPE to control library type. Valid values are 0 (sea-of-objects), 1 (static library) and 2 (dynamic library). The flag does not have to be an environment variable and should solve the problem you were facing.

PR #22, opened 2016-06-09
HIP build clobbers user specified CMAKE_CXX_FLAGS

kknox:
Users should be able to pass in their own compilation flags into
hip cmake project through -DCMAKE_CXX_FLAGS and -DCMAKE_C_FLAGS.  The
root cmakefile unconditionally clobbers user values.  Instead,
hip build should append implementation flags.


mangupta:
HIP build no longer clobbers CMAKE_CXX_FLAGS and CMAKE_C_FLAGS. See https://github.com/GPUOpen-ProfessionalCompute-Tools/HIP/blob/master/CMakeLists.txt#L131-L132

PR #41, opened 2016-10-02
__mul24 & __umul24 integer intrisic support

ekondis:
Please consider adding the 24bit integer multiplication functions which are more efficient than the equivalent 32bit operations on AMD platforms.


mangupta:
Thanks! Merged into internal AMD repo.

PR #44, opened 2016-10-13
Updated hip_blas header of hcc path

NEELMCW:
This commit incorporates few additional APIs and prototype changes on HCC path of hip_blas headers.

Inorder to avoid conflict with the installed header's via hcblas, I am attempting to update the hip_blas hosted under hip as well


mangupta:
No longer required to merge this since hip_blas headers have moved to the [hcblas repo](https://bitbucket.org/multicoreware/hcblas).

PR #49, opened 2016-11-18
Make nvcc_detail/hip_runtime_api.h C compliant

NEELMCW:
This commit adds #ifdef control statements to make the header C compliant
In C++, member type declaration and setting default parameters to function arguments are possible
which may not be case with C. So appropriate #ifdefs and else are added to accomodate C compliance


NEELMCW:
The indentation of the code might be slightly off. I would like to know the coding guidelines that HIP team is following to ensure uniformity. I happen to see tabs in certain places and spaces in other.

bensander:
Hi Neel – coding convention specifies spaces.  Sounds like we need to do a little housecleaning.
-ben

adityaatluri:
Hi @NEELMCW ,
Can you test developer preview branch and create pull request based off it?

PR #56, opened 2016-12-16
Use proper __fp16 intrinsic type for implementing half

AlexVlx:
This work concerns itself primarily with leveraging the proper FP16 support that is now available in HCC and in the back-end. More specifically:
      - it moves from using the quirky struct with an embedded 16-bit bitfield as a proxy for the type half to using the __fp16 intrinsic type;
      - for compatibility with CUDA apps a half POD-type is provided, but with an embedded __fp16 member as opposed to an integral one;
      - all of the arithmetic, comparison and conversion functions (the __h* family) have been refactored to use the native operators of the __fp16 type; they have also been moved into the hcc_detail/hip_fp16.h header as inline functions;
      - hdiv and h2div have been added (they were missing);
      - __hneg and __hneg2 have been correctly named (they were __hneQ before);
At the same time, __ldg has been refactored into a single function template - it made no sense to have numerous overloads that merely dereferenced a pointer. If in the future we will have some special functionality available for certain types we can provide overloads at that point.
Finally, an interesting example of why naked usage of operator new is considered harmful has been corrected in one of the unit tests. All the changes are unit-test neutral i.e. they neither increase nor diminish failure rate.
Please review and merge as you see fit. Thank you.


adityaatluri:
Hi @AlexVlx ,
Thank you for the huge change.
Can't merge it as,
1. It should be C compliant
2. This doesn't work with HCC 3.5
3. The LC is broken for half ops. It is generating conversion to f32.
4. FMA is not being compiled by clang inside HCC 4.0.
5. There is no guards for checking Hawaii and Tonga cards. 
6. Things get complicated with Half2. Can do SDWA on <2xhalf> without typecasting which is perf drop. 
7. There are lots of limitations from LC inorder to implement half2.

Half support is of utmost important and should have the best performance the hardware can support.
I am trying out optimized implementation for half2 and half support.

Here is my scratchpad for it: https://github.com/gpu0/amdgpu-code

We have an internal discussion going on for half support. I'll add you to that. 

I am leaving this open.

AlexVlx:
Hello Adityaa.

Thank you for clarifying. A few notes:
1. Right, for some reason I was unaware of that requirement...we should discuss this, but it might be opportune to provide a .hpp header for people who are C++ inclined, which could contain the C++ bits; I'll update this to include a C version, for future usage.
2. Indeed, and that's my error for not clarifying - this is intended to go along with the ToT upgrade, it won't be backported.
3. I did hear something about that; my understanding (please correct me if I am wrong) is that it'd be a performance bug and not a correctness one (this last bit is handwavingly inaccurate on my part:)) i.e. on the library level we could move to using __fp16 and when LC catches up reap the benefits.
4. Probably better to deal with this one offline.
5. Right, I had ignored Hawaii. However, Tonga is safe, it has FP16 (more or less same IP level as Fiji).
6. In the implementation associated with this PR half2 is handwaved / simplistic; it would be focused upon subsequently as I gather there are some special properties associated with it.
7. see 6:)

I do not disagree with the idea that half is important, on the contrary. Having said that, I would dare suggest that it is pretty important to have it out and about, in a form which allows experimentation. Perhaps we can cover this in the internal thread. Thank you for taking the time to peek. Cheers!

adityaatluri:
Great to have you on the same page. I am researching efficient ways to get the best performance. I am writing everything in LLVM IR inside HIP. Once, LC gets decent support, we can push to ROCDL.

PR #76, opened 2017-05-04
Fix problems with force push

pfultz2:
Someone has done a `git push -f`, which should never happen as it screws up everyone's repositories. This fixes this by merging in the previous master. There is no file changes.


pfultz2:
This problem can easily be reproduced by doing:

```
git clone git@github.com:GPUOpen-ProfessionalCompute-Tools/HIP.git

cd HIP
git reset roc-1.3.0 --hard
git pull origin
```

PR #104, opened 2017-07-18
update samples for vega support

7SK:


mangupta:
closing this since PR #108 fixes by modifying hipcc.

bensander:
Agree the hipcc mod is a better fix.

adityaatluri:
As these samples are used as "the way to be use hip", I think we should merge these changes.

mangupta:
@adityaatluri: I don't think making all HIP samples build fat binaries is the right way. We could change one samples to build as fatbin to demonstrate how to do it.

bensander:
Agree

PR #106, opened 2017-07-18
[ci] Remove repo.radeon.com build-args; HiP CI unit tests running 

kknox:
HiP docker images should inherit the repo information from the hcc image,
so no need to duplicate in HiP again


kknox:
I must have been mistaken, the master branch must already have a fix for the bugs I mentioned previously.  All unit tests are passing.

PR #113, opened 2017-07-26
[cmake] No longer rebuild cmake cache on each build

mangupta:
Cmake cache was being rebuilt on each build. This was being done
to update HIP_VERSION, HCC_VERSION, .hipInfo and .hipVersion.
However, rebuilding cache also re-runs HIT parser which is slow.
Removing the cache rebuild should speed up the build. But user
needs to explicitly rebuild the cache in case HIP_VERSION or
HCC_VERSION changes by calling "make rebuild_cache"


bensander:
Awesome!

PR #127, opened 2017-07-28
Pass kernel name to HCC dispatch_hsa_kernel, for debug/profile

bensander:


src/hip_module.cpp
```diff
@@ -452,10 +452,7 @@ hipError_t ihipModuleLaunchKernel(hipFunction_t f,
 
         lp.av->dispatch_hsa_kernel(&aql, config[1] /* kernarg*/, kernArgSize, 
                                   (startEvent || stopEvent) ? &cf : nullptr
-#define USE_NAMED_KERNEL 0
-#if USE_NAMED_KERNEL
                                   , f->_name.c_str()
```
mangupta: Looks like this needs to be guarded using hcc_workweek since ROCm 1.6.x HCC does not support this.

bensander: Yep, looks like this was post 1.6.  Here's an updated version with WW check - I know we had it by 17300 so was conservative.

PR #128, opened 2017-07-28
[hipcc] Cleanup amdgpu target logic

mangupta:
Existing logic has a bug. If user specifies targetA via commandline
options, while enumerator returns targetB, hipcc will create a fatbin
containing targets targetA and targetB. enumerator should only be used
when no target is specified by user (commandline or env var).

### DIFF
--- a/bin/hipcc
+++ b/bin/hipcc
@@ -75,6 +75,7 @@ $target_gfx801 = 0;
 $target_gfx802 = 0;
 $target_gfx803 = 0;
 $target_gfx900 = 0;
+$default_amdgpu_target = 1;
 
 if ($HIP_PLATFORM eq "hcc") {
     $HSA_PATH=$ENV{'HSA_PATH'} // "/opt/rocm/hsa";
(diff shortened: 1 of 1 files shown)


bensander:
@adityaaatluri can you review?

adityaatluri:
What does `$default_amdgpu_target` do? I think we don't need a default target now. Always use agent_enumerator to set the default target.

sunway513:
We should not set a default value as a fall back, instead we should report error that the either KFD is not properly installed or the enumerator is missing; so users can make it right.

mangupta:
@adityaatluri: Previous logic was blindly using rocm_agent_enumerator. So if a user had a gfx803 card in his build system and was building for gfx900 by specifying --amdgpu-target=gfx900, the previous logic would end up selecting both gfx803 (via enumerator) and gfx900 (via --amdgpu-target). So I refactored the code to avoid this. $default_amdgpu_target determines if we need to use the enumerator. If does not specify any default target in itself.

mangupta:
@sunway513: commit 85ff6e3 removes the fallback in addition to fixing a refactoring bug. hipcc will now die if we could not determine a valid target via --amdgpu-target, HCC_AMDGPU_TARGET and enumerator. However we only need to die in case linking was required (i.e. when GPU code is finalized).

adityaatluri:
@mangupta can you change the name of the variable `$default_amdgpu_target` to make it more intuitive?

PR #134, opened 2017-07-29
fix hipMemcpy2DAsync

gargrahul:
Fixed hipMemcpy2DAsync.


boxerab:
Thanks, will try this.

PR #140, opened 2017-08-03
[HIP Texture] The GPU virtual address for texture memory needs to be

weixingzhang:
aligned.

In hcc_am, a bigger buffer will be allocated for alignment purpose
and _unalignedDevicePointer is added in struct AmPointerInfo for
original allocated address.


mangupta:
@weixingzhang : Can I close this one since you have created a new PR #148 for the same feature?

PR #144, opened 2017-08-05
Update Makefile

MikalaiDrabovich:
Update Makefile for Vega, Pascal

### DIFF
--- a/samples/0_Intro/bit_extract/Makefile
+++ b/samples/0_Intro/bit_extract/Makefile
@@ -8,8 +8,10 @@ HIP_PLATFORM=$(shell $(HIP_PATH)/bin/hipconfig --platform)
 HIPCC=$(HIP_PATH)/bin/hipcc
 
 # Show how to use PLATFORM to specify different options for each compiler:
-ifeq (${HIP_PLATFORM}, nvcc)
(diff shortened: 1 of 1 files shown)


samples/0_Intro/bit_extract/Makefile
```diff
@@ -8,8 +8,10 @@ HIP_PLATFORM=$(shell $(HIP_PATH)/bin/hipconfig --platform)
 HIPCC=$(HIP_PATH)/bin/hipcc
 
 # Show how to use PLATFORM to specify different options for each compiler:
-ifeq (${HIP_PLATFORM}, nvcc)
-	HIPCC_FLAGS = -gencode=arch=compute_20,code=sm_20 
+ifeq (${HIP_PLATFORM}, hcc)
+	HIPCC_FLAGS = --amdgpu-target=gfx900 # this is for Vega, for previous gen - gfx803
```
mangupta: Latest bin/hipcc contains logic to autodetect the current amdgpu device on the build system. So as long as you don't need to cross-compile for another target device, we should no longer need to explicitly pass the ```--amdgpu-target=gfxYYY``` flag.

mangupta:
Closing this pull request without merging it since forcing the GPU target directly in the Makefile is not suitable for the intro HIP samples.

PR #148, opened 2017-08-08
[HIP Texture] The GPU virtual address for texture memory needs to be

weixingzhang:
The GPU virtual address for texture memory needs to be aligned.

In hcc_am, a bigger buffer will be allocated for alignment purpose
and _unalignedDevicePointer is added in struct AmPointerInfo for
original allocated address.


whchung:
@weixingzhang this is becoming urgent ,as HIP build is breaking now with tip of HCC.

bensander:
Maneesh added the version guards, I just merged the combined pr

PR #154, opened 2017-08-11
Thrust workarounds for unsupported HIP data structures and functions

sriharikarnam:


bensander:
Also need to rebase on master branch and resubmit to pick up automated testing issue.

sriharikarnam:
@gargrahul  
There are two enums in CUDA "CUsharedconfig" and "cudaSharedMemConfig",both have equal number of enumerators with same values. Few CUDA APIs accept CUsharedconfig enum and few accept cudaSharedMemConfig enum. As we have only one equivalent enum "hipSharedMemConfig" in HIP we are using and typecasting same enum to invoke CUDA APIs.

PR #155, opened 2017-08-15
Modify hip headers such that they are more compatible with strict C compiler

wsttiger:
Made changes to NVCC headers such that they are more C compiler friendly

Add missing CUDA tags to hipify-perl

### DIFF
--- a/include/hip/nvcc_detail/hip_runtime_api.h
+++ b/include/hip/nvcc_detail/hip_runtime_api.h
@@ -359,7 +359,7 @@ inline static hipError_t hipMemcpy(void* dst, const void* src, size_t sizeBytes,
 
 
 inline static hipError_t hipMemcpyAsync(void* dst, const void* src, size_t sizeBytes, hipMemcpyKind copyKind, hipStream_t stream __dparm(0)) {
-  return hipCUDAErrorTohipError(cudaMemcpyAsync(dst, src, sizeBytes, hipMemcpyKindToCudaMemcpyKind(copyKind), 0));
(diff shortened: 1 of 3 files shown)


include/hip/nvcc_detail/hip_runtime_api.h
```diff
@@ -351,25 +358,24 @@ inline static hipError_t hipMemcpy(void* dst, const void* src, size_t sizeBytes,
 }
 
 
-inline static hipError_t hipMemcpyAsync(void* dst, const void* src, size_t sizeBytes, hipMemcpyKind copyKind, hipStream_t stream=0) {
-  return hipCUDAErrorTohipError(cudaMemcpyAsync(dst, src, sizeBytes, hipMemcpyKindToCudaMemcpyKind(copyKind), stream));
+inline static hipError_t hipMemcpyAsync(void* dst, const void* src, size_t sizeBytes, hipMemcpyKind copyKind, hipStream_t stream __dparm(0)) {
+  return hipCUDAErrorTohipError(cudaMemcpyAsync(dst, src, sizeBytes, hipMemcpyKindToCudaMemcpyKind(copyKind), 0));
```
mangupta: Should still remain ```stream`` and not ```0```. Otherwise we cannot queue a cudaMemcpyAsync to any other stream other than NULL stream.

bensander: @wsttiger - can you make the change Maneesh suggested and also rebase on master to pick up a test env change from kent?

PR #171, opened 2017-08-30
[dtests] Restructuring dtests

phani544:
Restructured directed tests folders.


mangupta:
Discarding since this will be handled afresh in a new PR.

PR #173, opened 2017-08-31
Dtests memcpy

phani544:
Added tests for variants of hipMemcpyDtoH,HtoD, and Async.


mangupta:
Discarding since this will be handled afresh in a new PR.

PR #191, opened 2017-09-26
Fix module

bensander:
Several fixes related to module API and global variables.


bensander:
Redundant with other PR.

PR #195, opened 2017-10-05
[HIPIFY] CUDA RT Textures and Arrays support update

emankov:


hipify-clang/CMakeLists.txt
```diff
@@ -1,14 +1,16 @@
 cmake_minimum_required(VERSION 2.8.8)
 project(hipify-clang)
 
-set(BUILD_HIPIFY_CLANG 0 PARENT_SCOPE)
+if (PARENT_SCOPE)
+    set(BUILD_HIPIFY_CLANG 0 PARENT_SCOPE)
```
ChrisKitching: You could make this a cache variable to avoid having to check `if (PARENT_SCOPE)` all the time.

This will have the added advantage that it'll be available in higher scopes (your current approach will fail if scopes are nested more than two deep, with this at the bottom).

emankov: You've already done it in https://github.com/ROCm-Developer-Tools/HIP/pull/202 :)

PR #197, opened 2017-10-07
Update coherency tests

bensander:


tests/src/runtimeApi/memory/p2p_copy_coherency.cpp
```diff
@@ -163,12 +173,17 @@ void testMultiGpu(int dev0, int dev1, int numElements, bool hostSync, bool useMe
 #endif
 
     printf ("  test: init complete\n");
-    runTest(useMemcpy , hostSync, gpu0Stream, gpu1Stream, numElements, dataGpu0_0,dataGpu0_1, dataGpu1, dataHost, expected);
+    int stepAIsCopy = 0;
```
mangupta: I don't see stepAIsCopy and stepBIsCopy being used anywhere in this function.

tests/src/runtimeApi/memory/p2p_copy_coherency.cpp
```diff
@@ -163,12 +173,17 @@ void testMultiGpu(int dev0, int dev1, int numElements, bool hostSync, bool useMe
...
+    int stepBIsCopy = 1;
+    runTestImpl(true, hostSync, gpu0Stream, gpu1Stream, numElements, dataGpu0_0,dataGpu0_1, dataGpu1, dataHost, expected);
```
mangupta: We seem to be ignoring useMemcpy value that is passed to testMultiGpu. This results in L#211 and L#214 being the same test. Similarly L#212 and L#213 degenerate into the same test.

bensander:
@mangupta - can you re-review?

Issue #94, opened 2017-07-04
 The dependency target "lit" of target "test-hipify" does not exist.


BeamOfLight:
Hello!
I can't build hipify-clang. I have this error "The dependency target "lit" of target "test-hipify" does not exist."  Instruction: https://github.com/ROCm-Developer-Tools/HIP/blob/master/hipify-clang/README.md#build-and-install

Can you help me? Can i install hipify-clang from deb-package?

hipconfig:
> HIP version  : 1.2.17263
> 
> == hipconfig
> HIP_PATH     : /opt/rocm
> HIP_PLATFORM : hcc
> CPP_CONFIG   :  -D__HIP_PLATFORM_HCC__=   -I/opt/rocm/include -I/opt/rocm/hcc/include
> 
> == hcc
> HSA_PATH     : /opt/rocm/hsa
> HCC_HOME     : /opt/rocm/hcc
> HCC clang version 5.0.0  (based on HCC 1.0.17262-7871d33-3953233-56b5abe )
> Target: x86_64-unknown-linux-gnu
> Thread model: posix
> InstalledDir: /opt/rocm/hcc/bin
> LLVM (http://llvm.org/):
>   LLVM version 5.0.0svn
>   Optimized build.
>   Default target: x86_64-unknown-linux-gnu
>   Host CPU: haswell
> 
>   Registered Targets:
>     amdgcn - AMD GCN GPUs
>     r600   - AMD GPUs HD2XXX-HD6XXX
>     x86    - 32-bit X86: Pentium-Pro and above
>     x86-64 - 64-bit X86: EM64T and AMD64
> HCC-cxxflags :  -hc -std=c++amp -I/opt/rocm/hcc-1.0/include -I/opt/rocm/includeHCC-ldflags  :  -hc -std=c++amp -L/opt/rocm/hcc-1.0/lib -Wl,--rpath=/opt/rocm/hcc-1.0/lib -ldl -lm -lpthread -lunwind -Wl,--whole-archive -lmcwamp -Wl,--no-whole-archive
> 
> === Environment Variables
> ...
> 
> == Linux Kernel
> Hostname     : beamoflight-All-Series
> Linux beamoflight-All-Series 4.9.0-kfd-compute-rocm-rel-1.6-77 #1 SMP Wed Jun 28 07:30:27 CDT 2017 x86_64 x86_64 x86_64 GNU/Linux
> No LSB modules are available.
> Distributor ID:	Ubuntu
[...]

PR #201, opened 2017-10-15
Slight tidying up of the hipify-clang cmake script

ChrisKitching:
- Replace the scary broken-dependency warning described in #94 with a friendlier message.
- Slightly nicer and more standard behaviour when failing to find Python.


ChrisKitching:
Subsumed by #202 - sorry for the spam.

PR #202, opened 2017-10-15
Migrate hipify-clang to a newer version of clang

ChrisKitching:
One obstacle to the use of CUDA 8/9 with `hipify-clang` is that `hipify-clang` was pinned to a fairly old version of clang (3.9.0), which itself does not support CUDA beyond 7.

It turns out to be relatively straightforward to migrate `hipify-clang` to be compatible with the APIs of clang 5.0, in a backward-compatible fashion. Some slight cleanup was done along the way, but nothing super invasive.

Each commit explains itself pretty well. Many of the API changes can be adapted to without having explicit version checks. The program also seems to compile very much faster now - although I would suggest future work might involve breaking it into multiple translation units. Single monolithic files make compilers sad.

So yeah. It now works with CUDA 8/9 (at least, with the features it understands), and you can build it with any version of clang/llvm between 3.8.0 and 5.0. With that in mind, it no longer seems to make sense to restrict the `find_package` used to locate LLVM to exclusively looking in the user-provided path.

### DIFF
--- a/hipify-clang/CMakeLists.txt
+++ b/hipify-clang/CMakeLists.txt
@@ -3,6 +3,13 @@ project(hipify-clang)
 
 option(HIPIFY_CLANG_TESTS "Build the tests for hipify-clang, if lit is installed" ON)
 
+# Disable the tests if `lit` is not installed.
+find_program(LIT_COMMAND lit)
+if (NOT LIT_COMMAND)
+    set(HIPIFY_CLANG_TESTS OFF CACHE INTERNAL "")
+    message(STATUS "hipify-clang's tests are not being built because `lit` is not installed.")
+endif()
+
 if (PARENT_SCOPE)
     set(BUILD_HIPIFY_CLANG 0 PARENT_SCOPE)
 endif()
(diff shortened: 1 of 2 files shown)


ChrisKitching:
CI is apparently angry, but its DNS also doesn't resolve.

Is this a real problem, or is the infrastructure on fire?

y2kenny:
@ChrisKitching, it's IT.  We are seeing the same thing on rocm-ci.

hipify-clang/CMakeLists.txt
```diff
@@ -4,87 +4,78 @@ project(hipify-clang)
...
-    if (NOT ${LLVM_FOUND})
-        message(STATUS "hipify-clang will not be built. To build it please specify absolute path to LLVM 3.8 or LLVM 3.9 package using -DHIPIFY_CLANG_LLVM_DIR")
-    endif()
-endif()
-if (${LLVM_FOUND})
-    list(APPEND CMAKE_MODULE_PATH ${LLVM_CMAKE_DIR})
-    include(AddLLVM)
+find_package(LLVM PATHS ${HIPIFY_CLANG_LLVM_DIR} REQUIRED)
 
-    message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
```
emankov: Please leave this message if LLVM_FOUND for communicating with users and build engineers.

ChrisKitching: Note that I removed the `QUIET` argument from the call to `find_package(llvm)`. The LLVM finder-system prints:

```
-- Found LLVM 5.0.0
```

All by itself.

Do you want two messages?

ChrisKitching:
> Please merge it with the following:
> 87cf949

I've cherry-picked that commit onto the start of my branch, and tidied up around it for you. Hope this looks better now!

hipify-clang/CMakeLists.txt
emankov: No, single is enough :+1:

ChrisKitching:
Subsumed by #213. Maintaining four divergent branches has become too irritating to continue with, so I've merged together the bits that got positive review responses over there so I don't have to rebase so often ;)

Issue #206, opened 2017-10-16
[HIPIFY] hipify-clang does not cope with filenames that differ only in extension


ChrisKitching:
A common situation is for people to have header files and `.cu` files with the same name: such as `axpy.cu` and `axpy.h`/`axpy.cuh`. There might even be `axpy.cpp`, too.

When you feed all of those files to `hipify-clang` at the same time, it will decide it wants to output all of them to `axpy.hip`, and output an arbitrary one of them to that file, losing the other.

The obvious solution might be to decide on a file extension for "hip header". We've got `.cuh` for CUDA, so let's have something like `.hiph` or similar for hip?

The practical effect of this bug is that the only way to convert any meaningfully-sized project is to call `hipify-clang -o <something> <somefile> -- ...` in a loop for every file. Makes multiple-input mode a bit unuseful, really.

Unrelatedly, it would also be nice to be able to specify an output _directory_ when using multiple-file input mode...

One other solution might be to simply drop multiple-input-mode entirely, so it's clear to users that they have to write a loop script or suchlike.

PR #210, opened 2017-10-16
[HIPIFY] Alter output file behaviour slightly

ChrisKitching:
The LLVM APIs modify the input file in-place. To avoid overwriting the input file,
the original implementation was therefore copying the input to the output and
operating on the output. This fails, however, because it means relative includes
are being resolved relative to the output path.

To resolve this, I here change the behaviour to:
- Copy the input file to `<src>.hipify-tmp`
- Operate on that temporary file
- Move the temporary file to the output (which will overwrite the input if we're
  in in-place mode).

This strategy has the added advantage that you won't corrupt the input source file
if the program crashes while operating in in-place mode.

### DIFF
--- a/hipify-clang/src/Cuda2Hip.cpp
+++ b/hipify-clang/src/Cuda2Hip.cpp
@@ -4174,6 +4174,12 @@ void printAllStats(const std::string &csvFile, int64_t totalFiles, int64_t conve
   csv.close();
 }
 
+void copyFile(const std::string& src, const std::string& dst) {
+  std::ifstream source(src, std::ios::binary);
+  std::ofstream dest(dst, std::ios::binary);
+  dest << source.rdbuf();
+}
+
 int main(int argc, const char **argv) {
   auto start = std::chrono::steady_clock::now();
   auto begin = start;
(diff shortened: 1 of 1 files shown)


ChrisKitching:
> source.cu -> source.hip.cu (was source.hip)
> source.cuh -> source.hip.cuh (was source.hip), etc.

This was intentional to fix #206, but I think you're right it looks better to have `.hip` on the end. Maybe there's an existing convention for this stuff?

If there's a canonical file extension for hip-headers, then we probably should use that. So `*.cu` -> `*.hip` and `*.(cuh|h|hpp)` -> `*.hiph` (or whatever the extension is...). Might look better than just gluing `.hip` onto it somewhere?

emankov:
> If there's a canonical file extension for hip-headers, then we probably should use that. So *.cu -> *.hip and *.(cuh|h|hpp) -> *.hiph (or whatever the extension is...). Might look better than just gluing .hip onto it somewhere?

The problem here is that the source files might be with any possible extension. AFAIK, we have no any agreement on HIP header files - all the hipified files are .hip (**Ben, correct me if I wrong**), and that was the main reason for my first comment.

emankov:
Please change the following code in order to hipified code always has *.hip extension, for instance: 
source.cu -> source.cu.hip, source.cuh -> source.cuh.hip, etc.:
```c++
if (dst.empty()) {
  dst = src;
  if (!Inplace) {
    size_t pos = dst.rfind(".");
    if (pos != std::string::npos && pos + 1 < dst.size()) {
      dst = dst.substr(0, pos) + ".hip." + dst.substr(pos + 1, dst.size() - pos - 1);
    } else {
      dst += ".hip.cu";
    }
  }
}
```

ChrisKitching:
Done.

PR #212, opened 2017-10-17
[HIPIFY] Don't insert the hip include repeatedly

ChrisKitching:
The way the matchers were wired up, the hip include was being inserted
as many times as there are cuda language constructs in the file. Eep!

### DIFF
--- a/hipify-clang/src/Cuda2Hip.cpp
+++ b/hipify-clang/src/Cuda2Hip.cpp
@@ -3817,27 +3817,24 @@ public:
   }
 
   void run(const MatchFinder::MatchResult &Result) override {
-    do {
-      if (cudaCall(Result)) break;
-      if (cudaBuiltin(Result)) break;
-      if (cudaEnumConstantRef(Result)) break;
-      if (cudaEnumDecl(Result)) break;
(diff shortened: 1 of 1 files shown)


emankov:
The loop assumes that there might be more than one iteration in one run. Are you sure that it'll never happen?

ChrisKitching:
> The loop assumes that there might be more than one iteration in one run. Are you sure that it'll never happen?

You mean the `while(false)` loop?

emankov:
Right!

ChrisKitching:
> The loop assumes that there might be more than one iteration in one run. Are you sure that it'll never happen?

> Right!

I must be missing something...

It's a `do { ... } while (false)` loop. So it cannot possibly run for more than one iteration.

Perhaps I've not understood your question? :D

emankov:
Ouch... The logic is changed, there were a conditional loop exit previously, I recalled it and missed (false) condition. :D

ChrisKitching:
Subsumed by #213. Maintaining four divergent branches has become too irritating to continue with, so I've merged together the bits that got positive review responses over there so I don't have to rebase so often ;)

Issue #207, opened 2017-10-16
[HIPIFY] hipify-clang misbehaves in the presence of preprocessor directives


ChrisKitching:
Consider the following CUDA program:

```c++
__global__ void axpy_kernel(float a, float* x, float* y) {
    y[threadIdx.x] = a * x[threadIdx.x];
}

void axpy(float a, float* x, float* y) {
    axpy_kernel<<<1, 4>>> (a, x, y);
#ifdef SOME_MACRO
    axpy_kernel<<<1, 4>>> (a, x, y);
#endif
}
```

The result of hipifying this, if you don't pass `-DSOME_MACRO` is:

```c++
#include <hip/hip_runtime.h>
__global__ void axpy_kernel(float a, float* x, float* y) {
    y[hipThreadIdx_x] = a * x[hipThreadIdx_x];
}

void axpy(float a, float* x, float* y) {
    hipLaunchKernelGGL(axpy_kernel, dim3(1), dim3(4), 0, 0, a, x, y);
#ifdef SOME_MACRO
    axpy_kernel<<<1, 4>>> (a, x, y);
#endif

}
```

Respecting conditional macros isn't the right thing to do with this sort of mechanised refactoring - what you really want to do is walk the entire tree applying your refactor, regardless of preprocessor conditionals.

This is going to present a relatively nasty obstacle to people with complicated CUDA programs they want to translate...

emankov:
Preprocessor is needed to expand macroses and hipify them also. The obstacle is known, we tried a few approaches, including search of all macro conditions and calling hipify-clang recurrently on the same source with setting an appropriate define, but it turned out very complicated with a real performance drop. Actually we need a bit strange with hipifyication: we need preprocessor and at the same time we need the source code untouched by preprocessor.

ChrisKitching:
> Preprocessor is needed to expand macroses and hipify them also.

Can you give an example of a situation when you need to have a macro expanded?

Why isn't it enough to just look at the source code, fix all the macro _definitions_, and move on? When do you need the preprocessed source?

emankov:
> Why isn't it enough to just look at the source code, fix all the macro definitions, and move on?

Preprocessor directives are not included in AST.

> Can you give an example of a situation when you need to have a macro expanded?
> When do you need the preprocessed source?

CUDA_8.0/CUDASamples/common/inc/helper_cuda.h:

```c++
[...]

PR #213, opened 2017-10-17
[HIPIFY] Simplify (and accelerate) hipification of CUDA type identifiers

ChrisKitching:
See commit messages for details (they're long).

The high-level goal here is twofold:
- Making a third of the code go away, so solving more complex problems is
  easier
- Making things faster by reducing the size of the lookup tables being used
  for the hottest queries. More work can be done in this area.


hipify-clang/src/CUDAToHipMap.cpp
```diff
@@ -0,0 +1,2697 @@
+#include "CUDAToHipMap.h"
```
emankov: CUDA**2**HipMap.h and CUDA**2**HipMap.cpp

hipify-clang/src/Cuda2Hip.cpp
```diff
@@ -620,289 +620,50 @@ class Cuda2HipCallback : public MatchFinder::MatchCallback, public Cuda2Hip {
...
-            FullSourceLoc fullSL(sl, *SM);
-            insertReplacement(Rep, fullSL);
-          }
-        } else {
-          std::string msg = "the following reference is not handled: '" + name.str() + "' [struct var ptr].";
-          printHipifyMessage(*SM, sl, msg);
-        }
+      const hipCounter& hipCtr = found->second;
+      if (hipCtr.unsupported) {
+          llvm::errs() << "[HIPIFY] Unsupported type: " << typeName << "\n";
```
emankov: Better to give FullSourceLoc for the found unsupported type.

ChrisKitching: Oh, right. I was meaning to port all of these over to the Diagnostics API anyway.

emankov: What about testing? What sources were tested?

ChrisKitching: In addition to my CUDA code I'm trying to get working, I wrote the following contrived program to try to expose failures in my changed code:
https://gist.github.com/ChrisKitching/827dbc17acd939c3cd4f9b898006cc02

We should probably add some more testcases to the `lit` testsuite - current coverage there is relatively poor. While the automated test suite fails (and, indeed, it seems broken in general), that input file yields correct-looking output, too.

ChrisKitching: (Fixed both your review comments now, too)

emankov: We will definitely extend testing. I thought to add also golden files comparison. And btw lit testing is not the demand. As for you tremendous changes in this pull request, I'd like you to perform more testing manually and diff against current hipify-clang, or we may do it together.
1.	https://github.com/ROCm-Developer-Tools/HIP/blob/master/samples/0_Intro/square/square.cu
2.	https://github.com/ROCm-Developer-Tools/HIP/files/1389647/Interac-v2.zip
3.	https://github.com/ROCm-Developer-Tools/HIP/blob/master/tests/src/cudaRegister.cu
4.	https://github.com/AMDComputeLibraries/hip/blob/master/examples/rodinia_3.0/cuda/backprop/backprop_cuda.cu
5.	CUDA SDK samples, I’ll select a few.

ChrisKitching:
> I thought to add also golden files comparison.

I've added some commits to this PR that alter the existing  `lit` test to provide useful failure messages. Previously, the use of `grep` meant you'd never get useful failures from it, it'd just say "FAIL", with no information about what went wrong.

By using LLVM `FileCheck` instead, you now have far simpler check lines:

```diff
-  // RUN: sh -c "test `grep -c -F 'hipMalloc(&device_y, kDataLen * sizeof(float));' %t` -eq 2"
+  // CHECK: hipMalloc(&device_x, kDataLen * sizeof(float));
```

And you get _much_ better failure messages (in _colour_, like normal clang errors):

```
/home/chris/HIP/tests/hipify-clang/axpy.cu:31:12: error: expected string not found in input
 // CHECK: hipLaunchKernel(HIP_KERNEL_NAME(axpy), dim3(1), dim3(kDataLen), 0, 0, a, device_x, device_y);
           ^
<stdin>:31:2: note: scanning from here
 //
 ^
<stdin>:33:2: note: possible intended match here
 hipLaunchKernelGGL(axpy, dim3(1), dim3(kDataLen), 0, 0, a, device_x, device_y);
 ^
```

(And I fixed that failure, too ;) )

I'll go ahead and add some of those files you linked as additional `lit` tests, in addition to the diffing you requested.

tests/hipify-clang/axpy.cu
```diff
@@ -1,9 +1,9 @@
-// RUN: hipify "%s" -o=%t -- %cuda_args
+// RUN: hipify "%s" -o=%t -- %cuda_args && cat %t | sed -Ee 's|//.+|// |g' | FileCheck %s
```
ChrisKitching: One thing we might like to do later is make `hipify` use clang's normal output specification options. This'll make kludges like this one unnecessary, which would be nice. (If we had a way to output to stdout, this line could lose the `&& cat %t` and the `-o=%t` parts.)

The `sed` part is just to make failures nicer. We have to strip the `CHECK` directives from the input file, otherwise all the tests will triviall pass by matching with their own definitions. This sed line does that, leaving the `//` in, so the line-count is unchanged, and so the line number in error messages is correct.

hipify-clang/src/Cuda2Hip.cpp
ChrisKitching: > https://github.com/ROCm-Developer-Tools/HIP/files/1389647/Interac-v2.zip

This does not contain any CUDA.

ChrisKitching: > https://github.com/AMDComputeLibraries/hip/blob/master/examples/rodinia_3.0/cuda/backprop/backprop_cuda.cu

That's a 404. Is that an internal AMD repo I don't get to see, perhaps?

ChrisKitching:
The changes here got extensive enough that keeping them orthogonal to the changes in #212 and #202 became very irritating, and liable to introduce new problems.

So: I've prepended #212 and #202 (but _not_ #210, since we're not done arguing about that yet ;) ) onto this PR to keep maintainence easier.

To this PR I have added two new lit tests: `cudaRegister.cu` and `square.cu` that you pointed me to. Let me know if you have any questions about how to add more - it should be pretty self-explanatory (and you are firmly encouraged to go nuts adding lots more :smile:  ).

The new tests exposed a bug in the translation of unary preprocessor macros, which I have also fixed. There were some slight amendments to the earlier work here, as well.

Here are the diffs you requested - the ones for which you didn't give me broken links, anyway :wink: . Note that these diffs are _from_ old _to_ new:
- [axpy.cu](https://gist.github.com/ChrisKitching/c44f66e72eb0c94f10b49ecad4f31c52)
- [cudaRegister.cu](https://gist.github.com/ChrisKitching/84d90e1b20d6f6fbd1323520cc898a8d) - Yes - the _old_ version got this one wrong.
- [square.cu](https://gist.github.com/ChrisKitching/e7949040e18369778955f488c3d492d0) - the old version made a complete mess of this. See [my earlier comment](https://github.com/ROCm-Developer-Tools/HIP/issues/207#issuecomment-337521851) describing the bug here, fixed in d47ef74bacc481bd7a7e17b1225ff4b642dbebc0.

ChrisKitching:
To be completely clear: those diffs are the diffs of the _hip_, from old (current `master`) to new (the tip of this PR branch).

emankov:
Are we ready for merging? I'd like to stop for a while and to merge the changes. I'd prefer to split them (logically) too.

ChrisKitching:
> Are we ready for merging? I'd like to stop for a while and to merge the changes. I'd prefer to split them (logically) too.

How would you like to proceed?

I think we're happy to go with the clang-version-update changes?

The changes to the unit tests are independent of everything else, so those can also be merged separately.

This would then leave the changes to the compiler itself, which are surely the most hazardous and controversial part?

ChrisKitching:
> Are we ready for merging? I'd like to stop for a while and to merge the changes. I'd prefer to split them (logically) too.

To more generally answer your question: yes, I'm done adding changes to this branch. It's already a ridiculously large PR. I'm experimentally locally with some further work, but wasn't planning to submit anything else for review until we clear the current backlog.

We could maybe add even more testcases? But we can always do that later... I suspect that'll be an ongoing process now we have `lit` working properly :D

emankov:
May be to split the PR at least to the following parts: 
1. Migrate hipify-clang to a newer version of clang.
2. Simplify (and accelerate) hipification of CUDA type identifiers (incl. lit testing).
Or may be to even more parts.

ChrisKitching:
> May be to split the PR at least to the following parts:

You're more than welcome to review work here, too. Github has very handy tools for stepping through changesets one at a time.

> Migrate hipify-clang to a newer version of clang.

See #219 . It's literally a prefix of this branch, but now it's in it's own PR.

I'll resubmit prefixes of this branch one at a time as new PRs, then.

ChrisKitching:
This branch has been cut back to the very limited thing described in the title. Should be easier to review now. I'll add the other bug fixes and automated tests as later PRs as we clear the backlog.

ChrisKitching:
> Please double check the build and that tests are passing.

I've rebased on master. A couple of problems appeared:

- I realised I never applied your earlier comment about renaming to `CUDA2HipMap.h ` - sorted that.
- 062127c5ea1a8a7fa238c40e8526ac1b31a71ca3 broke the lit tests, because they relied on the redundant `BINARY_DIR` variable for substitution. Sorry for not realising that sooner, but it's very straightforward to fix: I've added a commit to this branch that sorts it out.

The tests fail, of course, but no moreso than they did before. The tests are written to the spec on my fork - where several more bugs have been fixed. It'll go green when we clear the review backlog :D

I'll file the next round of PRs now. Onward!

emankov:
Chris, could you please double check, that all hipify-clang options are also working?

PR #214, opened 2017-10-18
hipDeviceReset(): make sure to reinitialize the printf buffer in hcc RT

scchan:
- re-initialize the printf buffer in HCC RT after resetting the am memory tracker
- this will fix several HIP directed test failures observed with a recent HCC compiler
- this PR depends on  https://github.com/RadeonOpenCompute/hcc/pull/508

### DIFF
--- a/src/hip_hcc.cpp
+++ b/src/hip_hcc.cpp
@@ -614,6 +614,11 @@ void ihipDevice_t::locked_reset()
     _state = 0;
     am_memtracker_reset(_acc);
 
+    //FIXME - Calling am_memtracker_reset is really bad since it destroyed all buffers allocated by the HCC runtime as well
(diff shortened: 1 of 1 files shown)


scchan:
All the previous failures due to the printf enablement in HCC have been fixed.
The only HIP directed test that is failing is hip_ffs but I don't think it's caused by my changes.

bensander:
This is fine for now but for future HIP/HCC features let's create and use a flag-based feature mechanism rather than the version checks.  The basic idea would be that hcc provides an include file (hcc_features.h) with a set of defined (#define HCC_HAS_ INIT_PRINT_BUFFER).  This will provide more robust feature query in the face of cherry-picks and merges.  

Also - eventually reset will be performed at KFD level and remove all HCC structures.  So we need to have an HCC architecture that supports reset.  I expect that will grow to include more structures than just the print buffer.

PR #218, opened 2017-10-19
Add [[nodiscard]] attribute to hipError_t in C++17 mode

ChrisKitching:
This will yield a compiler warning any time you discard an error
return value from an API call.

Some thoughts:
- Should we move the definition of `__HIP_NODISCARD` into `hip_common.h` - it
  might well prove useful elsewhere, after all.
- A very common pattern (at least in CUDA code) is to wrap every API call in
  a macro that implements some flavour of "If exception, crash". Perhaps hip
  could provide one of those out of the box as a convenience?
- Perhaps there could be a macro that, when set, causes hip errors to propagate
  as exceptions instead of this C-style API? I guess that would be part of a
  more general C++ API for hip (which would, admittedly, slightly defeat the
  point of being a close match to CUDA. But CUDA is very much clunky to use in
  some situations)...

### DIFF
--- a/include/hip/hip_runtime_api.h
+++ b/include/hip/hip_runtime_api.h
@@ -143,6 +143,13 @@ typedef struct hipPointerAttribute_t {
  *
  */
 
+// Ignoring error-code return values from hip APIs is discouraged. On C++17,
+// we can make that yield a warning
+#if __cplusplus >= 201703L
+#define __HIP_NODISCARD [[nodiscard]]
+#else
+#define __HIP_NODISCARD
+#endif
 
 /*
  * @brief hipError_t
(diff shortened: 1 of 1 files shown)


ChrisKitching:
Oh, one other random comment. I notice you have this nifty C++ wrapper for `hipMalloc`:

```c++
template<class T>
static inline hipError_t hipMalloc ( T** devPtr, size_t size)
{
    return hipMalloc((void**)devPtr, size);
}
```

Perhaps there would also be value in having one that provides array-length semantics? ie:
```c++
template<class T>
static inline hipError_t hipMallocArray( T** devPtr, size_t num)
{
    return hipMalloc((void**)devPtr, num * sizeof(T));
}
```

After all, a ridiculously common mistake when using `malloc`-style interfaces is failure to multiply by the sizeof the allocated type...

These C++ overrides could also accept the return-ptr by reference not double-pointer - although that might introduce an unwelcome inconsistency w.r.t the C APIs...

bensander:
Thanks Chris!  I'm open to all of the suggestions above.  One of HIP's Prime Directives is that any improvements to the language have to work on both CUDA and ROCm platforms.  So any functional changes (new apis, new parms, etc) need to be supported on both paths.  
 
This one looks good - hip/hip_runtime.h is perfect location.  

We have a pretty good macro check in the tests/src/test_common/test_common.h - shows the error code and message.  I'd probably drop the colors in a more general macro.  We should also check to ensure the application doesn't already define the macro, a simple if not defined(...) would probably work just fine.

PR #220, opened 2017-10-20
Support default arguments in C as well as C++

ChrisKitching:
While strict C89 doesn't allow `void foo(int x = 3)`, you can
still use function overloading to achive the same effect.

This is necessary for parity with CUDA's (documented) API, and
should avoid a surprising failure for people who `hipify` C code
that relies on this behaviour.


ChrisKitching:
*ahem*

More coffee required.

PR #225, opened 2017-10-23
[HIPIFY] cmake: fix standalone build

emankov:
After switching HIPIFY_CLANG_TESTS on by default it turned out that standalone build is broken.
Change relies on an assumption that hipify-clang is in HIP folder as it is so for now, but it will be changed with moving of hipify-clang to a separate github repo.

### DIFF
--- a/hipify-clang/CMakeLists.txt
+++ b/hipify-clang/CMakeLists.txt
@@ -73,23 +73,15 @@ if (HIPIFY_CLANG_TESTS)
     # tests
     find_package(PythonInterp 2.7 REQUIRED EXACT)
 
-    set(BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
-
-    if (CMAKE_CURRENT_SOURCE_DIR EQUAL CMAKE_SOURCE_DIR)
-        set(SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
-    else()
-        set(SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/..)
-    endif()
-
     configure_file(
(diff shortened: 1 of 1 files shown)


hipify-clang/CMakeLists.txt
```diff
@@ -75,15 +75,21 @@ if (HIPIFY_CLANG_TESTS)
...
+        set(SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+    else()
+        set(SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/..)
+    endif()
+
     configure_file(
-        ${CMAKE_SOURCE_DIR}/tests/hipify-clang/lit.site.cfg.in
-        ${CMAKE_CURRENT_BINARY_DIR}/tests/hipify-clang/lit.site.cfg
+        ${SOURCE_DIR}/tests/hipify-clang/lit.site.cfg.in
+        ${BINARY_DIR}/tests/hipify-clang/lit.site.cfg
```
ChrisKitching: I'm pretty sure you can make this unconiditional by using `CMAKE_CURRENT_LIST_DIR` - that gets you the directory in which the current `CMakeLists.txt` lives. In this case, that'll get you the `hipify-clang` directory, and then you can just compute your paths from there, something like:

```
${CMAKE_CURRENT_LIST_DIR}/../tests/hipify-clang/lit.site.cfg.in
```

This also perhaps raises the question of why the `hipify-clang` tests aren't in the `hipify-clang` repo? :grin:

emankov: Historically. We'll move the whole HIPIFY to separate github repo soon, including tests and docs.

PR #226, opened 2017-10-23
add printf to HIP device functions

scchan:
- created a device version of `printf`
- added unit test


include/hip/hcc_detail/hip_runtime.h
```diff
@@ -50,10 +50,16 @@ THE SOFTWARE.
...
+// define HIP_ENABLE_PRINTF to enable printf
+#ifdef HIP_ENABLE_PRINTF
+  #define HCC_ENABLE_ACCELERATOR_PRINTF 1
+#endif 
+
 //---
 // Remainder of this file only compiles with HCC
 #if defined __HCC__
 #include <grid_launch.h>
+#include "hc_printf.hpp"
```
bensander: Does this hc_printf.hpp file exist on older (1.6.4) version of HCC?  If not we should guard with workweek check so HIP continues to compile with older HCC.

scchan: it builds with hcc 1.6.4 but the printf feature will not be available to hip

PR #227, opened 2017-10-24
Tweak some version numbers in clang version compatibility checks

ChrisKitching:
Apparently a couple of those APIs changed in clang 5, not 4.

Got an email from a disgruntled AMD person about this ;)

### DIFF
--- a/hipify-clang/src/Cuda2Hip.cpp
+++ b/hipify-clang/src/Cuda2Hip.cpp
@@ -3031,7 +3031,7 @@ public:
     : Cuda2Hip(R, mainFileName) {}
 
   virtual bool handleBeginSource(CompilerInstance &CI
-#if LLVM_VERSION_MAJOR < 4
+#if LLVM_VERSION_MAJOR <= 4
                                  , StringRef Filename
 #endif
                                  ) override {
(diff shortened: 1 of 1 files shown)


hipify-clang/src/Cuda2Hip.cpp
```diff
@@ -3114,7 +3114,7 @@ class HipifyPPCallbacks : public PPCallbacks, public SourceFileCallbacks, public
     if (_sm->isWrittenInMainFile(MacroNameTok.getLocation())) {
         // The getNumArgs function was rather unhelpfully renamed in clang 4.0. Its semantics
         // remain unchanged.
-#if LLVM_VERSION_MAJOR > 3
+#if LLVM_VERSION_MAJOR > 4
```
emankov: I've tested build on 3.8, 3.9 and 4.0 and there were no errors... Looks strange.

mattsinc: I can vouch that this change was necessary for me to avoid errors when running make ... and I'm using clang 4.0.

Matt

emankov: Do you mean the change where LLVM_VERSION_MAJOR >3 or >4?

emankov: Just verified. The misunderstanding was because of cmake change which has eliminated NO_DEFAULT_PATH in find_package for LLVM. That is why on my side LLVM 4.0.0svn has been found in system folder and hipify-clang was successfully compiled without this change.

ChrisKitching: Cmake has its own mechanisms for 
users to control how it searches for packages in their environment. The prefix and ignore paths for example.


 Excluding the system paths entirely is very much not the right solution here.

emankov: Chris, I've just answered here https://github.com/ROCm-Developer-Tools/HIP/pull/230#issuecomment-339035062 . This is a temporary solution. You right for 100%, but let's eliminates all the problems in the existed environment, and then move forward.

PR #239, opened 2017-10-30
[newTests]Adding tests for device APIs

phani544:
Added new tests for device runtime APIs. 
hipDeviceComputeCapability
hipDeviceGetName
hipDeviceSynchronize
hipDeviceTotalMem
hipGetDevice
hipRuntimeGetVersion
hipSetCachceConfig
hipSetDevice
hipSetDeviceFlags

### DIFF
--- a/tests/src/runtimeApi/device/hipDeviceGetName.cpp
+++ b/tests/src/runtimeApi/device/hipDeviceGetName.cpp
@@ -30,7 +30,7 @@ THE SOFTWARE.
 
 #include "test_common.h"
 
-const int len 256;
+#define len 256
 
 int main()
 {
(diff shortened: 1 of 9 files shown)


bensander:
Looks OK to me as a fine start.  In the future we need to expand to cover harder cases.  hipDeviceSynchronize is a good example of a typical programming pattern used to overlap host and device code.

bensander:
@mangupta - can you also review and merge if looks good to you?

tests/src/runtimeApi/device/hipDeviceGetName.cpp
```diff
@@ -0,0 +1,45 @@
...
+ * RUN: %t
+ * HIT_END
+ */
+
+#include "test_common.h"
+
+int main()
+{
+    int numDevices = 0;
+    char name[256];
```
bensander: Use a const int here rather than hard-coded 256

PR #246, opened 2017-11-01
Use natural dispatch syntax

AlexVlx:
This switches HIP from its currently convoluted macro + pfe based dispatch mechanism to a more natural one partially based on the existing module API. The basic idea is that HCC will always correctly emit __global__ functions: as empty-bodied stubs, on host, and as kernels, on device. It then becomes trivial to obtain the mangled name on host, at dispatch, from the function's address, and then to use the mangled name to retrieve the kernel. This should address all problems stemming from serialisation, dubious mismatches due to the manufactured functor, macro-isms et al. It also immediately enables support for generalised globals as a consequence of that being available in the module API. Finally, it will make debug much easier, since the actual names of the __global__ functions will automatically be used in traces etc. One detail is that due to how dispatch works now (hipLaunchKernel and hipLaunchKernelGGL are themselves  variadic function templates which deduce the function type of the callee), in certain cases it may be necesssary to insert explicit casts to ensure that the variadic argument list selects a viable overload - this can be observed in some unit tests. Eventually we may be able to remove this limitation, but for now it does not appear terribly onerous. The code is not extremely HIPpie, nor is it fully optimised, but rather is intended as a starting point for the HIP team to make its own.


src/grid_launch.cpp
```diff
@@ -21,76 +21,118 @@ THE SOFTWARE.
...
+        if (it0 == functions().cend()) {
+            throw runtime_error{
+                "No device code available for function: " +
+                name(function_address)
+            };
+        }
+
+        auto agent = target_agent(stream);
+
+        const auto it1 = find_if(
```
pfultz2: Shouldn't this be `std::find_if`, or is there a reason to leave this up to ADL?

src/program_state.cpp
```diff
@@ -0,0 +1,498 @@
...
+            lock_guard<std::mutex> lck{mtx};
+            globals.emplace_back(p, hsa_amd_memory_unlock);
+        }
+    }
+
+    template<typename P>
+    inline
+    section* find_section_if(elfio& reader, P p)
+    {
+        const auto it = find_if(
```
pfultz2: Also, here is another ADL-based `find_if`.

AlexVlx:
https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/104 is needed in HCC's upstream clang to support this.

whchung:
@AlexVlx I think this PR deserves some high-level overview, and possibly be broken into multiple PRs, as it seems to try achieve several different targets at the same time.

- remove trailing whitespaces and re-indent some kernel call sites: better be filed as a separate PR
- use `clang_offload_bundler` to bundle code objects: the syntax of the triple, and the header used for `clang_offload_bundler` should better be documented.
- revise kernel compilation / launch logic

whchung:
@AlexVlx one extra question: would `hccgenco.sh` be obsolete with this PR?

AlexVlx:
@pfultz2 : good eyes:)
@bensander: please see below for a partial answer.
@whchung:
a) trailing whitespace - I'll separate that, it's accidental but maybe useful;
b) re-indentation - that is actually a side-effect and the least important part of modifying the call-site to accommodate differences introduced by the new launch mechanism (argument pack must match a valid overload), therefore moving it elsewhere is undesirable since context would be lost;
c) not sure what you mean by "use `clang_offload_bundler` to..." - I think that this is what we use and have been using for more than a year now, so this merely builds on that; the format is documented here (perhaps adding a link in the header would be of use) https://code.woboq.org/llvm/clang/tools/clang-offload-bundler/ClangOffloadBundler.cpp.html#173; the triple is still possibly subject to change, so it appears uncertain that we want to lock that in just yet;
d) changing `hipMemset` to use natural launch syntax seems rather related, and avoids quaint side-effects / interactions with PFE; the alternative would be to change it later, but that may add noise in the current validation;
e) `hccgenco.sh` should become obsolete, yes, since we can directly load a bundled code object from an ELF file and can also easily build the functionality that takes a code object packaged as a binary blob (`foo.hipbin` or whatever, if we want to retain alignment with CUDA) without needing script magic.
Thank you for the feedback gentlemen!

AlexVlx:
@whchung @bensander #255 handles the noisy whitespace differences.

whchung:
@AlexVlx i guess my question for `clang-offload-bundler` is do we follow the same triple syntax currently used in `hcc` when we create a bundle:

```
hcc-amdgcn--amdhsa-$AMDGPU_TARGET
```

from the logic I think the answer is yes, but i'd like to double check with you.

for `hipMemset` I agree it's definitely good to use the new natural dispatch syntax.

since `hccgenco.sh` would be made obsolete, i believe relevant tests / samples should also be modified. do you expect to amend them in this PR, or they would come in a separate PR?

whchung:
invite @mangupta as directed tests / samples which depend on `hccgenco.sh` may have to be changed.

AlexVlx:
@whchung @mangupta making hccgenco.sh obsolete is a two step process, which I think can follow going through separately, after we're done with this:
a) due to https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/104 we should no longer need to use WrapperGen / define GENERIC_GRID_LAUNCH=0 for genco, since all __global__ functions will get emitted; we can and should explore this change now, since it may fix some lingering bugs in / due to WrapperGen;
b) we want to consider extending --filetype in llc (perhaps) to include --filetype=hsaco to emit just the code object for a source file / collection of source files (or a similar mechanism); once that is in place, genco's more nefarious role as gutripper is no longer necessary, and it can be fully removed.

whchung:
@bensander / @mangupta , https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/104 has been merged into `HCC` mainline now. I think we may need your extra round of review for this PR.

whchung:
@AlexVlx i agree it's time to try get rid of `WrapperGen` now.

for `llc --filetype=hsaco` I'm not sure if that's a good idea. doing this seems to mix 2 things: `llc` (IR -> object / assembly) and `lld`  (object -> ELF) together.

bensander:
Thanks Jack.  We need to debug some of the failing tests so may take a bit to settle out.  Does tip HCC still work with old HIP?

AlexVlx:
@bensander: the changes in the tip of HCC are non-intrusive, it will keep working with existing HIP.
@whchung: yeah, agreed, I think that instead of llc what I probably should have said is that we should consider introducing a flag which can be consumed by HCC and which yields a .hsaco file which is basically a serialised code object; that would simplify this workflow a lot.

src/hip_memory.cpp
```diff
@@ -1161,42 +1161,56 @@ hipError_t hipMemcpy3D(const struct hipMemcpy3DParms *p)
...
+    void hip_fill_n(RandomAccessIterator f, N n, T value)
+    {
+        const uint32_t grid_dim = hipGridDim_x;
 
-    int threads = wg * threads_per_wg;
-    if (threads > sizeBytes) {
-        threads = ((sizeBytes + threads_per_wg - 1) / threads_per_wg) * threads_per_wg;
+        size_t idx = hipBlockIdx_x * block_dim + hipThreadIdx_x;
+        while (idx < n) {
+            new (&f[idx]) T{value};
```
bensander: This is just a simple assignment correct?  What is motivation to use placement new here - especially inside device kernel?

src/hip_hcc_internal.h
```diff
@@ -247,7 +247,7 @@ static const DbName dbName [] =
 #if COMPILE_HIP_DB
 #define tprintf(trace_level, ...) {\
     if (HIP_DB & (1<<(trace_level))) {\
-        char msgStr[1000];\
+        char msgStr[2000];\
         snprintf(msgStr, 2000, __VA_ARGS__);\
```
bensander: ooops...

bensander:
LGTM to me, two requested changes before we close:
- Need fallback so it compiles with HCC 1.7
- Replace the placement new in memset with a straightforward copy.  In this case predictable performance and clear implementation for an integer copy routine is more important than a general purpose fill routine capable of using any class.

AlexVlx:
@bensander done and done - if you'd care to experiment with it and counter-verify that things pass on your end as well, including on old(er) HCC, it would be neat. Thank you.

bensander:
@AlexVlx - the 1.6 tests failed, see the Jenkins results.  Here is a snippet of the error:


+ make build_tests -i -j8
[  0%] Building HIPCC object CMakeFiles/directed_tests.kernel.hipTestConstant.dir/tests/src/directed_tests.kernel.hipTestConstant_generated_test_common.cpp.o
[  0%] Building HIPCC object CMakeFiles/directed_tests.kernel.hipGridLaunch.dir/tests/src/directed_tests.kernel.hipGridLaunch_generated_test_common.cpp.o
[  0%] Building HIPCC object CMakeFiles/directed_tests.kernel.hipPrintfKernel.dir/tests/src/directed_tests.kernel.hipPrintfKernel_generated_test_common.cpp.o
[  0%] Building HIPCC object CMakeFiles/directed_tests.kernel.hipTestMemKernel.dir/tests/src/directed_tests.kernel.hipTestMemKernel_generated_test_common.cpp.o
[  0%] Building HIPCC object CMakeFiles/directed_tests.kernel.hipLanguageExtensions.dir/tests/src/directed_tests.kernel.hipLanguageExtensions_generated_test_common.cpp.o
[  0%] Building HIPCC object CMakeFiles/directed_tests.kernel.hipLaunchParm.dir/tests/src/directed_tests.kernel.hipLaunchParm_generated_test_common.cpp.o
[  0%] Building HIPCC object CMakeFiles/directed_tests.kernel.inline_asm_vadd.dir/tests/src/kernel/directed_tests.kernel.inline_asm_vadd_generated_inline_asm_vadd.cpp.o
[  0%] Building HIPCC object CMakeFiles/directed_tests.kernel.hipEmptyKernel.dir/tests/src/directed_tests.kernel.hipEmptyKernel_generated_test_common.cpp.o
<inline asm>:1:2: error: invalid instruction mnemonic 'v_add_f32_e32'
        v_add_f32_e32 %xmm0, %xmm0, %xmm1
        ^~~~~~~~~~~~~
error: cannot compile inline asm
1 error generated.
[  0%] Building HIPCC object CMakeFiles/directed_tests.kernel.hipLaunchParm.dir/tests/src/kernel/directed_tests.kernel.hipLaunchParm_generated_hipLaunchParm.cpp.o
[  0%] Building HIPCC object CMakeFiles/directed_tests.kernel.hipTestMemKernel.dir/tests/src/kernel/directed_tests.kernel.hipTestMemKernel_generated_hipTestMemKernel.cpp.o
[  0%] Building HIPCC object CMakeFiles/directed_tests.kernel.hipLanguageExtensions.dir/tests/src/kernel/directed_tests.kernel.hipLanguageExtensions_generated_hipLanguageExtensions.cpp.o
[  0%] Building HIPCC object CMakeFiles/directed_tests.kernel.hipPrintfKernel.dir/tests/src/kernel/directed_tests.kernel.hipPrintfKernel_generated_hipPrintfKernel.cpp.o
[  0%] Building HIPCC object CMakeFiles/directed_tests.kernel.hipEmptyKernel.dir/tests/src/kernel/directed_tests.kernel.hipEmptyKernel_generated_hipEmptyKernel.cpp.o
[  0%] Building HIPCC object CMakeFiles/directed_tests.kernel.hipGridLaunch.dir/tests/src/kernel/directed_tests.kernel.hipGridLaunch_generated_hipGridLaunch.cpp.o
[  0%] Building HIPCC object CMakeFiles/directed_tests.kernel.hipTestConstant.dir/tests/src/kernel/directed_tests.kernel.hipTestConstant_generated_hipTestConstant.cpp.o
Scanning dependencies of target directed_tests.kernel.inline_asm_vadd
[  0%] Linking HIP executable directed_tests/kernel/inline_asm_vadd
/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../x86_64-linux-gnu/crt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
/var/jenkins/workspace/-Developer-Tools_HIP_PR-246-QRTDCR73DQXLTD7NCFVSTVKTUIIWTE7UK2UTWTRZVJCQUECR3I3Q/build/release/staging/lib/libhip_hcc.so: undefined reference to `Coordinates<&hc_get_workitem_id>::x'
/var/jenkins/workspace/-Developer-Tools_HIP_PR-246-QRTDCR73DQXLTD7NCFVSTVKTUIIWTE7UK2UTWTRZVJCQUECR3I3Q/build/release/staging/lib/libhip_hcc.so: undefined reference to `Coordinates<&hc_get_group_size>::X::operator unsigned int() const'
/var/jenkins/workspace/-Developer-Tools_HIP_PR-246-QRTDCR73DQXLTD7NCFVSTVKTUIIWTE7UK2UTWTRZVJCQUECR3I3Q/build/release/staging/lib/libhip_hcc.so: undefined reference to `Coordinates<&hc_get_num_groups>::x'
/var/jenkins/workspace/-Developer-Tools_HIP_PR-246-QRTDCR73DQXLTD7NCFVSTVKTUIIWTE7UK2UTWTRZVJCQUECR3I3Q/build/release/staging/lib/libhip_hcc.so: undefined reference to `Coordinates<&hc_get_num_groups>::X::operator unsigned int() const'
/var/jenkins/workspace/-Developer-Tools_HIP_PR-246-QRTDCR73DQXLTD7NCFVSTVKTUIIWTE7UK2UTWTRZVJCQUECR3I3Q/build/release/staging/lib/libhip_hcc.so: undefined reference to `Coordinates<&hc_get_group_id>::x'
/var/jenkins/workspace/-Developer-Tools_HIP_PR-246-QRTDCR73DQXLTD7NCFVSTVKTUIIWTE7UK2UTWTRZVJCQUECR3I3Q/build/release/staging/lib/libhip_hcc.so: undefined reference to `Coordinates<&hc_get_workitem_id>::X::operator unsigned int() const'
/var/jenkins/workspace/-Developer-Tools_HIP_PR-246-QRTDCR73DQXLTD7NCFVSTVKTUIIWTE7UK2UTWTRZVJCQUECR3I3Q/build/release/staging/lib/libhip_hcc.so: undefined reference to `Coordinates<&hc_get_group_id>::X::operator unsigned int() const'
/var/jenkins/workspace/-Developer-Tools_HIP_PR-246-QRTDCR73DQXLTD7NCFVSTVKTUIIWTE7UK2UTWTRZVJCQUECR3I3Q/build/release/staging/lib/libhip_hcc.so: undefined reference to `Coordinates<&hc_get_group_size>::x'
clang-6.0: error: linker command failed with exit code 1 (use -v to see invocation)
Died at /var/jenkins/workspace/-Developer-Tools_HIP_PR-246-QRTDCR73DQXLTD7NCFVSTVKTUIIWTE7UK2UTWTRZVJCQUECR3I3Q/build/release/staging/bin/hipcc line 500.
[  0%] Built target directed_tests.kernel.inline_asm_vadd

bensander:
LGTM.  @kknox - how do we get the CI results to run again?

AlexVlx:
@bensander I''d wait up a bit since I think I have a better solution that the one embodied in the latest commit (definitely less noisy). As for CI I think it automatically runs when the PR is updated.

PR #254, opened 2017-11-06
Event thread safe

bensander:
Events are now thread-safe using similar critical-section architecture as streams and contexts.  This is important to support cases where the same event is recorded multiple times, potentially from different threads.


bensander:
merging perf discussion @mangupta reviewed

PR #256, opened 2017-11-09
Texture driver APIs support

gargrahul:
First level support for texture driver APIs. Requires incremental support of further APIs.

### DIFF
--- a/include/hip/hcc_detail/hip_runtime_api.h
+++ b/include/hip/hcc_detail/hip_runtime_api.h
@@ -84,8 +84,6 @@ typedef struct ihipModule_t *hipModule_t;
 
 typedef struct ihipModuleSymbol_t *hipFunction_t;
 
-typedef void* hipDeviceptr_t;
-
 typedef struct ihipEvent_t *hipEvent_t;
 
 enum hipLimit_t
@@ -621,7 +619,7 @@ hipError_t hipStreamQuery(hipStream_t stream);
  *
  * This command is host-synchronous : the host will block until the specified stream is empty.
  *
- * This command follows standard null-stream semantics.  Specifically, specifying the null stream will cause the 
+ * This command follows standard null-stream semantics.  Specifically, specifying the null stream will cause the
  * command to wait for other streams on the same device to complete all pending operations.
  *
  * This command honors the hipDeviceLaunchBlocking flag, which controls whether the wait is active or blocking.
@@ -644,9 +642,9 @@ hipError_t hipStreamSynchronize(hipStream_t stream);
  * This function inserts a wait operation into the specified stream.
  * All future work submitted to @p stream will wait until @p event reports completion before beginning execution.
  *
- * This function only waits for commands in the current stream to complete.  Notably,, this function does 
- * not impliciy wait for commands in the default stream to complete, even if the specified stream is 
- * created with hipStreamNonBlocking = 0.  
+ * This function only waits for commands in the current stream to complete.  Notably,, this function does
+ * not impliciy wait for commands in the default stream to complete, even if the specified stream is
+ * created with hipStreamNonBlocking = 0.
  *
  * @see hipStreamCreate, hipStreamCreateWithFlags, hipStreamSynchronize, hipStreamDestroy
  */
@@ -756,7 +754,7 @@ hipError_t hipEventCreate(hipEvent_t* event);
  * If hipEventRecord() has been previously called on this event, then this call will overwrite any existing state in event.
  *
  * If this function is called on a an event that is currently being recorded, results are undefined - either
- * outstanding recording may save state into the event, and the order is not guaranteed.  
+ * outstanding recording may save state into the event, and the order is not guaranteed.
  *
  * @see hipEventCreate, hipEventCreateWithFlags, hipEventQuery, hipEventSynchronize, hipEventDestroy, hipEventElapsedTime
  *
@@ -1318,6 +1316,7 @@ hipError_t hipMallocArray(hipArray** array, const hipChannelFormatDesc* desc,
 hipError_t hipMallocArray(hipArray** array, const struct hipChannelFormatDesc* desc,
                           size_t width, size_t height, unsigned int flags);
 #endif
+hipError_t hipArrayCreate ( hipArray** pHandle, const HIP_ARRAY_DESCRIPTOR* pAllocateArray );
 /**
  *  @brief Frees an array on the device.
  *
@@ -1359,6 +1358,7 @@ hipError_t hipMalloc3DArray(hipArray_t *array,
  *  @see hipMemcpy, hipMemcpyToArray, hipMemcpy2DToArray, hipMemcpyFromArray, hipMemcpyToSymbol, hipMemcpyAsync
  */
 hipError_t hipMemcpy2D(void* dst, size_t dpitch, const void* src, size_t spitch, size_t width, size_t height, hipMemcpyKind kind);
+hipError_t hipMemcpy_2D(const hip_Memcpy2D* pCopy);
 
 /**
  *  @brief Copies data between host and device.
@@ -1968,6 +1968,7 @@ hipError_t hipModuleGetFunction(hipFunction_t *function, hipModule_t module, con
  */
 hipError_t hipModuleGetGlobal(hipDeviceptr_t *dptr, size_t *bytes, hipModule_t hmod, const char *name);
 
+hipError_t hipModuleGetTexRef(hipDeviceptr_t *dptr, hipModule_t hmod, const char* name);
 /**
  * @brief builds module from code object which resides in host memory. Image is pointer to that location.
  *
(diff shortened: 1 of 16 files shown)


include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -1359,6 +1358,7 @@ hipError_t hipMalloc3DArray(hipArray_t *array,
  *  @see hipMemcpy, hipMemcpyToArray, hipMemcpy2DToArray, hipMemcpyFromArray, hipMemcpyToSymbol, hipMemcpyAsync
  */
 hipError_t hipMemcpy2D(void* dst, size_t dpitch, const void* src, size_t spitch, size_t width, size_t height, hipMemcpyKind kind);
+hipError_t hipMemcpy_2D(const hip_Memcpy2D* pCopy);
```
bensander: What is the reason for both (subtly different) APIs ?

include/hip/hip_runtime_api.h
```diff
@@ -117,7 +117,9 @@ typedef struct hipDeviceProp_t {
  */
 enum hipMemoryType {
     hipMemoryTypeHost,   ///< Memory is physically located on host
-    hipMemoryTypeDevice  ///< Memory is physically located on device. (see deviceId for specific device)
+    hipMemoryTypeDevice,  ///< Memory is physically located on device. (see deviceId for specific device)
+    hipMemoryTypeArray,
```
bensander: Need doxygen comments for these types.

samples/2_Cookbook/11_texture_driver/tex2dKernel.cpp
```diff
@@ -0,0 +1,33 @@
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+#include "hip/hip_runtime.h"
+extern texture<float, 2, hipReadModeElementType> tex;
```
bensander: Is the "extern" only required on ROCm?

src/hip_memory.cpp
```diff
@@ -410,6 +410,113 @@ extern void getChannelOrderAndType(const hipChannelFormatDesc& desc,
...
+hipError_t hipArrayCreate ( hipArray** array, const HIP_ARRAY_DESCRIPTOR* pAllocateArray )
+{
+    HIP_INIT_SPECIAL_API((TRACE_MEM), array, pAllocateArray);
+    HIP_SET_DEVICE();
+    hipError_t  hip_status = hipSuccess;
+    if(pAllocateArray->width >0) {
+		auto ctx = ihipGetTlsDefaultCtx();
+
+		*array = (hipArray*)malloc(sizeof(hipArray));
+        array[0]->drvDesc = *pAllocateArray;
```
bensander: please review formatting - should replace tabs with spaces in all cases, indent 4

src/hip_memory.cpp
```diff
@@ -410,6 +410,113 @@ extern void getChannelOrderAndType(const hipChannelFormatDesc& desc,
                             hsa_ext_image_channel_order_t& channelOrder,
                             hsa_ext_image_channel_type_t& channelType);
```
bensander: Use pointers (rather than references) for variables that are to be written so the interface is clear.

src/hip_memory.cpp
```diff
@@ -410,6 +410,113 @@ extern void getChannelOrderAndType(const hipChannelFormatDesc& desc,
...
+			const unsigned am_flags = 0;
+			size_t size = pAllocateArray->width;
+			if(pAllocateArray->height > 0) {
+				size = size * pAllocateArray->height;
+			}
+            hsa_ext_image_channel_type_t channelType;
+			size_t allocSize = 0;
+			switch(pAllocateArray->format) {
+				case HIP_AD_FORMAT_UNSIGNED_INT8:
+					allocSize = size * sizeof(unsigned char);
```
bensander: use int8_t here

src/hip_memory.cpp
```diff
@@ -410,6 +410,113 @@ extern void getChannelOrderAndType(const hipChannelFormatDesc& desc,
...
+					channelType = HSA_EXT_IMAGE_CHANNEL_TYPE_UNSIGNED_INT8;
+					break;
+				case HIP_AD_FORMAT_UNSIGNED_INT16:
+					allocSize = size * sizeof(unsigned short);
```
bensander: int16_t

bensander: And more below - since the enumeration have specific size we should use the standard types with size encoded for these conversions.

src/hip_module.cpp
```diff
@@ -40,6 +40,8 @@ THE SOFTWARE.
 #include "hip_hcc_internal.h"
 #include "trace_helper.h"
 
+//TODO Add support for multiple modules
+static std::unordered_map<std::string, uintptr_t> coGlobals;
```
bensander: I know this is not new here but we need to remove this global or protect access from multiple threads.

include/hip/hcc_detail/hip_runtime_api.h
gargrahul: hipMemcpy2D is equivalent to cudaMemcpy2D  (CUDA Runtime API) whereas 
hipMemcpy_2D is equivalent to cuMemcpy2D (CUDA Driver API).

samples/2_Cookbook/11_texture_driver/tex2dKernel.cpp
gargrahul: Yes , "extern" only required on HIP/HCC path.

include/hip/hcc_detail/hip_runtime_api.h
bensander: This is pretty subtle.  Can we come up with a syntax that is more consistent with the rest of HIP - perhaps hipMemcpyParam2D to indicate this version uses copy parms inside a structure rather than as arguments?

gargrahul: Agree, we should use better name. Initially I used hipDrvMemcpy2D though wanted to avoid using "Drv" only at few places, so discarded that.

bensander:
Agree let’s avoid drv since hip does not have driver api.  Name after the parm?   Or is there consistent pattern?

On Nov 20, 2017, at 8:20 PM, gargrahul <notifications@github.com<mailto:notifications@github.com>> wrote:


@gargrahul commented on this pull request.

________________________________

In include/hip/hcc_detail/hip_runtime_api.h<https://github.com/ROCm-Developer-Tools/HIP/pull/256#discussion_r152164106>:

> @@ -1359,6 +1358,7 @@ hipError_t hipMalloc3DArray(hipArray_t *array,
  *  @see hipMemcpy, hipMemcpyToArray, hipMemcpy2DToArray, hipMemcpyFromArray, hipMemcpyToSymbol, hipMemcpyAsync
  */
 hipError_t hipMemcpy2D(void* dst, size_t dpitch, const void* src, size_t spitch, size_t width, size_t height, hipMemcpyKind kind);
+hipError_t hipMemcpy_2D(const hip_Memcpy2D* pCopy);


Agree, we should use better name. Initially I used hipDrvMemcpy2D though wanted to avoid using "Drv" only at few places, so discarded that.

bensander:
@gargrahul - I think just needs RELEASE notes and docs/markdown update and we can accept it.  In particular - any references that indicate textures are not supported.  Also note the need for "extern" in the porting guide.

tests/src/texture/hipTextureRef2D.cpp
```diff
@@ -14,7 +14,9 @@ texture<float, 2, hipReadModeElementType> tex;
 bool testResult = true;
 
 __global__ void tex2DKernel(float* outputData,
+#ifdef __HIP_PLATFORM_HCC__
```
bensander: @gargrahul - do we still need this different compilation paths on NVCC vs HCC ?  I thought with recent changes the only required change was "extern" before the texture global?

tests/src/texture/hipTextureRef2D.cpp
```diff
@@ -78,7 +80,7 @@ void runTest(int argc, char **argv)
 #ifdef __HIP_PLATFORM_HCC__
     hipLaunchKernelGGL(tex2DKernel, dim3(dimGrid), dim3(dimBlock), 0, 0, dData, tex.textureObject, width, height);
```
bensander: Same here, can we use global instead ?

gargrahul: These files shows examples for Texture runtime APIs. I updated the porting guide for runtime texture support too, therefore updated these files too for sync.

bensander: Some recent version of HCC is supposed to support platform globals as well so we should be able to use texture globals for this case as well.  Perhaps that already works?  If not - @AlexVlx which version of HCC supports this?

AlexVlx:
@bensander whatever comes after including this PR https://github.com/RadeonOpenCompute/hcc/pull/496 should work, but you must validate it for this use case first.

tests/src/texture/hipTextureRef2D.cpp
AlexVlx: @bensander also note that post #246 global support will no longer be tied to HCC, so you may want to take that into account and schedule validation for this bit in accordance.

gargrahul: Yes , post merging of #246 , I will update these runtime API example cases.

PR #266, opened 2017-11-15
Fixed half2 issue on gfx900

gargrahul:
Fixed hipTestHalf test example too.

### DIFF
--- a/bin/hipcc
+++ b/bin/hipcc
@@ -439,6 +439,7 @@ if($HIP_PLATFORM eq "hcc"){
     if ($target_gfx900 eq 1) {
         $HIPLDFLAGS .= " --amdgpu-target=gfx900";
         $HIPCXXFLAGS .= " -D__HIP_ARCH_GFX900__=1 ";
+        $ENV{HCC_EXTRA_LIBRARIES_GFX900}="$HIP_PATH/lib/hip_hc_gfx803.ll\n";
     }
 }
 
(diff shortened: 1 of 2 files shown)


bin/hipcc
```diff
@@ -439,6 +439,7 @@ if($HIP_PLATFORM eq "hcc"){
     if ($target_gfx900 eq 1) {
         $HIPLDFLAGS .= " --amdgpu-target=gfx900";
         $HIPCXXFLAGS .= " -D__HIP_ARCH_GFX900__=1 ";
+        $ENV{HCC_EXTRA_LIBRARIES_GFX900}="$HIP_PATH/lib/hip_hc_gfx803.ll\n";
```
whchung: since it's now used by `gfx900` as well, should this handcrafted `hip_hc_gfx803.ll` carry a better name?

gargrahul:
*@whchung* we can create a copy of hip_hc_gfx803.ll and name that as
hip_hc_gfx900.ll, though I thought of reusing existing ll. Its time we
should get rid of these ll files dependencies.

tests/src/deviceLib/hipTestHalf.cpp
```diff
@@ -19,13 +19,14 @@ THE SOFTWARE.
...
-#include "hip/hip_runtime_api.h"
+#include "hip/hip_runtime.h"
+#include "test_common.h"
 
 #define LEN 64
 #define HALF_SIZE 64*sizeof(__half)
 #define HALF2_SIZE 64*sizeof(__half2)
 
-#if __HIP_ARCH_GFX803__ > 0
+#if __HIP_ARCH_GFX803__ || __HIP_ARCH_GFX803__
```
mangupta: Do you mean GFX900 here? Actually the original line i.e. ```_HIP_ARCH_GFX803 > 0``` might still work correctly.

gargrahul:
*@mangupta* original line didn't work. I have fixed it for GFX900.

bensander:
@mangupta - can you review and merge if it looks good?

PR #267, opened 2017-11-17
Fix curious bitwise conversions.

AlexVlx:
This fixes some outright quaint choices made when implementing HIP's bitwise conversion functions, by using simple `reinterpret_cast`s, as is idiomatic. These functions are supposed to be re-entrant, correct and efficient. Sadly, they were neither: they hid a massive race condition against a value stored in global memory, which means that they were also unreasonably slow if they ever managed to be correct, and relied on union based type punning which is in a grey area of the standard. It is difficult to ascertain what may have been the reason for coming up with this quirky solution.


bensander:
@adityaaatluri - can you review ?

PR #270, opened 2017-11-19
Use natural CUDA indexing

AlexVlx:
This implements the trivial change needed to move back from the hip{Something}_{x, y, z} macros to the natural CUDA syntax of Something.{x, y, z}. This is contained in lines 384-404 in hip_runtime.h. All of the other changes have to do with moving unit tests to use this syntax. The macros are retained for backwards compatibility.

### DIFF
--- a/include/hip/hcc_detail/hip_runtime.h
+++ b/include/hip/hcc_detail/hip_runtime.h
@@ -381,6 +381,27 @@ __device__ void  __threadfence_system(void) ;
  * @}
  */
 
+template<typename std::common_type<
+    decltype(hc_get_group_id),
+    decltype(hc_get_group_size),
+    decltype(hc_get_num_groups),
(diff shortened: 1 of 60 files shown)


include/hip/hcc_detail/hip_runtime.h
```diff
@@ -381,6 +381,27 @@ __device__ void  __threadfence_system(void) ;
...
+    struct X { __device__ operator R() const { return f(0); } };
+    struct Y { __device__ operator R() const { return f(1); } };
+    struct Z { __device__ operator R() const { return f(2); } };
+public:
+    static constexpr X x{};
+    static constexpr Y y{};
+    static constexpr Z z{};
+};
+
+static constexpr Coordinates<hc_get_group_size> blockDim;
```
t-tye: Does making this static lead to ODR issues since each compilation init that includes the header will get a different version of these global symbols?

PR #271, opened 2017-11-21
Cppcheck

bensander:
Various changes to fix cpp-check errors.

### DIFF
--- a/src/hip_device.cpp
+++ b/src/hip_device.cpp
@@ -446,14 +446,14 @@ hipError_t hipChooseDevice( int* device, const hipDeviceProp_t* prop )
 {
     HIP_INIT_API(device,prop);
     hipDeviceProp_t  tempProp;
-    int deviceCount;
-    int inPropCount = 0;
-    int matchedPropCount = 0;
     hipError_t e = hipSuccess;
     if((device == NULL) || (prop == NULL)) {
         e = hipErrorInvalidValue;
     }
     if(e == hipSuccess) {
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -147,6 +134,7 @@ endif()
 # Build hip_hcc if platform is hcc
 if(HIP_PLATFORM STREQUAL "hcc")
     include_directories(${PROJECT_SOURCE_DIR}/include)
+    include_directories(/opt/rocm/include)
```
pfultz2: I dont think hip should hard-code the include path. This will cause problems when we want to use another hcc that is not installed in /opt/rocm. Ultimately, hip should just use `find_package(hcc PATHS /opt/rocm)` to get hcc's usage requirements.

include/hip/hcc_detail/grid_launch_GGL.hpp
```diff
@@ -27,7 +27,7 @@ THE SOFTWARE.
 #include "concepts.hpp"
 #include "helpers.hpp"
 
-#include "hc.hpp"
+#include "hcc/hc.hpp"
```
mangupta: The problem with this is if you build and install hcc from source using ```cmake <...>; make; make install```, hcc headers are installed in <hcc_install_root>/include and not <hcc_install_root>/include/hcc. Hence this change breaks building hip unless hcc is installed using deb or rpm packages.

PR #283, opened 2017-12-05
 [HIPIFY] Disable cudaBuiltin matcher.

emankov:
As HIP has started to support vanilla CUDA syntax for threadIdx, blockIdx, blockDim and gridDim.
Other CUDA builtins are not tracked for now.


hipify-clang/src/HipifyAction.cpp
```diff
@@ -371,7 +373,7 @@ std::unique_ptr<clang::ASTConsumer> HipifyAction::CreateASTConsumer(clang::Compi
         ).bind("cudaBuiltin"),
         this
     );
-
+    */
```
bensander: Let's just remove the dead code rather than comment it - we can always retrieve from Git revision history if desired.

emankov: Done, verified.

emankov:
Failed a few tests (not hipify-clang) - it is not related to this change.

PR #288, opened 2017-12-08
Purge hsagenco.sh

AlexVlx:
This introduces LipoProteinLipase (lpl), a simple tool for creating fat binaries. It represents a direct replacement of the creaky hccgenco.sh script, which had various issues. The format it uses is that of a code object bundle, generated by the Clang Offload Bundler. The output is always suffixed with the ".adipose" extension. It is shared with HCC. The hipcc script and associated tests are modified to use lpl. Help can be obtained by invoking lpl --help. A more computer-sciency / corporate friendly name is likely to be beneficial, which is a reason for choosing easily searchable/replaceable names such as lpl or adipose. An associated tool which can decompose the fat binary into its constituent code objects, and is called Cathecolamine, will be proposed separately.


mangupta:
@gargrahul Can you look into this PR? As per your previous tests on your local setup everything was working fine. However CI fails with several directed tests consistently failing.

PR #299, opened 2017-12-19
Implement hipStreamAddCallback

mangupta:


tests/src/runtimeApi/stream/hipStreamAddCallback.cpp
```diff
@@ -23,41 +23,69 @@ THE SOFTWARE.
  * HIT_END
  */
 
-// Test under-development. Call hipStreamAddCallback function and see if it works as expected.
-
+#include <stdio.h>
+#include <unistd.h>
```
emankov: unistd.h is a POSIX header. With supporting Windows in the future it should be replaced with one or multiple C/C++ headers and also supplemented with platform ifdefs, for instance:

```c++
+#ifdef _WIN32
+#include <windows.h>
+#define sleep(x) Sleep(x)
+#else
+#include <unistd.h>
+#endif 
```

emankov: It looks like unistd.h is added for sleep(1), so why don't replace it with STL's ones: 

```c++
std::this_thread::sleep_for(std::chrono::milliseconds(1000));
```

mangupta: @emankov: Updated the code to use sleep_for instead of unistd sleep.

src/hip_hcc.cpp
```diff
@@ -1409,9 +1410,38 @@ void ihipInit()
...
+    if (stream == hipStreamNull) {
+        ihipCtx_t *ctx = ihipGetTlsDefaultCtx();
+        ctx->locked_syncDefaultStream(true/*waitOnSelf*/, true/*syncToHost*/);
+    } else {
+        // note this does not synchornize with the NULL stream:
+        stream->locked_wait();
+        e = hipSuccess;
+    }
 
+    return e;
```
emankov: It looks like e is always hipSuccess.

src/hip_hcc.cpp
```diff
@@ -1409,9 +1410,38 @@ void ihipInit()
...
+{
+    hipError_t e = hipSuccess;
+
+    // Notify hipStreamAddCallback that callback handler thread is active
+    std::lock_guard<std::mutex> guard(cb->_mtx);
+    cb->_ready = true;
+
+    // Synchronize stream
+    tprintf(DB_SYNC, "ihipStreamCallbackHandler wait on stream %s\n", ToString(cb->_stream).c_str());
+    e = ihipStreamSynchronize(cb->_stream);
```
emankov: e might be initialized here.

tests/src/runtimeApi/stream/hipStreamAddCallback.cpp
```diff
@@ -23,41 +23,65 @@ THE SOFTWARE.
...
-};
+float *A_h, *C_h;
+size_t N = 1000000;
 
-void HIPRT_CB CallbackClass::Callback(hipStream_t stream, hipError_t status, void *userData)
+static void HIPRT_CB Callback(hipStream_t stream, hipError_t status, void *userData)
 {
-    CallbackClass* obj = (CallbackClass*) userData;
-    obj->callbackFunc(status);
+    for (size_t i=0; i<N; i++) {
```
emankov: ++i

tests/src/runtimeApi/stream/hipStreamAddCallback.cpp
```diff
@@ -23,41 +23,65 @@ THE SOFTWARE.
...
+    float *A_d, *C_d;
+    size_t Nbytes = N * sizeof(float);
+
+    A_h = (float*)malloc(Nbytes);
+    HIPCHECK(A_h == 0 ? hipErrorMemoryAllocation : hipSuccess );
+    C_h = (float*)malloc(Nbytes);
+    HIPCHECK(C_h == 0 ? hipErrorMemoryAllocation : hipSuccess );
+
+    // Fill with Phi + i
+    for (size_t i=0; i<N; i++)
```
emankov: ++i

src/hip_hcc.cpp
mangupta: ihipStreamSynchronize only returns when the stream is synchronized. So it will return hipSuccess. In To be consistent with the rest of the code, instead of directly using return hipSuccess, we initialize e to hipSuccess and then return e.

src/hip_hcc.cpp
mangupta: I am not sure what you mean by initialized here. Are you saying that we shouldn't initialize e to hipSuccess at the beginning of the function? It is a common practice in the rest of the code.

tests/src/runtimeApi/stream/hipStreamAddCallback.cpp
mangupta: Array index starts at 0. So i++ is correct.

tests/src/runtimeApi/stream/hipStreamAddCallback.cpp
mangupta: Array index starts at 0. So i++ is correct.

tests/src/runtimeApi/stream/hipStreamAddCallback.cpp
emankov: It is correct, but not because of index starting value - it doesn't matter. i++ reads the value of i then increments it. ++i increments the value of i then reads it. It is about performance, not correctness. Correctness is in case like following: for(int i=0, a = 0; i<10; a = i++), where result of iteration is used in place. Here i++ means increment, but evaluate to the previous value, but ++i means increment, but evaluate to the next value. As a result a will get different values (i when i++ and i+1 when ++i).

src/hip_hcc.cpp
emankov: Yes, it should be initialized where first assigned, as it is not used anywhere before. It just eliminates 1 unneeded assignment (performance).

src/hip_hcc.cpp
emankov: It always returns hipSuccess.

src/hip_hcc.cpp
mangupta: Firstly it is just that it is good coding practice to initialize the variable. Also I think the compiler will optimize this anyways.

tests/src/runtimeApi/stream/hipStreamAddCallback.cpp
mangupta: This is test code. And the for loop is for comparing the results generated by host and device. So it is not a performance critical piece.

PR #301, opened 2017-12-20
Return pass on single gpu in hipPeerToPeer_simple

gargrahul:


tests/src/p2p/hipPeerToPeer_simple.cpp
```diff
@@ -397,32 +397,30 @@ int main(int argc, char *argv[])
     if (gpuCount < 2)
     {
         printf("P2P application requires atleast 2 gpu devices\n");
-        return 0;
```
mangupta: Isn't it sufficient to just add a ```passed()``` before ```return 0```?

gargrahul:
Yes that is possible though used passed() only once in the end.

PR #304, opened 2017-12-20
Upgrade to nvidia-docker2 for CUDA testing

kknox:
This enables us to remove the driver version from startup string


kknox:
I may merge this PR in without formal review (since most are away for holiday).  This change only affects nvidia based builds & test, and increases stability by removing the need to explicitly list driver versions.

Issue #306, opened 2017-12-26
[HIPIFY] Second cuda main include is not hipified


emankov:
For instance:
```c++
// hipified to #include<hip/hip_runtime.h>
#include<cuda.h> // 1st cuda main include (Driver API)
// leaved unchanged
#include<cuda_runtime.h> // 2nd main include (Runtime API)
```
HIP has one header for both APIs hip_runtime.h, thus second cuda main include have to be eliminated.

PR #307, opened 2017-12-26
[HIPIFY][FIX][#306] Eliminate second cuda main include directive

emankov:
// hipified to #include<hip/hip_runtime.h>
#include<cuda.h> // 1st cuda main include (Driver API)
// to eliminate
#include<cuda_runtime.h> // 2nd cuda main include (Runtime API)

HIP has one header hip_runtime.h for both CUDA APIs, thus second cuda main include directive is eliminated entirely.

### DIFF
--- a/hipify-clang/src/HipifyAction.cpp
+++ b/hipify-clang/src/HipifyAction.cpp
@@ -157,11 +157,11 @@ void HipifyAction::InclusionDirective(clang::SourceLocation hash_loc,
     }
 
     // Special-casing to avoid duplication of the hip_runtime include.
+    bool secondMainInclude = false;
     if (found->second.hipName == "hip/hip_runtime.h") {
         if (insertedRuntimeHeader) {
-            return;
+          secondMainInclude = true;
         }
-
         insertedRuntimeHeader = true;
     }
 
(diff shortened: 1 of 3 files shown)


emankov:
ping

hipify-clang/src/HipifyAction.cpp
```diff
@@ -157,34 +157,40 @@ void HipifyAction::InclusionDirective(clang::SourceLocation hash_loc,
...
     clang::SourceLocation sl = filename_range.getBegin();
     if (found->second.unsupported) {
-        // An unsupported CUDA header? Oh dear. Print a warning.
         clang::DiagnosticsEngine& DE = getCompilerInstance().getDiagnostics();
         DE.Report(sl, DE.getCustomDiagID(clang::DiagnosticsEngine::Warning, "Unsupported CUDA header"));
         return;
     }
 
-    const char *B = SM.getCharacterData(sl);
+    char *B = nullptr;
```
dfukalov: you can move this `const char *B = SM.getCharacterData(sl);` next to `if` statement and avoid `const_cast` below (by removing both `B` assigns in lines 183 and 192)

PR #310, opened 2018-01-09
[HIPIFY][tests] Add Windows testing support

emankov:


tests/hipify-clang/run_test.bat
```diff
@@ -0,0 +1,17 @@
+@echo off
+
```
dfukalov: it possible would be good idea to add here `setlocal` to isolate variables that are set in the batch.
overwise LGTM

Issue #211, opened 2017-10-17
[HIPIFY] Inserted hip include should ideally be after include guard


ChrisKitching:
When hipify inserts the `#include <hip/hip_runtime.h>` line into a header, it would be best to do it after any existing `#pragma once` or traditional include guard. Saves the user a bit of tedious cleanup.

One elegant way to implement this might be to replace the current insertion behaviour with logic to _replace_ appropriate CUDA `#include` lines with hip ones. Unilaterally inserting this include into every file which you've made changes to is probably overkill - isn't it enough to replace `#include <cuda>` with the hip include?

emankov:
It was done so intentionally, because cuda header might be absent at all in a source and NVCC correctly compile such source. But this particular case doesn't negate the necessity of correctly handling of other cases. So the correct behaviour looks like as following:
1. Replace existing  #include <cuda> or #include <cuda_runtime.h>, m.b. others (taking into accout multiple includes).
2. Insert #include <hip/hip_runtime.h> on an appropriate line in case of absence of cuda header includes.

Chris, I've just assigned the issue on myself.

ChrisKitching:
> because cuda header might be absent at all in a source and NVCC correctly compile such source.

Oh right. I forgot about this - NVCC implicitly includes a bunch of cuda stuff, doesn't it? (and clang does the same in cuda mode for compatibility). That explains it.

emankov:
The last case to cover.
`#ifndef #define ... #endif` guard for the entire file. In this case HIP main include should be placed just after `#ifndef`, or after `#pragma once`, if it is also presented.

emankov:
Fixed by #1282

PR #319, opened 2018-01-15
[HIPIFY][fix][#211] Algorithm for explicit insert of hip include directive

emankov:
If in source CUDA file main header (cuda_runtime.h or cuda.h) is not presented, corresponding HIP main header (hip_runtime.h) should be explicitly included in output hipified file.

[Algorithm]
1. If #pragma once is presented, HIP main header should be placed just after it;
2. Otherwise if any other (not CUDA main) header is presented, HIP main header should be placed just before it;
3. Otherwise HIP main header should be placed in the beginning of output file.

P.S.
There might be one more situation when #ifndef #define ... #endif guard for the entire file is presented (make sense for *.h, *.hpp, *.cuh files). In this case HIP main include should be placed just after such #ifdef, or after #pragma once, if it is also presented. This situation will be handled in a separate change.

### DIFF
--- a/tests/hipify-clang/headers_test_03.cu
+++ b/tests/hipify-clang/headers_test_03.cu
@@ -1,0 +1,10 @@
+// RUN: %run_test hipify "%s" "%t" %cuda_args
+
+// CHECK: #pragma once
+// CHECK-NEXT: #include <hip/hip_runtime.h>
+#pragma once
+// CHECK-NOT: #include<hip/hip_runtime.h>
+int main(int argc, char* argv[]) {
+  return 0;
+}
+
--- a/tests/hipify-clang/headers_test_03.cu
+++ b/tests/hipify-clang/headers_test_03.cu
@@ -3,7 +3,7 @@
 // CHECK: #pragma once
 // CHECK-NEXT: #include <hip/hip_runtime.h>
 #pragma once
-// CHECK-NOT: #include<hip/hip_runtime.h>
+// CHECK-NOT: #include <hip/hip_runtime.h>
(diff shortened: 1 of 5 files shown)


tests/hipify-clang/headers_test_03.cu
```diff
@@ -0,0 +1,10 @@
+// RUN: %run_test hipify "%s" "%t" %cuda_args
+
+// CHECK: #pragma once
+// CHECK-NEXT: #include <hip/hip_runtime.h>
+#pragma once
+// CHECK-NOT: #include<hip/hip_runtime.h>
```
dfukalov: should it be a space after `#include`? possible it should be FILECHECK regex for a case hard-coded string would be changed?

tests/hipify-clang/headers_test_04.cu
```diff
@@ -0,0 +1,12 @@
+// RUN: %run_test hipify "%s" "%t" %cuda_args
+
+// CHECK: #include <hip/hip_runtime.h>
+// CHECK-NEXT: #include <stdio.h>
+// CHECK-NEXT: #include <iostream>
+#include <stdio.h>
+#include <iostream>
+// CHECK-NOT: #include<hip/hip_runtime.h>
```
dfukalov: ditto

tests/hipify-clang/headers_test_03.cu
emankov: It doesn't matter, as FileCheck by default canonicalizes input horizontal whitespace (spaces and tabs). It matters only with --strict-whitespace option, which we do not set.

PR #320, opened 2018-01-15
Fix hcc ctu build

kknox:
Getting jenkins back to green builds


kknox:
The one failing test is [hcc_ctu]: 
90 - directed_tests/kernel/hipPrintfKernel.tst (Timeout)

which is an old, recurring failure

PR #321, opened 2018-01-16
Added support for hipMemcpy Array functions-

gargrahul:
- hipMemcpyFromArray
- hipMemcpyAtoH
- hipMemcpyHtoA

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -1411,6 +1411,65 @@ hipError_t hipMemcpyToArray(hipArray* dst, size_t wOffset, size_t hOffset,
     return ihipLogStatus(e);
 }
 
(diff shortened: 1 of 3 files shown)


src/hip_memory.cpp
```diff
@@ -1411,6 +1411,65 @@ hipError_t hipMemcpyToArray(hipArray* dst, size_t wOffset, size_t hOffset,
     return ihipLogStatus(e);
 }
 
+hipError_t hipMemcpyFromArray(void* dst, hipArray_const_t srcArray, size_t wOffset, size_t hOffset,
```
mangupta: The implementation of the 3 functions is nearly identical. They could be implemented using a single ihipMemcpyFromArray function that takes ```kind``` with appropriate values being passed for it in the 3 functions.

PR #342, opened 2018-01-23
temporarily disable the printf test

scchan:
until stability issue has been resolved

PR #345, opened 2018-01-24
Fixing rocblas build failure

kknox:
with ::Bundled_code_header constructor

Disabling hipPrintfKernel test from CI

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -151,7 +151,7 @@ def docker_build_inside_image( def build_image, String inside_args, String platf
       // The rm command needs to run as sudo because the test steps below create files owned by root
       sh  """#!/usr/bin/env bash
           set -x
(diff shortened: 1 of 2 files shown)


scchan:
I have a pending PR (https://github.com/ROCm-Developer-Tools/HIP/pull/342) to turn off the printf test by asking hip's test to skip it but perhaps your change in the Jenkins script is a better to achieve that.  I'll close my PR.

kknox:
@scchan 
Can you look at the build failure for build #8 hcc-ctu on this PR; I think the problem is unrelated to my PR, I'm thinking the latest hcc-ctu broke hip.

scchan:
@kknox I wasn't able to reproduce the crash on my machine and it looks build #9 is passing

kknox:
Thx, i'm looking into that particular machine

PR #349, opened 2018-01-29
[HIPIFY] InclusionDirective refactoring

emankov:
Due to support of cuRAND headers.

+ compound test on all headers is added;
+ missing entities are added with updating the doc;
+ a couple cuRAND tests are added (https://github.com/ROCmSoftwarePlatform/rocRAND/tree/master/benchmark):
  - the following CUDA entities are still unsupported by hipRAND:
      curandMakeMTGP32Constants
      curandMakeMTGP32KernelState
      curandGetDirectionVectors32
      curandDirectionVectorSet_t
      CURAND_DIRECTION_VECTORS_32_JOEKUO6
      curandStateSobol64_t
      curandStateScrambledSobol64_t
      curandGenerateLongLong
  - and the following - by HIP:
      cudaRuntimeGetVersion
  - those entities are handled by CHECK-NOT directive for now.

### DIFF
--- a/tests/hipify-clang/cuRAND/benchmark_curand_generate.cpp
+++ b/tests/hipify-clang/cuRAND/benchmark_curand_generate.cpp
@@ -1,0 +1,393 @@
+// RUN: %run_test hipify "%s" "%t" %cuda_args
+
+// Copyright (c) 2017 Advanced Micro Devices, Inc. All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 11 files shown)


tests/hipify-clang/cuRAND/benchmark_curand_generate.cpp
```diff
@@ -0,0 +1,393 @@
...
+    // CHECK: CUDA_CALL(hipDeviceSynchronize());
+    CUDA_CALL(cudaDeviceSynchronize());
+    auto end = std::chrono::high_resolution_clock::now();
+    std::chrono::duration<double, std::milli> elapsed = end - start;
+
+    std::cout << std::fixed << std::setprecision(3)
+              << "      "
+              << "Throughput = "
+              << std::setw(8) << (trials * size * sizeof(T)) /
+                    (elapsed.count() / 1e3 * (1 << 30))
```
dfukalov: it seems `elapsed.count()` can possible return zero so it may be div/0 here and at line 110?

emankov: You'd better file an issue to [rocRAND](https://github.com/ROCmSoftwarePlatform/rocRAND/issues), as those benchmarks are taken from there.

jszuppe: It's highly unlikely that `elapsed.count()` will be zero (IIRC, on most platforms `std::chrono::high_resolution_clock` is steady), and even if that happens the expression will evaluate to `inf`. I don't think that's a problem, because if  `elapsed.count()` is zero, then the result is useless anyway and something must be wrong.

emankov: Well, I guess this might be merged as is for now.

PR #371, opened 2018-03-14
[ci] Continue testing against ROCm 1.6

mangupta:
Continue testing against ROCm 1.6.4 until we have a stable ROCm 1.7.x docker image


VincentSC:
Thanks for re-adding this test! 
Please continue testing for at least a year, as 1.6 is see as 'stable' and there are systems out there that cannot brought down that easy.

PR #376, opened 2018-03-19
correct use of constant address space as 2, it is 4

david-salinas:
With the latest upstream address space changes, the constant address space has change from 2 to 4.  The Texture tests have 2 hard coded.


mangupta:
Replaced by #382 which passed CI tests correctly. So merging that instead.

PR #388, opened 2018-04-04
Add support for clang offload bundles and <<<>>> kernel launch.

lmoriche:


gargrahul:
@lmoriche , can you add a test for this too?

lmoriche:
@gargrahul , all tests are applicable, but we need to compile them with clang

PR #394, opened 2018-04-11
hipMemset refactoring

gargrahul:
hipMemset refactoring

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -1546,7 +1546,7 @@ typedef enum ihipMemsetDataType {
     ihipMemsetDataTypeChar   = 0,
     ihipMemsetDataTypeShort  = 1,
     ihipMemsetDataTypeInt    = 2
-}ihipMemsetCopyDataType;
+}ihipMemsetDataType;
 
 hipError_t ihipMemset(void* dst, int  value, size_t sizeBytes, hipStream_t stream, enum ihipMemsetDataType copyDataType  )
 {
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -1542,6 +1542,63 @@ void ihipMemsetKernel(hipStream_t stream, T* ptr, T val, size_t sizeBytes) {
                        sizeBytes, std::move(val));
 }
 
+typedef enum ihipMemsetCopyDataType {
+    ihipMemsetCopyDataTypeChar   = 0,
+    ihipMemsetCopyDataTypeShort  = 1,
+    ihipMemsetCopyDataTypeInt    = 2
+}ihipMemsetCopyDataType;
+
+hipError_t ihipMemset(void* dst, int  value, size_t sizeBytes, hipStream_t stream, enum ihipMemsetCopyDataType copyDataType  )
+{
+    hipError_t e = hipSuccess;
+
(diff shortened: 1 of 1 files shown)


src/hip_memory.cpp
```diff
@@ -1542,44 +1542,73 @@ void ihipMemsetKernel(hipStream_t stream, T* ptr, T val, size_t sizeBytes) {
                        sizeBytes, std::move(val));
 }
 
-
-// TODO-sync: function is async unless target is pinned host memory - then these are fully sync.
-hipError_t hipMemsetAsync(void* dst, int value, size_t sizeBytes, hipStream_t stream) {
-    HIP_INIT_SPECIAL_API((TRACE_MCMD), dst, value, sizeBytes, stream);
-
+typedef enum ihipMemsetCopyDataType {
```
mangupta: ihipMemsetCopyDataType seems a very odd name. Isn't ihipMemsetDataType sufficient? In case you have plans to share this across memcpy routines as well, maybe we can use ihipMemOpDataType or something like that.

src/hip_memory.cpp
```diff
@@ -1542,44 +1542,73 @@ void ihipMemsetKernel(hipStream_t stream, T* ptr, T val, size_t sizeBytes) {
...
+    ihipMemsetCopyDataTypeChar   = 0,
+    ihipMemsetCopyDataTypeShort  = 1,
+    ihipMemsetCopyDataTypeInt    = 2
+}ihipMemsetCopyDataType;
+
+hipError_t ihipMemset(void* dst, int  value, size_t sizeBytes, hipStream_t stream, enum ihipMemsetCopyDataType copyDataType  )
```
mangupta: I don't see any calls to ihipMemset with the copyDataType being set to Short or Int in the rest of the code. Are those changes coming in later?

src/hip_memory.cpp
gargrahul: I can change it to ihipMemsetDataType.

src/hip_memory.cpp
gargrahul: Yes, those changes will be coming today with support of hipMemsetD16/32 variants.

src/hip_memory.cpp
mangupta: Changing it to ihipMemsetDataType will be better

src/hip_memory.cpp
```diff
@@ -1542,44 +1542,73 @@ void ihipMemsetKernel(hipStream_t stream, T* ptr, T val, size_t sizeBytes) {
...
-
-// TODO-sync: function is async unless target is pinned host memory - then these are fully sync.
-hipError_t hipMemsetAsync(void* dst, int value, size_t sizeBytes, hipStream_t stream) {
-    HIP_INIT_SPECIAL_API((TRACE_MCMD), dst, value, sizeBytes, stream);
-
+typedef enum ihipMemsetDataType {
+    ihipMemsetDataTypeChar   = 0,
+    ihipMemsetDataTypeShort  = 1,
+    ihipMemsetDataTypeInt    = 2
+}ihipMemsetCopyDataType;
```
mangupta: should be ihipMemsetDataType

PR #396, opened 2018-04-12
[ci] Renable testing against nvcc

mangupta:


mangupta:
Closing this since we have modified the CI flow significantly and this will have to be redone.

PR #398, opened 2018-04-13
Fix missing ihipLogStatus

pzins:


gargrahul:
@pzins I have fixed this in #400. Ideally we should not call hip API from a hip API.

mangupta:
Thanks for submitting this change. However abandoning this since this change was included in a different patch that was fixing this in a cleaner way.

PR #399, opened 2018-04-13
Fix bug when using older rocr naming

aaronenyeshi:
This change will allow HIP to run with old ROCR and new ROCR naming conventions.

### DIFF
--- a/src/code_object_bundle.cpp
+++ b/src/code_object_bundle.cpp
@@ -14,23 +14,37 @@ hsa_isa_t hip_impl::triple_to_hsa_isa(const std::string& triple) {
     static constexpr const char OffloadKind[] = "hcc-";
     hsa_isa_t Isa = {0};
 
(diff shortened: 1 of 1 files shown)


gargrahul:
@mangupta/ @aaronenyeshi  , is it working? I am seeing issue on system with older rocr.

mangupta:
You need the corresponding patch from hcc as well.
-Maneesh

gargrahul:
@mangupta, even while using old rocr and old HCC ?

mangupta:
Combinations that should work I believe:

  *   Old rocr + old hip + old hcc
  *   Old rocr + new hip + new hcc
  *   New rocr + new hip + new hcc

gargrahul:
I think we should make following combination working too as part of backward compatibility.

Old rocr + New hip + Old hcc

gargrahul:
Thanks @aaronenyeshi

PR #401, opened 2018-04-16
Fix hipMemsetAsync() for NULL and zero inputs

luckynikki:
hipMemsetAsync() gives hipSuccess and also Aborted(core dumped) when the first parameter is passed as NULL or 0. Ran all the tests in memory module under directed tests  in my local machine and found  no issues.


gargrahul:
@luckynikki you need to re-base to latest code and put dest NULL check in ihipMemset().

gargrahul:
@luckynikki I have fixed this as part of #404

PR #403, opened 2018-04-16
Clean up triple_to_hsa_isa gfx naming

aaronenyeshi:
Clean up the triple_to_hsa_isa code and fix new case: using old HCC/HIP generated code on new HCC RT.


aaronenyeshi:
This is passing all HIP unit tests, and also passing few HIP samples I ran.

PR #405, opened 2018-04-18
Fix null check in hipMemsetAsync API

luckynikki:
hipMemsetAsync() gives hipSuccess and also Aborted(core dumped) when the first parameter is passed as NULL or 0. Ran all the tests in memory module under directed tests in my local machine and found no issues.

As suggested earlier, I made changes in ihipMemset().

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -1551,6 +1551,11 @@ typedef enum ihipMemsetDataType {
 hipError_t ihipMemset(void* dst, int  value, size_t sizeBytes, hipStream_t stream, enum ihipMemsetDataType copyDataType  )
 {
     hipError_t e = hipSuccess;
+	if(dst==NULL || dst==0)
+        {
+         e=hipErrorInvalidValue;
+        return ihipLogStatus(e);
+        }
 
     if (stream) {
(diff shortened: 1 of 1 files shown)


gargrahul:
@luckynikki this is fixed in #404 , is that not enough or working?

src/hip_memory.cpp
```diff
@@ -1551,6 +1551,11 @@ typedef enum ihipMemsetDataType {
 hipError_t ihipMemset(void* dst, int  value, size_t sizeBytes, hipStream_t stream, enum ihipMemsetDataType copyDataType  )
 {
     hipError_t e = hipSuccess;
+	if(dst==NULL || dst==0)
```
mangupta: Shouldn't dst==NULL be sufficient?

src/hip_memory.cpp
```diff
@@ -1551,6 +1551,11 @@ typedef enum ihipMemsetDataType {
...
+        {
+         e=hipErrorInvalidValue;
+        return ihipLogStatus(e);
```
mangupta: ihipMemset should just return e and not ihipLogStatus(e).

mangupta:
Since the fix has already been merged via #404, closing this PR.

PR #407, opened 2018-04-19
Fix Atomics test misplaced passed function

phani544:
passed was misplaced earlier, so test failure was never reported.

### DIFF
--- a/tests/src/deviceLib/hipSimpleAtomicsTest.cpp
+++ b/tests/src/deviceLib/hipSimpleAtomicsTest.cpp
@@ -241,7 +241,12 @@ int main(int argc, char** argv) {
 
     hipDeviceReset();
     printf("%s completed, returned %s\n", sampleName, testResult ? "OK" : "ERROR!");
-    exit(testResult ? EXIT_SUCCESS : EXIT_FAILURE);
+    
+    if (testResult) {
+      passed();
+    } else {
+      exit(EXIT_FAILURE);
+    }
 }
 
 void runTest(int argc, char** argv) {
@@ -292,5 +297,5 @@ void runTest(int argc, char** argv) {
     free(hOData);
     hipFree(dOData);
 
-    passed();
+    //passed();
 }


whchung:
@phani544 @mangupta @bensander To prevent future incidents I'd like to recommend a review on all existing HIP tests to make sure no such incorrect control flow take place. The bug was hidden and took the team working on TensorFlow literally 2 months to figure out.

tests/src/deviceLib/hipSimpleAtomicsTest.cpp
```diff
@@ -242,6 +242,7 @@ int main(int argc, char** argv) {
     hipDeviceReset();
     printf("%s completed, returned %s\n", sampleName, testResult ? "OK" : "ERROR!");
     exit(testResult ? EXIT_SUCCESS : EXIT_FAILURE);
+    passed();
```
mangupta: This doesn't look right. The previous statement is exit(). So this piece of code is going to be unreachable.

tests/src/deviceLib/hipSimpleAtomicsTest.cpp
```diff
@@ -292,5 +293,5 @@ void runTest(int argc, char** argv) {
     free(hOData);
     hipFree(dOData);
 
-    passed();
+    //passed();
```
mangupta: The right way is to conditionally execute passed() based on the value of testResult.

mangupta:
@phani544 Can you address the review comments and update the PR?

mangupta:
@phani544 When can you update the PR based on the review comments?

phani544:
Hi Maneesh, I have amended the change as per comments. This worked in my local machines. Please review.

whchung:
@phani544 The PR has gone stale. Could you help revise it? This PR is crucial to ensure the correctness of atomic functions for quite a few HIP applications.

mangupta:
Dropping this since this requires rework which can be done only after CI fixes are in place.

PR #408, opened 2018-04-20
SWDEV-141661

luckynikki:
Added a fix for the issue ::hipMemcpy() gives seg. fault when the first or second parameter is passed as NULL or zero. Verified in the local machine the fail scenario and also ran the directed tests. Things are fine.

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -1125,6 +1125,14 @@ hipError_t hipMemcpy(void* dst, const void* src, size_t sizeBytes, hipMemcpyKind
 
     hipError_t e = hipSuccess;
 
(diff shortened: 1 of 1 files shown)


src/hip_memory.cpp
```diff
@@ -1125,6 +1125,14 @@ hipError_t hipMemcpy(void* dst, const void* src, size_t sizeBytes, hipMemcpyKind
 
     hipError_t e = hipSuccess;
 
+    if(dst==NULL || dst==0 || src==NULL || src==0)
```
mangupta: HIP runtime is compiled using a C++ compiler. So comparison to both NULL as well as 0 does not make sense. Just comparing to NULL is sufficient.

mangupta:
Closing in favor of #422

PR #409, opened 2018-04-20
SWDEV-141618

luckynikki:
Fixed issue for :::hipMallocPitch() gives seg. fault when NULL or zero is provided for any of four parameter.
Verified the fix in my local machine also found that fix has no impact on memory related tests under directed tests.


src/hip_memory.cpp
```diff
@@ -339,6 +339,12 @@ hipError_t hipHostAlloc(void** ptr, size_t sizeBytes, unsigned int flags) {
 // width in bytes
 hipError_t ihipMallocPitch(void** ptr, size_t* pitch, size_t width, size_t height, size_t depth) {
     hipError_t hip_status = hipSuccess;
+	if(ptr==NULL || ptr==0)
```
mangupta: HIP runtime is compiled using a C++ compiler. So comparison to both NULL as well as 0 does not make sense. Just comparing to NULL is sufficient.

mangupta:
Closing in favor of #422

PR #412, opened 2018-04-20
Placeholder PR for HIP program state re-initialization logic

whchung:
Please do NOT merge this PR yet as it shall it be amended, and perhaps rebased to target `master` branch.

This PR tries to re-initialize HIP runtime data structures in `program_state.cpp`. In applications such as TensorFlow it was evident that HIP kernels within shared libraries may not be identified if they are loaded later after initialization by `dlopen()`. Instead of raising an exception immediately we re-initialize data structures within `program_state.cpp` with a `rebuild` flag.

The current implementation may deserve to be amended so that:

- rebase to target `master` branch.
- maps / vectors are stored in TLS so we don't need mutex to guard `function()`.


mangupta:
@whchung Are all the changes in or do we still need to hold off from merging it. Also I guess this needs to be merged to the master branch and not the roc-1.7.x branch.

whchung:
@mangupta Please hold it for now. I need to rebase it to let the PR go into `master` branch, instead of `roc-1.7.x`.

PR #414, opened 2018-04-23
Remove unsupported targets gfx801 and gfx802

aaronenyeshi:
These architectures are no longer supported. Should not compile for them.


aaronenyeshi:
@mangupta @gargrahul we are removing unsupported targets https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/134

PR #415, opened 2018-04-24
Checkin to fix bugs in math functions.

deven-amd:
This change fixes the following bugs that were discovered while debuggnig TF unit test failures (cwise_ops_test)

1. __hisinf and __hisnan routines
   Both had incorrect implementations.

2. abs
   A "long long" (64bit int) version was missing, resulting in the 32bit version being used for 64bit ints (which resulted in incorrect results, when the value passed in was outside the 32bit int range)

3. lgamma
  We seemed to have a custom version for the 'double' datatype (which was giving incorrect results). Replaced it with a call to the 'double' version of the underlying 'hc::precision_math::lgamma'

### DIFF
--- a/include/hip/hcc_detail/math_functions.h
+++ b/include/hip/hcc_detail/math_functions.h
@@ -57,6 +57,7 @@ __device__ float exp2f(float x);
 __device__ float expf(float x);
 __device__ float expm1f(float x);
 __device__ int abs(int x);
+__device__ long long abs(long long x);
 __device__ float fabsf(float x);
 __device__ float fdimf(float x, float y);
 __device__ float fdividef(float x, float y);
--- a/src/hip_fp16.cpp
+++ b/src/hip_fp16.cpp
@@ -29,10 +29,6 @@ struct hipHalfHolder {
     };
 };
 
-#define HINF 65504
-
-__device__ static struct hipHalfHolder __hInfValue = {HINF};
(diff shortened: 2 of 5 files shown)


whchung:
@deven-amd How about add some new tests so we can prevent future issues from happening earlier at `HIP` layer? Perhaps adding new some tests in https://github.com/ROCm-Developer-Tools/HIP/tree/master/tests/src/deviceLib ?

whchung:
@deven-amd test result shows one failure. though unlikely the case, please help check if it's related to changes introduced in this PR:

```
[hcc_1_7] 99% tests passed, 1 tests failed out of 96
[hcc_1_7] 
[hcc_1_7] Total Test time (real) = 181.52 sec
[hcc_1_7] 
[hcc_1_7] The following tests FAILED:
[hcc_1_7] 	 71 - directed_tests/runtimeApi/multiThread/hipMultiThreadStreams2.tst (SEGFAULT)
[hcc_1_7] Errors while running CTest
```

gargrahul:
@whchung No this test failure seems random and failing with other PRs too.
For #419 this fails with hcc-ctu though passes with hcc-1.7.

whchung:
@gargrahul thanks for clarification. in such case could we call this PR fine to be merged? @deven-amd could amend the tests in another PR.

PR #418, opened 2018-04-24
Add host and device targets

pfultz2:
This adds cmake targets for host and device to make it easier to link against the host or device version of the library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -215,7 +215,10 @@ if(HIP_PLATFORM STREQUAL "hcc")
     foreach(TARGET hip_hcc hip_hcc_static hip_device)
         target_include_directories(${TARGET} SYSTEM INTERFACE $<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>;${HSA_PATH}/include)
     endforeach()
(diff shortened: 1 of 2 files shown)


pfultz2:
> However I would like to point out that strictly speaking hip_hcc is not a host-only library.

Yes, but it still needs to be linked when using hip from only a host compiler. `hip_hcc` also includes headers from hc runtime as well which hcc is required. 

The `hip::host` is for when the user is only using the host part of the library and most likely is compiling with a host-only compiler like g++, thus it does not include hcc runtime in its usage requirements.

mangupta:
Linking against hip::host would still require including the hcc runtime since as I mentioned libhip_hcc.so contains a GPU kernel (See https://github.com/ROCm-Developer-Tools/HIP/blob/master/src/hip_memory.cpp#L1522-L1551). Hence both hip::host and hip::device have a dependency on hcc runtime. There was some issue with moving this to the device library. Once that is resolved, what you said will be true. So I approve the merge as long as we documentation or comment somewhere that hip::host currently still has a dependency on hcc runtime.

pfultz2:
Ah, so there is no way to use hip with gcc? We can document it for now until it gets fixed.

pfultz2:
Can this be merged?

PR #425, opened 2018-05-03
Keeping HIP in sync with CUDA

Jorghi12:
Added the Integrated field to the device properties.


Jorghi12:
@bensander - Here https://github.com/ROCm-Developer-Tools/HIP/pull/425/files

bensander:
Don't we have to initialize these fields on both CUDA and ROCM paths in the implementation of hipDeviceGetProperties?

gargrahul:
@Jorghi12 @bensander @mangupta #489 should resolve this.

PR #432, opened 2018-05-05
Add space between `###` and `Notes` in hip_terms

moosichu:
Makes `Notes` an H3 heading.


moosichu:
I'm assuming that was the original intention with the markdown. Please close this if I'm wrong!

PR #445, opened 2018-05-11
Add support for the hipFuncGetAttributes interface.

AlexVlx:
This adds support for hipFuncGetAttributes but, more profoundly, percolates the whole amd_kernel_code_t header structure and makes it accessible from a Kernel_descriptor. This should help with a number of other APIs. Note that this is not fully symmetric with NVIDIA - theirs works on the __device__ side, whereas ours does not yet do so. One of the HIP experts should actually carefully review the mapping from CUDA attribute to HIP attribute, the currently proposed one is minimalistic.

### DIFF
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -483,14 +483,17 @@ namespace
         hipDeviceProp_t prop{};
         hipGetDeviceProperties(
             &prop, ihipGetTlsDefaultCtx()->getDevice()->_deviceId);
-
+        // TODO: at the moment there is no way to query the count of registers
+        //       available per CU, therefore we hardcode it to 64 KiRegisters.
+        prop.regsPerBlock = prop.regsPerBlock ? prop.regsPerBlock : 64 * 1024.
+        
         r.localSizeBytes = header.workitem_private_segment_byte_size;
         r.sharedSizeBytes = header.workgroup_group_segment_byte_size;
         r.maxDynamicSharedSizeBytes =
             prop.sharedMemPerBlock - r.sharedSizeBytes;
         r.numRegs = header.workitem_vgpr_count;
-        r.maxThreadsPerBlock = r.numRegs ? // TODO: proper query.
-            std::min(prop.maxThreadsPerBlock, 64 * 1024 / r.numRegs) :
+        r.maxThreadsPerBlock = r.numRegs ?
+            std::min(prop.maxThreadsPerBlock, prop.regsPerBlock / r.numRegs) :
             prop.maxThreadsPerBlock;
         r.binaryVersion =
             header.amd_machine_version_major * 10 +
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -485,7 +485,7 @@ namespace
             &prop, ihipGetTlsDefaultCtx()->getDevice()->_deviceId);
         // TODO: at the moment there is no way to query the count of registers
         //       available per CU, therefore we hardcode it to 64 KiRegisters.
-        prop.regsPerBlock = prop.regsPerBlock ? prop.regsPerBlock : 64 * 1024.
+        prop.regsPerBlock = prop.regsPerBlock ? prop.regsPerBlock : 64 * 1024;
         
         r.localSizeBytes = header.workitem_private_segment_byte_size;
         r.sharedSizeBytes = header.workgroup_group_segment_byte_size;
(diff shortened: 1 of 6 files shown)


src/hip_module.cpp
```diff
@@ -471,6 +473,61 @@ hipError_t hipModuleGetGlobal(hipDeviceptr_t* dptr, size_t* bytes, hipModule_t h
...
+            &prop, ihipGetTlsDefaultCtx()->getDevice()->_deviceId);
+
+        r.localSizeBytes = header.workitem_private_segment_byte_size;
+        r.sharedSizeBytes = header.workgroup_group_segment_byte_size;
+        r.maxDynamicSharedSizeBytes =
+            prop.sharedMemPerBlock - r.sharedSizeBytes;
+        r.numRegs = header.workitem_vgpr_count;
+        r.maxThreadsPerBlock = r.numRegs ? // TODO: proper query.
+            std::min(prop.maxThreadsPerBlock, 64 * 1024 / r.numRegs) :
+            prop.maxThreadsPerBlock;
```
bensander: this is important query  - what needs to be done to for proper query?

tests/src/runtimeApi/module/hipFuncGetAttributes.cpp
```diff
@@ -0,0 +1,53 @@
...
+    hipFuncAttributes attr{};
+
+    auto r = hipFuncGetAttributes(&attr, reinterpret_cast<const void*>(&fn));
+
+    if (r != hipSuccess || attr.maxThreadsPerBlock == 0) {
+        failed("Failed to read attributes.");
+    }
+
+    passed();
+}
```
bensander: Thank you for including a test!

AlexVlx:
@bensander could you please provide details about what you mean by "output from a few systems" for review? Is there a particular test setup that you have in mind? Note that as things are today for some function foo() the attributes would essentially be the same on gfx8 or gfx9, modulo register allocation differences between the two, since the limits are roughly the same.

src/hip_module.cpp
```diff
@@ -471,6 +473,64 @@ hipError_t hipModuleGetGlobal(hipDeviceptr_t* dptr, size_t* bytes, hipModule_t h
...
+        return x.first == agent;
+    });
+
+    if (it1 == it0->second.cend()) return hipErrorInvalidDeviceFunction;
+
+    const auto header = static_cast<hipFunction_t>(it1->second)->_header;
+
+    if (!header) throw runtime_error{"Ill-formed Kernel_descriptor."};
+
+    *attr = make_function_attributes(*header);
```
scchan: should check attr!=nullptr before using it?

src/hip_module.cpp
```diff
@@ -179,8 +179,10 @@ hipError_t ihipModuleLaunchKernel(hipFunction_t f, uint32_t globalWorkSizeX,
         aql.grid_size_x = globalWorkSizeX;
         aql.grid_size_y = globalWorkSizeY;
         aql.grid_size_z = globalWorkSizeZ;
-        aql.group_segment_size = f->_groupSegmentSize + sharedMemBytes;
-        aql.private_segment_size = f->_privateSegmentSize;
+        aql.group_segment_size =
+            f->_header->workgroup_group_segment_byte_size + sharedMemBytes;
```
scchan: the total group memory size should be checked against the device limit

AlexVlx: Err, it wasn't doing it before. That's an interesting and useful improvement, but not one for this PR, I am certain the HIP team can pursue it.

src/hip_module.cpp
AlexVlx: Why not? Amended:)

src/hip_module.cpp
```diff
@@ -471,6 +473,65 @@ hipError_t hipModuleGetGlobal(hipDeviceptr_t* dptr, size_t* bytes, hipModule_t h
...
+    hipFuncAttributes make_function_attributes(const amd_kernel_code_t& header)
+    {
+        hipFuncAttributes r{};
+
+        hipDeviceProp_t prop{};
+        hipGetDeviceProperties(
+            &prop, ihipGetTlsDefaultCtx()->getDevice()->_deviceId);
+        // TODO: at the moment there is no way to query the count of registers
+        //       available per CU, therefore we hardcode it to 64 KiRegisters.
+        prop.regsPerBlock = prop.regsPerBlock ? prop.regsPerBlock : 64 * 1024.
```
whchung: @AlexVlx build error here

PR #447, opened 2018-05-11
Switch over to using native vector types, for better codegen. Remove noise.

AlexVlx:
This changes the HIP vector types to use the native support provided by Clang/LLVM via the ext_vector_type attribute. This has a number of benefits in terms of the quality of generated code. At the same time, the current implementation of vector types is... wanting. Note that we are still super-setting what CUDA specifies for its vector types, in https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#built-in-vector-types; however, we are sub-setting what C++AMP/HC offer in their respective short vector libraries. Given that HIP aims to match CUDA, this seems like a better place to be in. I've also re-factored the vector type unit tests, which were a testament to the perils of excessive use of copy & paste.


include/hip/hcc_detail/hip_vector_types.h
```diff
@@ -34,1132 +34,90 @@ THE SOFTWARE.
...
-    MAKE_COMPONENT_CONSTRUCTOR_FOUR_COMPONENT(longlong4, signed long long)
-#endif
-    signed long x, y, z, w;
-} __attribute__((aligned(32)));
+#if defined(__clang__)
+    #define __NATIVE_VECTOR__(n, ...) __attribute__((ext_vector_type(n)))
+#elif defined(__GNUC__) // N.B.: GCC does not support .xyzw syntax.
+    #define __NATIVE_VECTOR__(n, T) __attribute__((vector_size(n * sizeof(T))))
+#endif
+
```
scchan: Essentially codes compiling with clang and gcc will be using different data types?   Will that affect extern functions that have vector types in its signature such that code compiling by one compiler can't call the function compiled by the other compiler due to name mangling difference?

AlexVlx: I believe that the only safe way to call across compilers is in extern "C", otherwise you are on your own anyway. Thus there'd be no mangling difference given that extern "C" functions are not mangled. Even if it were a mere struct there are no guarantees of mangling being matched between compilers (on the contrary). So whereas before that'd have been a quiet way to get yourself into trouble (oh, ok, I'm calling my super interesting template across ABI boundaries and it works yay... oh, snap, what just broke after this GCC upgrade?), this may fail early (annoyingly enough though, mangling is mostly matched at the moment).

PR #448, opened 2018-05-14
Provide correct __mul64hi and __umul64hi builtins, using code from ROCm-Device-Libs

### LINKED ISSUE #356: Some builtins in device_functions.cpp seem wrong
Looking at the following [code](https://github.com/ROCm-Developer-Tools/HIP/blob/f421f2fe0ae136d13977d1bee669072607b7a7f9/src/device_functions.cpp#L498):
```cpp
struct intHolder {
  union {
    signed int si[2];
    signed int long sl;
  };
}__attribute__((aligned(8)));

struct uintHolder {
  union {
    signed int ui[2];
    signed int long ul;
  };
}__attribute__((aligned(8)));

__device__ long long __mul64hi(long long int x, long long int y)
{
  struct intHolder iHold1;
  struct intHolder iHold2;
  iHold1.sl = x;
  iHold2.sl = y;
  iHold1.sl = iHold1.si[1] * iHold2.si[1];
  return iHold1.sl;
}

__device__ unsigned long long __umul64hi(unsigned long long int x, unsigned long long int y)
{
  struct uintHolder uHold1;
  struct uintHolder uHold2;
  uHold1.ul = x;
  uHold2.ul = y;
  uHold1.ul = uHold1.ui[1] * uHold2.ui[1];
  return uHold1.ul;
}
```
Even if we changed the fields in `uintHolder` to `unsigned`, to fix the the most obviously wrong thing here, I believe the code still will not work. If you look at it, it really does nothing but multiply the respective 32 high-order bits of its arguments, treated as 32 bit values -- but the maximum unsigned 64-bit value is a trivial example of where this leads to a wrong result:
```
(0xFFFFFFFFFFFFFFFF * 0xFFFFFFFFFFFFFFFF) >> 64         == 0xFFFFFFFFFFFFFFFE
(0xFFFFFFFFFFFFFFFF >> 32) * (0xFFFFFFFFFFFFFFFF >> 32) == 0xFFFFFFFE00000001
```

[NVIDIA spec:](http://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__INTRINSIC__INT.html#group__CUDA__MATH__INTRINSIC__INT_1gefaf2a1bb2ad8986b2c248b74461e12e)

> Calculate the most significant 64 bits of the 128-bit product x * y, where x and y are 64-bit unsigned integers. 

I personally don’t know how to implement this without 128 bit types, else I’d file a pull request myself.

For now I’m using this here (if you assume a little-endian system, instead of shifting you can also address the high-order bits directly), which hipcc/hcc accepts and which hopefully works correctly on the GPU as well.
```cpp
#define MAXLONG 0xFFFFFFFFFFFFFFFF
__device__ __forceinline__ uint64_t umul64hi( __uint128_t m1, __uint128_t m2 ) {
        __uint128_t res = m1 * m2;
       return (uint64_t) ( (res >> 64) & MAXLONG );
}
```

949f45ac:
Fixes #356.

The new code is taken verbatim from [ROCm-Device-Libs](https://github.com/RadeonOpenCompute/ROCm-Device-Libs/blob/master/opencl/src/integer/int.h#L70), i.e. the ROCm OpenCL implementation of these very same functions. Hence I hope there are no licensing issues or anything, as both projects are part of the ROCm effort.

Let me know if I should add tests somewhere; I haven’t looked into that because the previous implementation was apparently untested as well.

Cheers!

### DIFF
--- a/src/device_functions.cpp
+++ b/src/device_functions.cpp
@@ -276,6 +276,13 @@ struct ucharHolder {
     };
 } __attribute__((aligned(4)));
 
+struct uchar2Holder {
+    union {
+        unsigned int ui[2];
+        unsigned char c[8];
+    };
+} __attribute__((aligned(8)));
+
 __device__ unsigned int __byte_perm(unsigned int x, unsigned int y, unsigned int s) {
     struct uchar2Holder cHoldVal;
     struct ucharHolder cHoldKey;
--- a/src/device_functions.cpp
+++ b/src/device_functions.cpp
@@ -276,28 +276,6 @@ struct ucharHolder {
     };
 } __attribute__((aligned(4)));
 
-struct uchar2Holder {
-    union {
-        unsigned int ui[2];
-        unsigned char c[8];
-    };
-} __attribute__((aligned(8)));
-
-struct intHolder {
-    union {
-        signed int si[2];
-        signed int long sl;
-    };
-} __attribute__((aligned(8)));
-
-struct uintHolder {
-    union {
-        signed int ui[2];
-        signed int long ul;
-    };
-} __attribute__((aligned(8)));
-
-
 __device__ unsigned int __byte_perm(unsigned int x, unsigned int y, unsigned int s) {
     struct uchar2Holder cHoldVal;
     struct ucharHolder cHoldKey;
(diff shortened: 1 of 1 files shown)


src/device_functions.cpp
```diff
@@ -276,28 +276,6 @@ struct ucharHolder {
     };
 } __attribute__((aligned(4)));
 
-struct uchar2Holder {
```
mangupta: This struct is used by ```__byte_perm``` device function. So we cannot remove it.

PR #457, opened 2018-05-18
HIP program state re-initialization logic

whchung:
This PR tries to re-initialize HIP runtime data structures in program_state.cpp. In applications such as TensorFlow it was evident that HIP kernels within shared libraries may not be identified if they are loaded later after initialization by dlopen(). Instead of raising an exception immediately we re-initialize data structures within program_state.cpp with a rebuild flag.

### DIFF
--- a/src/program_state.cpp
+++ b/src/program_state.cpp
@@ -74,11 +74,15 @@ vector<string> copy_names_of_undefined_symbols(const symbol_section_accessor& se
 }
 
 const std::unordered_map<std::string, std::pair<ELFIO::Elf64_Addr, ELFIO::Elf_Xword>>&
-symbol_addresses() {
+symbol_addresses(bool rebuild = false) {
     static unordered_map<string, pair<Elf64_Addr, Elf_Xword>> r;
     static once_flag f;
 
-    call_once(f, []() {
+    auto cons = [rebuild]() {
+        if (rebuild) {
+            r.clear();
+        }
+
         dl_iterate_phdr(
             [](dl_phdr_info* info, size_t, void*) {
                 static constexpr const char self[] = "/proc/self/exe";
@@ -108,7 +112,12 @@ symbol_addresses() {
                 return 0;
             },
             nullptr);
-    });
+    };
+
+    call_once(f, cons);
+    if (rebuild) {
+        cons();
+    }
 
     return r;
 }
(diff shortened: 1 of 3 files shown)


whchung:
/cc @adityaaatluri @sabreshao for awareness

The patch to HIP for program state re-initialization has been updated so it could possibly be merged with the tip of HIP. Shall this PR be merged projects such as RCCL & PaddlePaddle should run properly.

whchung:
Build issue in CI #5 should have nothing to do with the PR as the PR has had successfully CI build #4 before.

Meanwhile please hold this PR as we need to make sure end applications do function properly with this PR.

mangupta:
@whchung Is this PR fine to merge now? Or do we still need to verify some end applications?

whchung:
TensorFlow is fine but PaddlePaddle is not. Dynamic shared library loading behavior of PaddlePaddle is more complicated and we’ll hit ROCR runtime limitations. The PR needs to be improved to only construct HSA executables for newly identified code objects.

whchung:
Working on reducing data structures need to be reinitialized when we hit a missing kernel function address.

src/program_state.cpp
```diff
@@ -295,13 +342,19 @@ void load_code_object_and_freeze_executable(
...
         static const auto accelerators = hc::accelerator::get_all();
 
+        if (rebuild) {
+            // do NOT clear r so we reuse instances of hsa_executable_t
+            // created previously
+            code_object_blobs(rebuild);
+        }
+
         for (auto&& acc : accelerators) {
             auto agent = static_cast<hsa_agent_t*>(acc.get_hsa_agent());
```
whchung: further improvement should be made here to reduce calls to `load_executables`

whchung:
@mangupta
TensorFlow nightly builds have been adopting the updated PR for 3 days and everything looks fine. Now we just need confirmation from PaddlePaddle team so we can merge this in.

whchung:
@mangupta Earlier today PaddlePaddle team reported an error and I've fixed it in commit# 32789a8 . Now both PaddlePaddle and TensorFlow could properly load dynamic libs without bumping into ROCR runtime limits. Please help review this PR once again and consider merge it. Thanks.

Also it would be preferable if this PR could be made into `roc-1.8.x` release branch for upcoming 1.8.2 release.

jichangjichang:
With below 3 commits, PaddlePaddle can dynamic load rccl library and run rccl kernels normally.
- HIP program state re-initialization logic
- Improve performance of re-initialization logic
- Keep the map which tracks GPU kernel symbols to grow monotonically

whchung:
@mangupta thanks!

@parallelo now HIP PR 457 has been merged I’ll submit a PR to change Dockerfile for tensorflow build

PR #465, opened 2018-05-23
Optimize memcpy2D kernel use

gargrahul:
Optimize memcpy2D kernel use

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -1496,19 +1496,15 @@ __global__ void hip_copy2d_n(T* dst, const T* src, size_t width, size_t height, 
     size_t idx = blockIdx.x * blockDim.x + threadIdx.x;
     size_t idy = blockIdx.y * blockDim.y + threadIdx.y;
     size_t floorWidth = (width/sizeof(T));
-    if((idx < width) && (idy < height)) {
-        if((idx < floorWidth)){
-            T *dstPtr = (T *)((uint8_t*) dst + idy * destPitch);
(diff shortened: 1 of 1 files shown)


src/hip_memory.cpp
```diff
@@ -1495,10 +1495,18 @@ __global__ void hip_copy2d_n(T* dst, const T* src, size_t width, size_t height,
...
     size_t idx = blockIdx.x * blockDim.x + threadIdx.x;
     size_t idy = blockIdx.y * blockDim.y + threadIdx.y;
-    if((idx < width) && (idy < height)){
+    size_t floorWidth = (width/sizeof(T));
+    if((idx < floorWidth)){
         T *dstPtr = (T *)((uint8_t*) dst + idy * destPitch);
         T *srcPtr = (T *)((uint8_t*) src + idy * srcPitch);
         dstPtr[idx] = srcPtr[idx];
+    } else {
+        size_t bytesToCopy = width - floorWidth;
```
mangupta: Shouldn't this be ```width - floorWidth*sizeof(T)```?

fronteer:
Have you tried to get a corresponding device pointer for the host pointer before calling the kernel.  When the host memory is allocated through malloc()+hipHostRegister(),  the device pointer is different from the host pointer,  you codes could not call the kernel directly by passing the host pointer as parameter.

gargrahul:
@mangupta please review the changes again.

gargrahul:
@fronteer Thanks. Malloc + hipHostRegister is a valid case and will be fixed in a separate PR.

PR #466, opened 2018-05-23
Feature use _Float16 and match CUDA __half behaviour.

AlexVlx:


Jorghi12:
This is great! Thank you SO much @AlexVlx for stepping up on this.

Jorghi12:
Quick question, why is the CI not building the commit? I'm not internal to AMD so I can't see why the merge is blocked.

AlexVlx:
@Jorghi12 a weird mix-up pertaining to how unit tests are built as part of the test procedure.

Jorghi12:
Here's a summary of desirable changes.

1) To avoid operator overloading issues, operators defined for more types would be beneficial.
**__device__ float abs(float x);**
**__device__ double abs(double x);**

2) Missing the **labs** mathematics function. CUDA provides this function for computing the absolute value of int64_t (long int). This is a trivial addition, but yet another helpful one to reduce the number of code modifications when compiling under NVCC or HCC.

`__device__ long int labs(long int x);`

3) The new version of hip_fp16.h appears to allow for implicit conversions to float. This results in operator overloading issues in PyTorch. This type of behavior **will** cause issues for frameworks that rely on the [__CUDA_NO_HALF_CONVERSIONS__](https://github.com/ptillet/isaac/blob/master/include/isaac/external/CUDA/cuda_fp16.hpp#L142) flag. It's a desirable flag since implicit conversions may lead to unintended behavior. 

However, given that the __half / __half_raw types are heavily bound to float, removing the conversion operators breaks the implementation.

For now in PyTorch, we'll simply work around this.

AlexVlx:
@Jorghi12 the implicit conversion follows what latest CUDA does. The equivalent HIP_NO_HALF_CONVERSIONS define is available if you do not want the conversion (this, again, follows the current CUDA convention AFAICS). Could you please verify that it works as intended?

For (2) and (3) that is a separate issue that we should handle elsewhere. It is probable that there are gaps in the implementation of the library, so perhaps opening a separate bug that tracks all of them (that you observed) could be opportune.

Jorghi12:
@AlexVlx These are just nitpicking comments. The half implementation works great otherwise! As for the math library issues, I can send a PR for that to HIP.

**hip_fp16:**
Even with HIP_NO_HALF_CONVERSIONS enabled, there's still an implicit conversion from __half to float.

Here's an example invocation)
**hcc test.cpp -I/opt/rocm/hip/include -DHIP_NO_HALF**

`void fmaxType(half x, half y) {}

void fmaxType(float x, half y) {}

void fmaxType(float x, float y) {}

void fmaxType(double x, double y) {}

int main(){
   __half x;
   __half y;
   fmaxType(x, y);
}` 

test.cpp:14:4: error: call to 'fmaxType' is ambiguous
   fmaxType(x, y);
   ^~~~~~~~
test.cpp:3:6: note: candidate function
void fmaxType(half x, half y) {}
     ^
test.cpp:5:6: note: candidate function
void fmaxType(float x, half y) {}
     ^
test.cpp:7:6: note: candidate function
void fmaxType(float x, float y) {}
     ^
test.cpp:9:6: note: candidate function
void fmaxType(double x, double y) {}

Jorghi12:
@AlexVlx These are just nitpicking comments. The half implementation works great otherwise. As for the math library issues, I can send a PR for that to HIP.

**hip_fp16:**
Even with HIP_NO_HALF_CONVERSIONS enabled, there's still an implicit conversion from __half to float.

**hcc test.cpp -I/opt/rocm/hip/include -DHIP_NO_HALF_CONVERSIONS**

```
void fmaxType(half x, half y) {}

void fmaxType(float x, half y) {}

void fmaxType(float x, float y) {}

void fmaxType(double x, double y) {}

int main(){
   __half x;
   __half y;
   fmaxType(x, y);
}
```

test.cpp:14:4: error: call to 'fmaxType' is ambiguous
   fmaxType(x, y);
   ^~~~~~~~
test.cpp:3:6: note: candidate function
void fmaxType(half x, half y) {}
     ^
test.cpp:5:6: note: candidate function
void fmaxType(float x, half y) {}
     ^
test.cpp:7:6: note: candidate function
void fmaxType(float x, float y) {}
     ^
test.cpp:9:6: note: candidate function
void fmaxType(double x, double y) {}

AlexVlx:
@Jorghi12, I believe there is a bit of confusion here. Let's step back a bit. The full example you provided is not quite supposed to work:
- __half_raw is just a POD, there's no __hdiv taking __half_raw arguments;
- __half has a single argument (conversion) constructor from __half_raw, which gets invoked to construct temporary arguments to the __hdiv calls, fine so far;
- fmaxType has no overload taking __half (a direct match), however it has multiple overloads with equal scores, since they are one user defined conversion away (there are conversions to float, __half_raw etc., which follow CUDA's lead here) - this yields an ambiguity;
- __hdiv is a __device__ only function therefore this would merely fail semantic analysis later.
Could you please share the original example where this fails? I'd like us to get to the bottom of it and fix things so that you don't need to work around anything in the client application. Thank you.

Jorghi12:
@AlexVlx  This works with CUDA actually.

Jorghi12:
@AlexVlx So every point you provided is correct except #3. 

With CUDA, when we have __CUDA_NO_HALF_CONVERSIONS__, then the highest scoring overload will be fmaxType(__half_raw, __half_raw) actually.

Jorghi12:
For example, when running the exact same code using CUDA's fp16.h structures, the overloading issue doesn't appear. Does this help?

```
typedef struct {
    unsigned short x;
} __half_raw;

struct  __half {
protected:
    unsigned short __x;

public:
#if __cplusplus >= 201103L
    __half() = default;
#else
    __half() { }
#endif /* __cplusplus >= 201103L */

    /* Convert to/from __half_raw */
     __half(const __half_raw &hr) : __x(hr.x) { }
     __half &operator=(const __half_raw &hr) { __x = hr.x; return *this; }
     operator __half_raw() const { __half_raw ret; ret.x = __x; return ret; }
};

void fmaxType(__half_raw x, __half_raw y) {}

void fmaxType(float x, __half_raw y) {}

void fmaxType(__half_raw x, float y) {}

void fmaxType(float x, float y) {}

void fmaxType(int x, int y) {}

void fmaxType(double x, double y) {}

int main(){
   __half x;
   __half y;
   fmaxType(x, y);
}
```

AlexVlx:
@Jorghi12 Righto, it was my fault: unlike CUDA, we support FP16 natively, and use the _Float16 type liberally. The converting constructor for __half taking a _Float16 was not guarded under the macro, and this yielded the ambiguity. Correcting that also required modifying the implementation of various functions. I've updated the PR to reflect the new state of affairs. I suspect that there may be another potential challenge in that we also provide an implicit conversion to _Float16, which may come into play when we expose FP16 native math (e.g. sqrt(_Float16) as opposed to hsqrt(__half)), but we'll cross that bridge when we get there. If you could assess the new implementation it'd be highly appreciated.

AlexVlx:
@Jorghi12 Righto, that was on me, I had misplaced the macro-guards. I expect the new iteration to be closer in behaviour to what CUDA does.

Jorghi12:
@AlexVlx - Yes, the new iteration is much closer in behavior to what CUDA does now. Awesome 2000+ line differential, @AlexVlx. This looks great, thanks!!

Jorghi12:
The header for hip_fp16_gcc.h is missing a type definition for half.

```
#if !defined(HIP_NO_HALF)
    using half = __half;
    using half2 = __half2;
#endif
```

AlexVlx:
@Jorghi12 added, thank you.

deven-amd:
@AlexVlx  do you plan to add support for `half` and `half2` be added to the `__shfl_*` routines (similar to what is present in the `cuda_fp16.hpp` file). If yes, can it be included in this PR. thanks

AlexVlx:
@deven-amd there is no such plan, but I'm pretty sure that subsequent work can be done (by the HIP team or other contributors) to adopt those as well. So no, that won't be part of this particular PR, sorry.

deven-amd:
@mangupta @AlexVlx 

I need to update the TF code along the lines of

```
#if HIPCC_VERSION >  "version_number that has the new __half from this commit"
...
#else
...
#endif
```

What is the version number I should check for? Where can I look it up in the codebase?

Thanks

deven

whchung:
https://github.com/ROCm-Developer-Tools/HIP/blob/master/bin/hipcc#L65

I guess we need to use `HIP_VERSION_PATCH` with number > `18225`?

deven-amd:
@mangupta  can you confirm the number....it should be `18225` (given that the merge happened last friday), but when I build HIP using the latest code in `master`, the HIP_VERSION_PATCH number I get is `18215` (it does contain this merge)

PR #477, opened 2018-05-30
Add signature for hipFuncGetAttributes to nvcc_details

mangupta:


pradeepisro:
@mangupta it is `cuFuncGetAttribute` not `cuFuncGetAttributes`

mangupta:
@pradeepisro hipFuncGetAttributes is the equivalent for cudaFuncGetAttributes (which is a runtime API). cuFuncGetAttribute is a driver API which does not have an equivalent in HIP as yet.

PR #482, opened 2018-05-31
Switch to using ROCDL directly, as opposed to via HC. Add missing bits.

AlexVlx:
This makes HIP use ROCDL directly, which is highly preferable. Function definitions are provided inline in the relevant header. Certain functions which were previously unimplemented are added.


include/hip/hcc_detail/math_functions.h
```diff
@@ -20,237 +20,1125 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
+}
+__device__
+inline
+float normcdff(float x) { return __ocml_ncdf_f32(x); }
+__device__
+inline
+float normcdfinvf(float x) { return __ocml_ncdfinv_f32(x); }
+__device__
+inline
+float normf(int dim, const float* a)
```
b-sumner: Accuracy is poor and no overflow protection.  Better add comment that this is just a placeholder until there is demand for this function.

include/hip/hcc_detail/math_functions.h
```diff
@@ -20,237 +20,1125 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
+__device__
+inline
+float j0f(float x) { return __ocml_j0_f32(x); }
+__device__
+inline
+float j1f(float x) { return __ocml_j1_f32(x); }
+__device__
+inline
+float jnf(int n, float x)
+{   // TODO: we could use Ahmes multiplication and the Miller & Brown algorithm
```
b-sumner: We could do a lot of things with varying accuracy and performance.  Better just comment for now that this is just a placeholder until we see some demand and have some accuracy requirements.

include/hip/hcc_detail/math_functions.h
```diff
@@ -20,237 +20,1125 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
+
+__device__
+inline
+float rnorm4df(float x, float y, float z, float w)
+{
+    return __ocml_rlen4_f32(x, y, z, w);
+}
+__device__
+inline
+float rnormf(int dim, const float* a)
```
b-sumner: Poor accuracy and no overflow protection.  Better add comment that this is a placeholder until we see demand for this function.

include/hip/hcc_detail/math_functions.h
```diff
@@ -20,237 +20,1125 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
+__device__
+inline
+float y0f(float x) { return __ocml_y0_f32(x); }
+__device__
+inline
+float y1f(float x) { return __ocml_y1_f32(x); }
+__device__
+inline
+float ynf(int n, float x)
+{   // TODO: we could use Ahmes multiplication and the Miller & Brown algorithm
```
b-sumner: Same comment as for jnf

include/hip/hcc_detail/math_functions.h
```diff
@@ -20,237 +20,1125 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
+float __cosf(float x) { return __llvm_amdgcn_cos_f32(x); }
+__device__
+inline
+float __exp10f(float x) { return __ocml_exp10_f32(x); }
+__device__
+inline
+float __expf(float x) { return __ocml_exp_f32(x); }
+__device__
+inline
+float __fadd_rd(float x, float y) { return __llvm_add_rte_f32(x, y); }
```
b-sumner: These special rounding operations are in OCML and should be used from there.  BTW, they do not work currently.

include/hip/hcc_detail/math_functions.h
```diff
@@ -20,237 +20,1125 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
+int isnan(double x) { return __ocml_isnan_f64(x); }
+__device__
+inline
+double j0(double x) { return __ocml_j0_f64(x); }
+__device__
+inline
+double j1(double x) { return __ocml_j1_f64(x); }
+__device__
+inline
+double jn(int n, double x)
```
b-sumner: Same comment as for single precision

include/hip/hcc_detail/math_functions.h
```diff
@@ -20,237 +20,1125 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
+}
+__device__
+inline
+double nearbyint(double x) { return __ocml_nearbyint_f64(x); }
+__device__
+inline
+double nextafter(double x, double y) { return __ocml_nextafter_f64(x, y); }
+__device__
+inline
+double norm(int dim, const double* a)
```
b-sumner: Same comment as for single precision

include/hip/hcc_detail/math_functions.h
```diff
@@ -20,237 +20,1125 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
+}
+__device__
+inline
+double rhypot(double x, double y) { return __ocml_rhypot_f64(x, y); }
+__device__
+inline
+double rint(double x) { return __ocml_rint_f64(x); }
+__device__
+inline
+double rnorm(int dim, const double* a)
```
b-sumner: Same comment as for single precision

include/hip/hcc_detail/math_functions.h
```diff
@@ -20,237 +20,1125 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
+double trunc(double x) { return __ocml_trunc_f64(x); }
+__device__
+inline
+double y0(double x) { return __ocml_y0_f64(x); }
+__device__
+inline
+double y1(double x) { return __ocml_y1_f64(x); }
+__device__
+inline
+double yn(int n, double x)
```
b-sumner: Same comment as for single precision

include/hip/hcc_detail/math_functions.h
```diff
@@ -20,237 +20,1125 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
+        x1 = x2;
+    }
+
+    return x1;
+}
+
+// BEGIN INTRINSICS
+__device__
+inline
+double __dadd_rd(double x, double y) { return __llvm_add_rtp_f64(x, y); }
```
b-sumner: Same comment as for single precision

include/hip/hcc_detail/ocml_math_fwd.h
```diff
@@ -0,0 +1,431 @@
...
+float __ocml_tan_f32(float);
+__attribute__((pure))
+float __ocml_tanh_f32(float);
+float __ocml_tgamma_f32(float);
+__attribute__((const))
+float __ocml_trunc_f32(float);
+float __ocml_y0_f32(float);
+float __ocml_y1_f32(float);
+
+// BEGIN INTRINSICS
```
b-sumner: These should all be removed.  Specially rounded operations are in OCML (but not working)

AlexVlx: Done, thanks.

include/hip/hcc_detail/math_functions.h
AlexVlx: @b-sumner This pretty much applies to all other items - yes. However, these wonderful things have no place in the HIP RT, but rather should be in OCML. Ideally, at some point in the future, __ocml_jn_f32(...) will be available, and the HIP RT will just add a forward declare and call that.

include/hip/hcc_detail/ocml_math_fwd.h
```diff
@@ -0,0 +1,431 @@
...
+*/
+
+#pragma once
+
+#if defined(__cplusplus)
+    extern "C" {
+#endif
+
+// BEGIN FLOAT
+__attribute__((const))
```
yxsamliu: CUDA/HIP does not infer __device__ attribute. device function can only call device function, therefore, these functions need __device__ attribute to work with hip-clang.

AlexVlx: I was unaware that this is now a shared repository, and that we're doing shared development for the two nowadays...

yxsamliu: Since the device code for HCC and hip-clang are similar, so hip-clang now uses large portion of code in hcc_details to save time. This works pretty fine so far.

include/hip/hcc_detail/math_functions.h
b-sumner: Agreed.

include/hip/hcc_detail/ocml_math_fwd.h
AlexVlx: Yes but you have no dedicated unit tests, and create extra overhead for people that may not want to be involved with that project. The present situation outlines both of those aspects - since there are no hip-clang unit tests in this repository any breakage trickles down via the grapevine, inefficiently.

include/hip/hcc_detail/ocml_math_fwd.h
```diff
@@ -0,0 +1,431 @@
...
+float __ocml_fma_rte_f32(float, float, float);
+__attribute__((const))
+float __ocml_fma_rtn_f32(float, float, float);
+__attribute__((const))
+float __ocml_fma_rtp_f32(float, float, float);
+__attribute__((const))
+float __ocml_fma_rtz_f32(float, float, float);
+
+__attribute__((const))
```
b-sumner: Why are these 4 here?  I don't think they belong.

include/hip/hcc_detail/ocml_math_fwd.h
```diff
@@ -0,0 +1,431 @@
...
+double __ocml_fma_rte_f64(double, double, double);
+__attribute__((const))
+double __ocml_fma_rtn_f64(double, double, double);
+__attribute__((const))
+double __ocml_fma_rtp_f64(double, double, double);
+__attribute__((const))
+double __ocml_fma_rtz_f64(double, double, double);
+
+__attribute__((const))
```
b-sumner: Why are these two here?  I don;t think they belong.

include/hip/hcc_detail/ocml_math_fwd.h
```diff
@@ -0,0 +1,431 @@
...
+double __llvm_amdgcn_rcp_f64(double) __asm("llvm.amdgcn.rcp.f64");
+__attribute__((const))
+double __llvm_amdgcn_rsq_f64(double) __asm("llvm.amdgcn.rsq.f64");
+// END INTRINSICS
+// END DOUBLE
+
```
b-sumner: FWIW, OCML provides half precision variants of the math functions.  Should they be included here for later use?

include/hip/hcc_detail/ocml_math_fwd.h
AlexVlx: They are there because e.g.: https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__INTRINSIC__DOUBLE.html. Do you have an actual technical concern here or is it just a thing?

include/hip/hcc_detail/ocml_math_fwd.h
AlexVlx: See above.

include/hip/hcc_detail/ocml_math_fwd.h
AlexVlx: They are already used in the __half / __half2 implementations. If this gets merge moving them into this header may be worth considering. Not a concern for this PR.

include/hip/hcc_detail/ocml_math_fwd.h
yxsamliu: hip-clang is not ready for unit test yet, but will be.

If it causes any regressions in hip-hcc it will be caught by existing unit test. And we will fix hip-clang issues.

include/hip/hcc_detail/ocml_math_fwd.h
b-sumner: Sorry, I meant the 4 following __llvm_amdgcn_<fn>_f32 (and similarly for the two f64 below.  These should never be called directly from HIP.

include/hip/hcc_detail/ocml_math_fwd.h
yxsamliu: Actually hip-clang uses the same directed unit-tests as hip-hcc. Just need to set some environment variables. It is OK if you would not like to make changes that are needed by hip-clang. Aaron or I can add it in a separate commit.

include/hip/hcc_detail/ocml_math_fwd.h
AlexVlx: I don't believe that is true. Note that CUDA actually exposes intrinsics, amongst which there is e.g. https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__INTRINSIC__DOUBLE.html#group__CUDA__MATH__INTRINSIC__DOUBLE_1g76b42ce1d523e8655454a061d0ed3728, which are understood to be hardware specific. There's no reason for us not to use the equivalent capability; furthermore OCML doesn't expose things like reciprocal, and penalising the user by doing this as 1 / x appears counterproductive.

b-sumner: HIP does not define and document a "native" API, so I'm not sure this is a good idea.  At least change the function names to something meaningful like "amd_native_xyz" instead of those deliberately ugly  __llvm_amdgcn_xyz_fMN.

AlexVlx: I'm not sure I follow. The user calls the HIP intrinsics, which MUST match the CUDA name. This is not a matter of whether or not it is a good idea, it is practically infeasible to not support them as ported code will break. As for the intrinsic, that's a forward declare that matches the actual name, just like all other forward declares in that header, and is an implementation detail - the user calls `__frcp_rd(x)` and not `__llvm_amdgcn_rcp_f32(x)`. Making something that is intended to signal "danger danger" cute would subvert the initial intent, ignoring the fact that it simply does not work - at some point something must call `__llvm_amdgcn_rcp_f32`, and a declaration of that must be in scope.

b-sumner: I think this finally starts to answer my original question "Why are these 4 here".  And the answer appears to be "they are used in some other functions such as __frcp which we have to implement and are supposed to provide native ISA semantics".  Now I understand why these exist at all, but these still don't fit in a file named ocml_math_fwd.h.

AlexVlx: Brilliant! Yes,  the name of the header is suboptimal in that it is not generic. I've adjusted this, thanks.

gargrahul:
@AlexVlx @b-sumner This PR is causing performance drop in rocrand (SWDEV-157382 or SWDEV-157378). Can we do something for performance improvement?

AlexVlx:
@gargrahul this is due to the more conservative implementation of the float intrinsics; at least the normal_distribution test is likely to be affected by `__sincosf` no longer relying on AMDGCN native `sin` + `cos`. That being said, unless there are even cursory targeted performance tests in place to measure this and present it outside of "this dense codebase seems to suffer between two HIP commits; I had to go and grok rocRAND to deduce what may be happening), which can drive optimisation here, it would be extremely unwise to do anything. So this is a chance for CQE to shine and present a nice thorough analysis.

PR #483, opened 2018-05-31
Fix hipMemcpy2D

gargrahul:
Fix hipMemcpy2D


gargrahul:
Yes, I think this can be ignored.

PR #490, opened 2018-06-02
Switch the atomic implementation to use Clang  builtins.

AlexVlx:
Atomics need not be grabbed from HC, but can rather rely on builtins, which we lower correctly nowadays.


include/hip/hcc_detail/hip_atomic.h
```diff
@@ -0,0 +1,265 @@
...
+    while (tmp < val) { tmp = atomicCAS(address, tmp, val); }
+
+    return tmp;
+}
+
+__device__
+inline
+unsigned int atomicInc(unsigned int* address, unsigned int val)
+{
+    extern unsigned int __builtin_amdgcn_atomic_inc(
```
yxsamliu: hip-clang needs __device__ here

include/hip/hcc_detail/hip_atomic.h
```diff
@@ -0,0 +1,265 @@
...
+
+    return __builtin_amdgcn_atomic_inc(
+        address, val, __ATOMIC_RELAXED, 1 /* Device scope */, false);
+}
+
+__device__
+inline
+unsigned int atomicDec(unsigned int* address, unsigned int val)
+{
+    extern unsigned int __builtin_amdgcn_atomic_dec(
```
yxsamliu: need __device__

yxsamliu:
LGTM other than the two comments. I tested that this makes HIP-Example reduction pass with hip-clang.

Great work! Thanks!

AlexVlx:
@yxsamliu whoops, apologies, forgot about them pesky `__device__`s; added, thank you!

include/hip/hcc_detail/hip_atomic.h
```diff
@@ -0,0 +1,269 @@
...
+unsigned long long atomicAdd(
+    unsigned long long* address, unsigned long long val)
+{
+    return __atomic_fetch_add(address, val, __ATOMIC_RELAXED);
+}
+__device__
+inline
+float atomicAdd(float* address, float val)
+{
+    unsigned int* uaddr{reinterpret_cast<unsigned int*>(uaddr)};
```
yxsamliu: I got warning when compiling hipCaffe:
 warning: variable 'uaddr' is uninitialized when used within its own initialization [-Wuninitialized]
    unsigned int* uaddr{reinterpret_cast<unsigned int*>(uaddr)};
                  ~~~~~                                 ^~~~~
It seems this line tries to initialize uaddr with its own value. Is this intended?

include/hip/hcc_detail/hip_atomic.h
```diff
@@ -0,0 +1,269 @@
...
+        old = atomicCAS(uaddr, r, __float_as_uint(val + __uint_as_float(r)));
+    } while (r != old);
+
+    return __uint_as_float(r);
+}
+__device__
+inline
+double atomicAdd(double* address, double val)
+{
+    unsigned long long* uaddr{reinterpret_cast<unsigned long long*>(uaddr)};
```
yxsamliu: same warning as above

AlexVlx:
@yxsamliu not intended at all, those are pretty terrible typos on my part, thank you for spotting them; we shall have to revisit the unit test in the future since those shouldn't have gone through.

PR #491, opened 2018-06-02
callback handling: don't need to wait for the thread to become ready

scchan:


scchan:
@mangupta This fix will have to go into the roc-1.8.x too

scchan:
@mangupta one concern I have with the HIP runtime spawning a new thread to invoke the callback function is that if the callback uses any TLS variable, then it will be broken.

whchung:
@scchan I think it's a concern we shouldn't be worrying about. The callbacks are supplied by user applications, and it'd be of little reason for the user-supplied callbacks to use TLS storage, as all relevant information are supplied by HIP runtime.

whchung:
@mangupta On the other hand I could see improvement for the documentation of `hipStreamAddCallback`. It seem has `cuda` reference in the document. And it's not very clear in which thread the callback be invoked.

whchung:
@mangupta This fix will have to go into the roc-1.8.x too

PR #493, opened 2018-06-05
Callback/Activity tracer frontend

eshcherb:


whchung:
@eshcherb I'd recommend you first create a PR against `master` so it's in the bleeding-edge of HIP. Whether to adopt it into `roc-1.8.x` can be discussed later after review / approval from @mangupta . Thus I'd recommend you do either:

1)  close this PR, do `git rebase` in your branch so your changes happen after the tip of `master`.
2) fix merge conflicts in this PR

src/hip_intercept.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include "hsa/hsa_ext_amd.h"
+#include "hip/hip_runtime.h"
+#include "hip_hcc_internal.h"
+#include "trace_helper.h"
+#include "hip/hip_cbapi.h"
+
+///////////////////////////////////////////////////////////////////////////////////
+// HIP API callbacks/activity
+
+extern "C" size_t HSAOp_get_record_size();
```
whchung: Having to `extern "C"` is not a good idea. Before long HIP runtime may need to adopt another interface called VDI. These interfaces breaks such abstraction and mandates HIP runtime be coupled with existing HCC runtime.

whchung:
@eshcherb Could you explain the dependency of this PR with https://github.com/ROCmSoftwarePlatform/roctracer . It's not clear to me why there is no need to explicit include headers of `roctracer` in this PR.

eshcherb:
My changes were mad against roc-1.8.x branch
So to check then against 'master' I need to reaply them to 'master' branch
Will do it after I back from vacation the next week

eshcherb:
I think it is better to define activity records types in ROC-tracer and HIP/HCC will include external headers with the records definitions and to check runtime that  the correct tracer version is used .

PR #495, opened 2018-06-05
add more device functions for hip-clang

yxsamliu:


include/hip/hcc_detail/hip_complex.h
```diff
@@ -24,7 +24,7 @@ THE SOFTWARE.
 #define HIP_INCLUDE_HIP_HCC_DETAIL_HIP_COMPLEX_H
 
 #include "hip/hcc_detail/hip_vector_types.h"
-#include <math.h>
+#include <cmath>
```
AlexVlx: @yxsamliu I'm not sure this is correct - math is only available in C++, so it'd seem more reasonable to put it under the __cplusplus guard, methinks?

AlexVlx: That should be fine. What is the issue with HIP-Clang? There should only be a __device__ exp in HIP, which should not clash (in C++).

yxsamliu: I am still investigating whether this is a clang bug. Basically, if we have exp defined in std namespace, and we have using std::exp in default name space, a device function calling exp will call this std::exp instead of device exp. It is supposed to check the device attribute but it does not do that.

Probably I should put a "//TODO: Fix bug about device function calling std::exp" there and remove this after the bug is fixed.

include/hip/hcc_detail/hip_runtime.h
```diff
@@ -106,15 +109,16 @@ extern int HIP_TRACE_API;
 #include <hip/hcc_detail/host_defines.h>
 #include <hip/hcc_detail/math_functions.h>
 #include <hip/hcc_detail/device_functions.h>
+#if __HCC___
```
mangupta: There is an extra ```_``` at the end which breaks all texture and surface directed tests on HCC path.

PR #500, opened 2018-06-07
Get channel descritor

luckynikki:


mangupta:
I believe #514 replaces this PR. So it should be fine to close this. @luckynikki Can you please confirm that is indeed true? In case it isn't please reopen & update the PR based on the review comments.

luckynikki:
Hi Maneesh,
#514 is not a replacement for #500. The additional tests with the name: 1) hipGetAlgntoffset2D.cpp  and 2)  hipGetChanDesc.cpp are pushed through the new branch with the PR #514.
Yes, I am fine with the closure of #500.

PR #510, opened 2018-06-11
Fix hipHostMalloc test

saleelk:


saleelk:
Unrelated test failure .
directed_tests/texture/hipBindTexRef1DFetch.tst

tests/src/runtimeApi/memory/hipHostMalloc.cpp
```diff
@@ -72,11 +72,14 @@ void CheckHostPointer(int numElements, int* ptr, unsigned eventFlags, int syncMe
 
     const int expected = 13;
 
+    int *dPtr;
+    HIPCHECK(hipHostGetDevicePointer((void**)&dPtr, ptr, 0));
```
mangupta: why do we need this? nvcc does not seem to require this nor does hcc require this. If hip-clang requires this to work correctly then it is probably something that should be fixed there. If you still need it here, it is probably best to guard it under the requisite ```#ifdef```.

PR #511, opened 2018-06-11
Adding struct test for hipLaunchKernel()

Srinivasuluch:


gargrahul:
@Srinivasuluch Why not create a new test for this instead of modifying existing one?

Srinivasuluch:
Hi @gargrahul  - I thought of keeping all the hipLaunchkernel() tests in one place so that it will be easy for some one to go through it.

gargrahul:
yes @Srinivasuluch , this looks fine.

Can you remove following includes?
#include "hip/hip_runtime_api.h"
#include <iostream>

Srinivasuluch:
Hi @gargrahul  - Done (cleaned)

yxsamliu:
Can you add verification that the struct kernel arguments are correctly read by the kernel?

Also we need to test char type struct member and structs with padding, and structs with bitfields.

e.g.

struct A { char c1; long l1; char c2; long l2;}
struct A { char bf1: 1; char bf2 : 1; long l1; char bf3 : 1;}

tests/src/kernel/hipLaunchParm.cpp
```diff
@@ -25,11 +25,61 @@ THE SOFTWARE.
...
+  char bf1;
+  char bf2;
+  long l1;
+  char bf3;
+  bool result;  // default is false, will be set to true if the condition is met
+} hipLaunchKernelStruct_t3;
+
+
+// Passing struct to a hipLaunchKernel(), read and write into the same struct
+__global__ void hipLaunchKernelStructFunc1(hipLaunchParm lp, hipLaunchKernelStruct_t1* hipLaunchKernelStruct_) {
```
yxsamliu: we want to test passing struct by value, not by pointer. same as below

tests/src/kernel/hipLaunchParm.cpp
```diff
@@ -25,11 +25,140 @@ THE SOFTWARE.
...
 #include "test_common.h"
-#include "hip/hip_runtime_api.h"
-#include <iostream>
 
-__global__ void vAdd(hipLaunchParm lp, float* a) {}
+#define DISABLE_TEST 0
+
+static const int  BLOCK_DIM_SIZE = 1024;
+
+// This test is to verify Struct with variables to check the hipLaunchKernel() support, read and write into the same struct
```
yxsamliu: should remove the comment "read and write into the same structure" as we only read from the struct passed by value. same as below.

tests/src/kernel/hipLaunchParm.cpp
```diff
@@ -25,11 +25,140 @@ THE SOFTWARE.
...
+    result_d3[x] =  ((hipLaunchKernelStruct_.bf1 == 1) && (hipLaunchKernelStruct_.bf2 == 1)
+                                           && (hipLaunchKernelStruct_.l1 == 1.0) && (hipLaunchKernelStruct_.bf3 == 1) ) ? true : false;
+}
+
+// Passing empty struct to a hipLaunchKernel(), check the size of 1Byte, set the result_d4 to true if condition met
+__global__ void hipLaunchKernelStructFunc4(hipLaunchParm lp, hipLaunchKernelStruct_t4 hipLaunchKernelStruct_, bool* result_d4) {
+    int x = blockIdx.x * blockDim.x + threadIdx.x;
+
+    // set the result to true if the condition met
+    result_d4[x] =  (sizeof(hipLaunchKernelStruct_) == 1) ? true : false;
```
yxsamliu: redundant ? true : false, same as below

tests/src/kernel/hipLaunchParm.cpp
Srinivasuluch: Done

tests/src/kernel/hipLaunchParm.cpp
```diff
@@ -25,11 +25,140 @@ THE SOFTWARE.
 
 #include "hip/hip_runtime.h"
 #include "test_common.h"
-#include "hip/hip_runtime_api.h"
-#include <iostream>
 
-__global__ void vAdd(hipLaunchParm lp, float* a) {}
+#define DISABLE_TEST 0
```
yxsamliu: should be named as "ENABLE_ALIGNMENT_TEST"?

mangupta:
@Srinivasuluch hipLaunchParam test segfaults in CI. Have you tested this updated test case locally? Can you confirm that it works fine for you.

Srinivasuluch:
Yes Maneesh, it's tested locally on HCC & hip-clang, failing scenarios are disabled for test side. We have ticket for that

tests/src/kernel/hipLaunchParm.cpp
```diff
@@ -25,73 +25,436 @@ THE SOFTWARE.
...
     MY_LAUNCH_WITH_PAREN(hipMalloc(&A, 100), true, "launch2");
 
 #ifdef EXTRA_PARENS_2
     // MY_LAUNCH_WITH_PAREN wraps cmd in () which can cause issues.
-    MY_LAUNCH_WITH_PAREN(hipLaunchKernel(vAdd, dim3(1024), dim3(1), 0, 0, Ad), true, "firstCall");
+    MY_LAUNCH_WITH_PAREN(hipLaunchKernel(vAdd, dim3(1024),
+                         dim3(1), 0, 0, Ad), true, "firstCall");
 #endif
 
+    hipFree((void **)&result_h1);
```
mangupta: Memory allocated via ```hipHostMalloc``` should be freed via ```hipHostFree```. It would be a good idea to wrap all HIP API calls in ```HIPCHECK()``` macro to ensure that we are checking the return error codes from all the APIs.

Srinivasuluch: Hi @mangupta 
Done.
For HIPCHECK() : i have a few more tests ready, will update all the HIP calls as per HIPCHECK() into new taht pull request. Pls excuse for now

mangupta:
@Srinivasuluch Thanks for adding HIPCHECK to the free APIs. One final change and I think we are ready to go. For completeness can you also add HIPCHECK to all the other HIP APIs in the test case?

mangupta:
Closing this since as per @Srinivasuluch #568 covers this as well.

PR #512, opened 2018-06-11
 creating math_constants.h file

deven-amd:
This the HIP equivalent of the CUDA file of the same name.

Currently the contents are limited only to what is needed when
compiling Eigen with HIP. More contents can be added as/when needed.


include/hip/math_constants.h
```diff
@@ -0,0 +1,39 @@
...
+
+// Some standard header files, these are included by hc.hpp and so want to make them avail on both
+// paths to provide a consistent include env and avoid "missing symbol" errors that only appears
+// on NVCC path:
+
+
+#if defined(__HIP_PLATFORM_HCC__) && !defined(__HIP_PLATFORM_NVCC__)
+#include <hip/hcc_detail/math_constants.h>
+#elif defined(__HIP_PLATFORM_NVCC__) && !defined(__HIP_PLATFORM_HCC__)
+//#include <hip/nvcc_detail/math_constants.h>
```
mangupta: There needs to be a nvcc_detail/math_constants.h that maps the HIPRT_* constants to the CUDA equivalent ones.

deven-amd: @mangupta 

The CUDA equivalent of "math_constants.h" is part of the CUDA header files (i.e. it is already present at  `<cuda_install_dir>/include/math_constants.h`)

Do we still need to have a `nvcc_detail/math_constants.h`? If yes, then can we simply copy the one from the CUDA install?

mangupta: @deven-amd Maybe I wasn't clear enough.
Say there is a HIP sample that uses ```HIPRT_INF_F```. When compiling on HIP_PLATFORM_HCC, the macro gets resolved via ```hip/hcc_detail/math_constants.h```. However the same macro will not get resolved when compiling on HIP_PLATFORM_NVCC. What needs to happen is that there needs to be a ```hip/nvcc_detail/math_constants.h``` that maps the HIPRT_* macros to equivalent CUDART_* so that compilation works on both platforms.

mangupta:
@deven-amd Any ETA on addressing the review comments? I could have fixed them myself, but looks like you have deleted the branch from which the PR was made. So I am unable to submit additional commits to this PR.

include/hip/math_constants.h
deven-amd: Ok, now I see what you mean...let me make the update and resubmit this PR.

deven-amd: BTW, what is workflow to submit a second PR from my fork, while another is still outstanding (#541)

gargrahul:
You can use different branches in your fork for different PRs.

deven-amd:
Filed pR #551...closing out this one

PR #515, opened 2018-06-12
Add hipclang amdgcn functions

aaronenyeshi:
These are moving from hipclang in device library to hip headers. These are required for the functionality of HIPclang project.


aaronenyeshi:
So far I have added assertfail, ballot, clock, lanemask, memory, named sync, smid, sync, and trap. These previously existed as .cl and .ll files in https://github.com/RadeonOpenCompute/ROCm-Device-Libs/tree/9b46f23b1795b56600bb39ff2594ce129ab0222e/hip/src

include/hip/hcc_detail/hip_runtime.h
```diff
@@ -656,8 +713,136 @@ __DEVICE__ void inline __assert_fail(const char * __assertion,
...
+                  const char *__function,
+                  size_t charsize)
+{
+    // ignore all the args for now.
+    __device_trap();
+}
+
+// hip.amdgcn.bc - sync threads
+// extern "C" __device__ __attribute__((noduplicate)) void __syncthreads();
+#define CLK_LOCAL_MEM_FENCE    0x01
```
yxsamliu: please prefix with __, since we don't want to pollute users' name space

include/hip/hcc_detail/hip_runtime.h
```diff
@@ -656,8 +713,136 @@ __DEVICE__ void inline __assert_fail(const char * __assertion,
...
+#define local __attribute__((address_space(3)))
```
yxsamliu: prefix local with __

include/hip/hcc_detail/hip_runtime.h
```diff
@@ -656,8 +713,136 @@ __DEVICE__ void inline __assert_fail(const char * __assertion,
...
+
+typedef unsigned cl_mem_fence_flags;
```
yxsamliu: prefix with __

include/hip/hcc_detail/hip_runtime.h
```diff
@@ -656,8 +713,136 @@ __DEVICE__ void inline __assert_fail(const char * __assertion,
...
+
+typedef enum memory_scope {
```
yxsamliu: prefix with __

include/hip/hcc_detail/hip_runtime.h
```diff
@@ -656,8 +713,136 @@ __DEVICE__ void inline __assert_fail(const char * __assertion,
...
+  memory_scope_work_item = __OPENCL_MEMORY_SCOPE_WORK_ITEM,
+  memory_scope_work_group = __OPENCL_MEMORY_SCOPE_WORK_GROUP,
+  memory_scope_device = __OPENCL_MEMORY_SCOPE_DEVICE,
+  memory_scope_all_svm_devices = __OPENCL_MEMORY_SCOPE_ALL_SVM_DEVICES,
+  memory_scope_sub_group = __OPENCL_MEMORY_SCOPE_SUB_GROUP
+} memory_scope;
+
+// enum values aligned with what clang uses in EmitAtomicExpr()
+typedef enum memory_order
```
yxsamliu: prefix with __

include/hip/hcc_detail/hip_runtime.h
```diff
@@ -313,8 +360,11 @@ __device__ int __hip_move_dpp(int src, int dpp_ctrl, int row_mask, int bank_mask
 __host__ __device__ int min(int arg1, int arg2);
 __host__ __device__ int max(int arg1, int arg2);
 
-__device__ void* __get_dynamicgroupbaseptr();
+extern "C" __device__ void* get_dynamic_group_segment_base_pointer();
 
+__device__
+inline
+void* __get_dynamicgroupbaseptr() { return get_dynamic_group_segment_base_pointer(); }
```
yxsamliu: can we remove get_dynamic_group_segment_base_pointer and implement __get_dynamicgroupbaseptr directly? I think the disabled code in __amdgcn_get_dynamicgroupbaseptr is related to this.

include/hip/hcc_detail/hip_runtime.h
```diff
@@ -656,8 +713,158 @@ __DEVICE__ void inline __assert_fail(const char * __assertion,
...
+extern "C" __device__ void __llvm_fence_acq_sys(void);
+
+extern "C" __device__ void __llvm_fence_rel_sg(void);
+extern "C" __device__ void __llvm_fence_rel_wg(void);
+extern "C" __device__ void __llvm_fence_rel_dev(void);
+extern "C" __device__ void __llvm_fence_rel_sys(void);
+
+__device__
+inline
+static void hc_work_group_barrier(__cl_mem_fence_flags flags, __memory_scope scope)
```
yxsamliu: can we rename it to __work_group_barrier?

include/hip/hcc_detail/hip_runtime.h
```diff
@@ -656,8 +713,158 @@ __DEVICE__ void inline __assert_fail(const char * __assertion,
...
+          case __memory_scope_all_svm_devices:  __llvm_fence_acq_sys(); break;
+        }
+    } else {
+        __builtin_amdgcn_s_barrier();
+    }
+}
+
+__device__
+inline
+static void hc_barrier(int n)
```
yxsamliu: can we rename it as __barrier?

yxsamliu:
I think it is cleaner to create two new header files:

llvm_intrinsics.h - containing declarations for wrapper functions for llvm intrinsics like llvm.amdgcn.s.barrier

device_library_decls.h - containing declarations for types and functions in device library

Other changes in this patch can be moved to device_functions.h

aaronenyeshi:
I'm working on separating this into 3 files, llvm_intrinsics.h, device_library_decls.h, and device_functions.h. Thanks @yxsamliu

aaronenyeshi:
It doesn't seem that

aaronenyeshi:
I've rebased commits on master, and removed device_functions.cpp and hip_hc.ll and added them into HIP headers. Also working on removing ISA asm.

aaronenyeshi:
I'm not sure why 2 tests are failing: directed_tests/deviceLib/hipTestHalf and directed_tests/deviceLib/hipTestNativeHalf

include/hip/hcc_detail/device_library_decls.h
```diff
@@ -0,0 +1,62 @@
...
+extern "C" __device__ uint __ockl_sad_u32(uint, uint, uint);
+
+extern "C" __device__ float __ocml_floor_f32(float);
+extern "C" __device__ float __ocml_rint_f32(float);
+extern "C" __device__ float __ocml_ceil_f32(float);
+extern "C" __device__ float __ocml_trunc_f32(float);
+
+// Introduce local address space
+#define __local __attribute__((address_space(3)))
+__device__ inline static __local char* __to_local(unsigned x) { return (__local char*)x; }
```
mangupta: HCC throws the warning:
```warning: cast to '__attribute__((address_space(3))) char *' from smaller integer type 'unsigned int' [-Wint-to-pointer-cast]```

Comment (aaronenyeshi) [reply include/hip/hcc_detail/device_library_decls.h]: Should I hide this warning by adding CPP flag -Wno-int-to-pointer-cast ?

Comment (scchan) [reply include/hip/hcc_detail/device_library_decls.h]: The signature of this function doesn't look right to me.  I would expect `x` to be a generic pointer which is 64-bit.  Is there a unit test for this

Comment (yxsamliu) [reply include/hip/hcc_detail/device_library_decls.h]: This is a workaround for a compiler bug. It relies on compiler inserting an addrspacecast when calling this function.

Aaron, since this relies on a compiler workaround which is not in amd-common branch, there is no point of putting this change into HIP. Can you separate this change (and those depending on it) and keep it in hip-clang branch only? Then you can update this pull request without this part of the change.

Comment (yxsamliu) [reply include/hip/hcc_detail/device_library_decls.h]: sorry I mistaken this function with another function.

This is not the OpenCL __to_local function. It is a static function which is called by  __get_dynamicgroupbaseptr.

**Comment by aaronenyeshi** [comment]
@mangupta I've merged to master resolving conflicts. Also I've added a fix for the warning message by using larger integer type before addrspacecast. Passing on my end and Jenkins now.

@scchan Thanks, I've fixed the arg `x` to be a generic unsigned 64 bit int before the addrspacecast now.

@yxsamliu It was the `__local_to_generic` function that encounters clang bug with addrspacecast. `__to_local` works fine in C++ compiler.

File: include/hip/hcc_detail/device_library_decls.h (hunk 9 above)
Comment (yxsamliu) [reply include/hip/hcc_detail/device_library_decls.h]: The warning comes from host compile stage. Although there is no IR emitted for this device function in host compilation stage, it still needs to be parsed. Then a pointer in addrspace 3 is 64 bit for host, which causes the warning. Since this function is valid only for device compilation, we should guard it with __HIP__DEVICE_COMPILE.

File: include/hip/hcc_detail/device_library_decls.h [hunk 10]
```diff
@@ -45,7 +45,8 @@ extern "C" __device__ float __ocml_trunc_f32(float);
 
 // Introduce local address space
 #define __local __attribute__((address_space(3)))
-__device__ inline static __local char* __to_local(unsigned x) { return (__local char*)x; }
+// TODO: Remove cast to uint64_t after clang bug is fixed. Should use uint32_t
+__device__ inline static __local char* __to_local(uint64_t x) { return (__local char*)x; }
```
scchan: should the return type be `void*` instead of `char*`?

include/hip/hcc_detail/device_library_decls.h
aaronenyeshi: I will guard the original __to_local with `__HIP_DEVICE_COMPILE__` since a local pointer should be the same size as uint32_t on device side.

include/hip/hcc_detail/device_library_decls.h
aaronenyeshi: Thanks, I will change that

include/hip/hcc_detail/device_library_decls.h
aaronenyeshi: Seems if I surround __to_local with that guard, I will need to surround get_dynamicgroupbaseptr caller functions as well.

aaronenyeshi:
@mangupta I've added a few more changes to allow hip-clang path to pass tests: hipSurfaceObj2D, hip_test_ldg, hip_threadfence_system, and hipNullStream. Thanks!

include/hip/hcc_detail/device_functions.h
```diff
@@ -142,72 +516,466 @@ CUDA implements half as unsigned short whereas, HIP doesn't.
...
+int64_t __lanemask_lt()
+{
+    int32_t activelane = __ockl_activelane_u32();
+    int64_t ballot = __ballot64(1);
+    if (activelane == 0)
+        return 0;
+    return ballot;
+}
+
+#ifdef __HIP_DEVICE_COMPILE__
```
yxsamliu: we still need the device functions to be declared for host compilation, therefore the code should be like this
`#ifdef __HIP_DEVICE_COMPILE__
__device__
inline
void* __get_dynamicgroupbaseptr()
{
    // Get group segment base pointer.
    return (char*)__local_to_generic(__to_local(__llvm_amdgcn_groupstaticsize()));
}
#else
__device__
void* __get_dynamicgroupbaseptr();
#endif // __HIP_DEVICE_COMPILE__
`

aaronenyeshi: Hi Sam, I've cherry-picked your fixes into this PR related to your review. Thanks!

aaronenyeshi:
@yxsamliu @scchan @mangupta I believe this is ready to land. Please help review. It passes all unit tests on HCC/HIP, also it passes 97/106 unit tests on HIP-clang on HCC rt. Thank you!

yxsamliu:
Pleas hold on. We need Brian to review it. Thanks.

aaronenyeshi:
@b-sumner - I will make a new commit to remove the old usage of __activelanemake_v4_b64_b1.

include/hip/hcc_detail/device_library_decls.h
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+#ifndef HIP_INCLUDE_HIP_HCC_DETAIL_DEVICE_LIBRARY_DECLS_H
+#define HIP_INCLUDE_HIP_HCC_DETAIL_DEVICE_LIBRARY_DECLS_H
+
+#include "hip/hcc_detail/host_defines.h"
+
+extern "C" __device__ int32_t __ockl_activelane_u32(void);
+
+extern "C" __device__ uint __ockl_mul24_u32(uint, uint);
```
b-sumner: Please copy the function attributes these functions are given in ocml.h.

include/hip/hcc_detail/hip_runtime.h
```diff
@@ -235,85 +201,15 @@ __device__ int __hip_move_dpp(int src, int dpp_ctrl, int row_mask, int bank_mask
...
- * @Returns void
- *
- * @param void
- *
- * @warning __threadfence_system is a stub and map to no-op.
- */
-//__device__ void  __threadfence_system(void) __attribute__((deprecated("Provided with workaround
-//configuration, see hip_kernel_language.md for details")));
-__device__ void __threadfence_system(void);
+__device__ inline static int min(int arg1, int arg2) {
```
b-sumner: These can't work.  An f32 can not exactly represent every i32.  Just use arg1 < arg2 ? arg1 : arg2.  The compiler will do the rest.

include/hip/hcc_detail/llvm_intrinsics.h
```diff
@@ -0,0 +1,68 @@
...
+ *  @brief Contains declarations for wrapper functions for llvm intrinsics
+ *         like llvm.amdgcn.s.barrier.
+ */
+
+#ifndef HIP_INCLUDE_HIP_HCC_DETAIL_LLVM_INTRINSICS_H
+#define HIP_INCLUDE_HIP_HCC_DETAIL_LLVM_INTRINSICS_H
+
+#include "hip/hcc_detail/host_defines.h"
+
+__device__ ulong __llvm_amdgcn_icmp_i32(uint x, uint y, uint z) __asm("llvm.amdgcn.icmp.i32");
```
b-sumner: This should have the convergent attribute if it is really needed.

include/hip/hcc_detail/device_functions.h
```diff
@@ -142,72 +516,457 @@ CUDA implements half as unsigned short whereas, HIP doesn't.
...
-__device__ int __float2int_rd(float x);
-__device__ int __float2int_rn(float x);
-__device__ int __float2int_ru(float x);
-__device__ int __float2int_rz(float x);
+__device__ static inline int __float2int_rd(float x) { return (int)__ocml_floor_f32(x); }
+__device__ static inline int __float2int_rn(float x) { return (int)__ocml_rint_f32(x); }
+__device__ static inline int __float2int_ru(float x) { return (int)__ocml_ceil_f32(x); }
+__device__ static inline int __float2int_rz(float x) { return (int)__ocml_trunc_f32(x); }
+
+__device__ static inline long long int __float2ll_rd(float x) { return (long long int)x; }
```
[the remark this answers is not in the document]
aaronenyeshi: I copied these conversion functions from old HIP implementation. How can I fix it?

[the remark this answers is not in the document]
yxsamliu: I think the correct versions cause isel failures in backend. Probably just add ToDo.

aaronenyeshi:
Thanks @b-sumner, I've added the missing attributes, changed llvm.amdgcn.icmp.i32 asm and changed min/max as you suggested. For the conversion functions, I've left a TODO comment for the future.

b-sumner:
Looks good to me.  I will work on the conversion functions after this is checked in.

aaronenyeshi:
@mangupta ping - we need this merged for internal HIP testing and investigation.

Issue #467, opened 2018-05-24
Error while building HIP using ninja


pradeepisro:
Error:
`ninja: error: build.ninja:2433: bad $-escape (literal $ must be written as $$)`

Build Command:
`ccmake -G Ninja ..`

But there were no errors while using `make`
I tried fixing manually by turning $(VERBOSE) to $$(VERBOSE)
It worked.

Fix:
It can be fixed by changing `add_custom_target(build_tests DEPENDS directed_tests)` to `add_custom_target(build_tests DEPENDS ${directed_tests})` in the CMakeLists.txt

mangupta:
@pradeepisro #522 has been merged. so this issue should be fixed now I guess. Can you confirm and close the ticket?

PR #522, opened 2018-06-19
added changes to FindHIP.cmake which would allow us to build using ninja #467

pradeepisro:


cmake/FindHIP.cmake
```diff
@@ -471,7 +471,13 @@ macro(HIP_PREPARE_TARGET_COMMANDS _target _format _generated_files _source_files
                 INPUT "${custom_target_script_pregen}"
                 )
             set(main_dep DEPENDS ${source_file})
-            set(verbose_output "$(VERBOSE)")
+            if(CMAKE_GENERATOR MATCHES "Makefiles")
+                set(verbose_output "$(VERBOSE)")
+            elseif(HIP_VERBOSE_BUILD)
```
mangupta: Where is HIP_VERBOSE_BUILD defined? How is it set? Is this an auto-created variable used by Ninja?

pradeepisro: I just update the file. It was actually taken from findCUDA.cmake

PR #525, opened 2018-06-20
Add hipStructKernel test - passing struc to kernel

aaronenyeshi:
This test will pass a structure filled with values to kernel, then verify the values in int array.


mangupta:
@aaronenyeshi @Srinivasuluch I believe the test in this PR is covered by #511. If so, we should be able to close this. Can you confirm that?

aaronenyeshi:
Yes, I think it's covered there, Thanks!

PR #532, opened 2018-06-22
Patch 5

Jorghi12:
This PR includes necessary overloads in order to avoid ambiguous __shfl_* function calls when types other than float/int are passed as the first argument. In the future, the HC / Kalmar precision library will need to add implementations for these __shfl* operations for double precision.


whchung:
@aaronenyeshi perhaps you may want to augment your patch at: https://github.com/ROCm-Developer-Tools/HIP/commit/2a7ea47614824c440f2c081571cc23852674d821 ?

aaronenyeshi:
Thanks Jack, let me add the double precision ones to my commit as well.

Jorghi12:
Thanks @aaronenyeshi.

PR #536, opened 2018-06-25
Removes use of unimplemented OCML functionality.

AlexVlx:
Fixes SWDEV-157088.


sunway513:
@AlexVlx I believe you were referring to SWDEV-157117? The SWDEV-157088 was identified to be not valid and I've suggested to close it.

AlexVlx:
No I was referring exactly to what I wrote, and has absolutely nothing to do with the other thing you mentioned. It just happens to not manifest in some narrower context, as you observed; let's not mix things up please.

sunway513:
Thanks @AlexVlx , I know what you are referring now -- since that has been worked around in the TF code base, I'll revert it and test against this PR.

PR #537, opened 2018-06-25
Move __hip_hc_memcpy and __hip_hc_memset from device_utils.cpp to dev…

yxsamliu:
…ice_functions.h as inline functions


yxsamliu:
@b-sumner could you please review this? Thanks.

b-sumner:
What are the requirements for the source and destination pointers?  The current implementation assumes that the source and destination are at least 4 bytes apart.

yxsamliu:
This should work for any distance between dst and src since it is used to implement memcpy. I will fix it.

yxsamliu:
@b-sumner If abs(dst-src)<4, the size must be less than 4 since dst and src cannot overlap. In this case the implementation still works due to the switch statement. Is that right?

PR #538, opened 2018-06-25
Include host_defines.h in hip_fp16.h since it uses __host__ __device_…

yxsamliu:
…_ attributes


aaronenyeshi:
@AlexVlx @rocm-hcc Could you review this patch? @mangupta I think this needs to be fixed for D9 SWDEV-157508

PR #541, opened 2018-06-27
Adding long/float/double overloads for min and max.

deven-amd:
Tensorflow codebase expects the presence of float/double overloads for min/max. If not present, an implicit conversion from float/double to int occurs, leading to incorrect results, and hence this addition.

Thought about making this a templatized function, but decided against it because
1. a generic `template<typename T> __device__ T ...` would declare min/max overloads for all types, not sure that is what we want here. Also could not figure out a clean way of limiting the template instantiation to just the three types that we need for the time being.
2. did not want to change the existing implementation for the `int` type.

### DIFF
--- a/include/hip/hcc_detail/hip_runtime.h
+++ b/include/hip/hcc_detail/hip_runtime.h
@@ -238,6 +238,21 @@ __device__ int __hip_move_dpp(int src, int dpp_ctrl, int row_mask, int bank_mask
 __host__ __device__ int min(int arg1, int arg2);
 __host__ __device__ int max(int arg1, int arg2);
 
+__host__ __device__ unsigned int min(unsigned int arg1, unsigned int arg2);
+__host__ __device__ unsigned int max(unsigned int arg1, unsigned int arg2);
+
+__host__ __device__ long min(long arg1, long arg2);
+__host__ __device__ long max(long arg1, long arg2);
+
+__host__ __device__ unsigned long min(unsigned long arg1, unsigned long arg2);
+__host__ __device__ unsigned long max(unsigned long arg1, unsigned long arg2);
+
+__host__ __device__ float min(float arg1, float arg2);
+__host__ __device__ float max(float arg1, float arg2);
+
+__host__ __device__ double min(double arg1, double arg2);
+__host__ __device__ double max(double arg1, double arg2);
+
 __device__ void* __get_dynamicgroupbaseptr();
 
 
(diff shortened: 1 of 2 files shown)


deven-amd:
I will be updating this PR with another commit tomorrow...need to add more overloads to make the TF build happy :(

include/hip/hcc_detail/hip_runtime.h
```diff
@@ -235,8 +235,43 @@ __device__ int __hip_move_dpp(int src, int dpp_ctrl, int row_mask, int bank_mask
 
 #endif  //__HIP_ARCH_GFX803__ == 1
 
-__host__ __device__ int min(int arg1, int arg2);
-__host__ __device__ int max(int arg1, int arg2);
+__host__ __device__ int          min(         int arg1,          int arg2);
+__host__ __device__ unsigned int min(         int arg1, unsigned int arg2);
```
mangupta: This signature looks incorrect to me. ```min(int,unsigned int)``` should be ```int``` in my opinion. For example ```min(-100,20)``` is -100 which is ```int```. Same applies to all other min variants that compare a signed and unsigned value.

deven-amd: yeah, in theory we should not need to have min/max with one arg signed and the other unsigned :)

The reason I needed to add the sign mismatch overloads is because this scenario manifests itself in the TF codebase.  I saw the same overloads on the CUDA side, and am assuming they added those overloads atleast in part to address the scenario in the TF codebase. It would have been nicer if the TF codebase was fixed instead, but we have no control over that.

The return type was chosen to match what CUDA has. Either choice is problematic as it will not cover the entire range of one of the two args. We can switch the return type to signed if matching equivalent CUDA funcs is not that important.

deven-amd: For the sign mis-match overloads, the CUDA implementation simply casts the signed argument to unsigned, and then calls the unsigned overload, which is why the return type is also unsigned

deven-amd: I had rather change the return type to signed as well for those sign mismatch overloads, so if CUDA compatibility is not a requirement / preference, then let me know and I will update the PR accordingly

deven-amd: actually come to think of it, returning `signed` for `min` and `unsigned` for max` is the correct choice. That is what I will go with, if the HIP signature can be different than CUDA

mangupta:
Closing this since most of the min/max variants are now supported by use of templates. If there are any missing, we can add them as a separate PR.

PR #543, opened 2018-06-29
Support std::complex for hip-clang

yxsamliu:
This is for supporting TensorFlow

### DIFF
--- a/bin/hipcc
+++ b/bin/hipcc
@@ -114,14 +114,19 @@ if ($HIP_PLATFORM eq "clang") {
     }
     $ROCM_PATH=$ENV{'ROCM_PATH'} // "/opt/rocm";
     $HIPCC="$HIP_CLANG_PATH/clang++";
-    $HIPCXXFLAGS .= "-std=c++11 -I$HIP_PATH/include";
-    $HIPLDFLAGS = "--hip-link --hip-device-lib-path=$DEVICE_LIB_PATH -L$HIP_PATH/lib -lhip_hcc";
 
     # If $HIPCC clang++ is not compiled, use clang instead
     if ( ! -e $HIPCC ) {
         $HIPCC="$HIP_CLANG_PATH/clang";
-        $HIPLDFLAGS = "--driver-mode=g++ " . $HIPLDFLAGS;
+        $HIPLDFLAGS = "--driver-mode=g++";
     }
+
+    $HIP_CLANG_VERSION = `$HIPCC --version`;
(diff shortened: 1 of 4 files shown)


bin/hipcc
```diff
@@ -114,7 +114,7 @@ if ($HIP_PLATFORM eq "clang") {
     }
     $ROCM_PATH=$ENV{'ROCM_PATH'} // "/opt/rocm";
     $HIPCC="$HIP_CLANG_PATH/clang++";
-    $HIPCXXFLAGS .= "-std=c++11 -I$HIP_PATH/include";
+    $HIPCXXFLAGS .= "-std=c++11 -isystem $HIP_CLANG_PATH/../lib/clang/7.0.0/include -I$HIP_PATH/include";
```
mangupta: Does this need to be hard-coded to clang 7.0? Is there some way to get this path from some clang config/binary?

yxsamliu: Fixed. Thanks.

tests/src/deviceLib/hipStdComplex.cpp
```diff
@@ -0,0 +1,152 @@
...
+all copies or substantial portions of the Software.
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANNTY OF ANY KIND, EXPRESS OR
+IMPLIED, INNCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANNY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER INN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR INN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
```
mangupta: The test does not seem to have a HIT BLOCK. So ctest will not pick this up for testing. Or is this test meant to be run manually?

yxsamliu: Added HIT block. Thanks.

yxsamliu: Made changes to the test. Now it is passing.

PR #553, opened 2018-07-06
hipClassKernel to test class scenarios for hipClang

phani544:
starting with empty class size and ptr checking. More tests to follow.


yxsamliu:
can we have a class with a constructor and destructor with __device__ attribute? Define an automatic variable of the class type in a kernel. Check the constructor and destructor are called by the kernel.

phani544:
<just for tracking> I have a test that does access constructor and destructor from kernel but was throwing error , logged a JIRA ticket for that.

phani544:
this test is covered in another PR#564. this PR can be closed

PR #555, opened 2018-07-06
Fix hip_mbcnt test typo

aaronenyeshi:
Ctest did not catch this test failure. When running this test manually before typo, I got __mbcnt_hi() FAILED! . This fix will fix this test for HCC and HIP clang path.


aaronenyeshi:
Before this fix, running the test manually looks like:
 System minor 0
 System major 3
 agent prop name Device 67df
hip Device prop succeeded 
```
__mbcnt_lo() PASSED!
__mbcnt_hi() FAILED!
__lane_id() PASSED!
```

PR #558, opened 2018-07-06
Fix the FindHIP.cmake script.

Jorghi12:
1) Added the CMake Flags necessary to maintain standard using of regular CMake commands when building a HIP target.
2) When building HIP targets, COMPILE_DEFINTIONS were dropped. So append the COMPILE DEFINITIONS from the directory into the hcc invocation line.
3) Also extract the include directories and pass it into the correct variable.


cmake/FindHIP.cmake
```diff
@@ -529,6 +550,10 @@ endmacro()
 # HIP_ADD_LIBRARY
 ###############################################################################
 macro(HIP_ADD_LIBRARY hip_target)
+    # Add the include directories && compile definitions
+    list(APPEND HIP_HIPCC_INCLUDE_ARGS_USER "-I$<JOIN:$<TARGET_PROPERTY:${hip_target},INCLUDE_DIRECTORIES>, -I>")
+    list(APPEND HIP_HCC_FLAGS "-D$<JOIN:$<TARGET_PROPERTY:${hip_target},COMPILE_DEFINITIONS>, -D>")
```
mangupta: FindHIP supports hcc, nvcc and rudimentary support for hip-clang. So we need to target compile definitions for all paths i.e. atleast hcc and nvcc for now.
FindHIP does this by supporting ```HIP_HIPCC_FLAGS```, ```HIP_HCC_FLAGS``` and ```HIP_NVCC_FLAGS```. Since it is possible that the user may want different definitions to be enabled on each path, we cannot blindly add all ```COMPILE_DEFINITIONS``` to anyone specific compiler path. You should change this to ```HIP_HIPCC_FLAGS``` here since that targets all compiler paths.
It probably makes sense to have the same behavior in ```HIP_ADD_EXECUTABLE``` as well. So it is better to move these two statements to ```HIP_PREPARE_TARGET_COMMANDS```.

Jorghi12: @mangupta Great. So it looks like I'll have to extend my change to fix the NVCC path as well using HIP_HIPCC_FLAGS. I didn't catch that since I've only been using HIP/HCC, never the NVCC path which relies on marshaling within HIP.

I'll send an update!

cmake/FindHIP.cmake
```diff
@@ -383,6 +403,10 @@ macro(HIP_PREPARE_TARGET_COMMANDS _target _format _generated_files _source_files
     HIP_PARSE_HIPCC_OPTIONS(HIP_HCC_FLAGS ${_hcc_options})
     HIP_PARSE_HIPCC_OPTIONS(HIP_NVCC_FLAGS ${_nvcc_options})
 
+    # Add the include directories && compile definitions
+    list(APPEND HIP_HIPCC_INCLUDE_ARGS_USER "-I$<JOIN:$<TARGET_PROPERTY:${_target},INCLUDE_DIRECTORIES>, -I>")
```
mangupta: Appending to ```HIP_HIPCC_INCLUDE_ARGS_USER``` needs to move before line392 where we initialize ```HIP_HIPCC_INCLUDE_ARGS```.

Jorghi12:
@mangupta Changes made.

cmake/FindHIP.cmake
```diff
@@ -370,6 +391,10 @@ macro(HIP_PREPARE_TARGET_COMMANDS _target _format _generated_files _source_files
     # Initialize list of includes with those specified by the user. Append with
     # ones specified to cmake directly.
     set(HIP_HIPCC_INCLUDE_ARGS ${HIP_HIPCC_INCLUDE_ARGS_USER})
+
+    # Add the include directories
+    list(APPEND HIP_HIPCC_INCLUDE_ARGS "-I$<JOIN:$<TARGET_PROPERTY:${_target},INCLUDE_DIRECTORIES>, -I>")
```
mangupta: I didn't notice this before. The expression assumes that ```TARGET_PROPERTY:INCLUDE_DIRECTORIES``` is non-empty. It needs to be checked for non-empty, else a dangling ```-I``` would be appended to ```HIP_HIPCC_INCLUDE_ARGS```. See https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#output-expressions for how to do that.

cmake/FindHIP.cmake
```diff
@@ -383,6 +408,9 @@ macro(HIP_PREPARE_TARGET_COMMANDS _target _format _generated_files _source_files
     HIP_PARSE_HIPCC_OPTIONS(HIP_HCC_FLAGS ${_hcc_options})
     HIP_PARSE_HIPCC_OPTIONS(HIP_NVCC_FLAGS ${_nvcc_options})
 
+    # Add the compile definitions
+    list(APPEND HIP_HIPCC_FLAGS "-D$<JOIN:$<TARGET_PROPERTY:${_target},COMPILE_DEFINITIONS>, -D>")
```
mangupta: Same as above. The expression assumes that ```TARGET_PROPERTY:COMPILE_DEFINITIONS``` is non-empty. It needs to be checked for non-empty, else a dangling ```-D``` would be appended to ```HIP_HIPCC_FLAGS```. See https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#output-expressions for how to do that.

Jorghi12:
I don't have access to the ROCm CI so it's basically a black box from my side. But, I'm assuming your suggested fix doesn't work.

Could you post the error log from your CI?

mangupta:
@Jorghi12 The build issue that is occurring now is due to a typo that got introduced in your last commit. I have fixed it and pushed the fix to your branch. The PR works locally for me. Hopefully it will pass on the CI as well. I will approve and merge it in case it passes.

PR #564, opened 2018-07-11
Adding class size>8bytes tests

phani544:
Adding tests to validate class size >8 bytes along with packing and other tests. This also includes the empty   class tests from other PR as well.

### DIFF
--- a/tests/src/compiler/hipClassKernel.cpp
+++ b/tests/src/compiler/hipClassKernel.cpp
@@ -1,0 +1,270 @@
+/*
+Copyright (c) 2015-Present Advanced Micro Devices, Inc. All rights reserved.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+/* HIT_START
+ * BUILD: %t %s ../test_common.cpp
+ * RUN: %t
+ * HIT_END
+ */
+#include "hipClassKernel.h"
+
+// check sizeof empty class is 1
+__global__ void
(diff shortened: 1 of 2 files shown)


tests/src/compiler/hipClassKernel.cpp
```diff
@@ -0,0 +1,492 @@
...
+__global__ void
+sizeVirtDerMultiClassKernel(bool* result_ecd) {
+  int tid = threadIdx.x + blockIdx.x * blockDim.x;
+  result_ecd[tid] = (sizeof(testSizeDerMulti)%48 == 0);
+}
+
+int main() {
+  bool *result_ecd, *result_eob, *result_ech, *result_eoh;
+  size_t NBOOL = BLOCKS * sizeof(bool);
+  HIPCHECK(hipHostMalloc(reinterpret_cast<void**>(&result_ech),
```
yxsamliu: Many tests have similar steps. Can you write a template function for such tests and call the template function?

tests/src/compiler/hipClassKernel.cpp
```diff
@@ -0,0 +1,492 @@
...
+emptyClassKernelObj(bool* result_ecd) {
+  int tid = threadIdx.x + blockIdx.x * blockDim.x;
+  testClassEmpty ob1, ob2;
+  result_ecd[tid] = (&ob1 != &ob2);
+}
+
+__global__ void
+sizeClassP1Kernel(bool* result_ecd) {
+  int tid = threadIdx.x + blockIdx.x * blockDim.x;
+  result_ecd[tid] = (sizeof(testSizeP1)%6 == 0);
```
yxsamliu: why checking sizeof(testSizeP1)%6 == 0? why not checking the sizeof(testSizeP1)==6?

tests/src/compiler/hipClassKernel.h
```diff
@@ -0,0 +1,153 @@
...
+#ifndef COMPILER_HIPCLASSKERNEL_H_
+#define COMPILER_HIPCLASSKERNEL_H_
+
+#include <iostream>
+#include "hip/hip_runtime.h"
+#include "test_common.h"
+
+static const int BLOCKS = 512;
+static const int THREADS_PER_BLOCK = 1;
+static const int ENABLE_DESTRUCTOR_TEST = 0;
```
yxsamliu: Is this test fails for both hcc and hip-clang? If not, should differentiate hcc and hip-clang by using macro __HCC__

tests/src/compiler/hipClassKernel.cpp
```diff
@@ -0,0 +1,492 @@
...
+  testDeviceClass ob2;
+  ob2.iVar = 10;
+}
+#endif
+
+// tests for classes >8 bytes
+__global__ void
+sizeClassKernel(bool* result_ecd) {
+  int tid = threadIdx.x + blockIdx.x * blockDim.x;
+  result_ecd[tid] = (sizeof(testSizeA)%16 == 0);
```
yxsamliu: should check the size itself instead of checking modulo of the size

tests/src/compiler/hipClassKernel.cpp
```diff
@@ -0,0 +1,492 @@
...
+}
+
+__global__ void
+sizeDerivedClassKernel(bool* result_ecd) {
+  int tid = threadIdx.x + blockIdx.x * blockDim.x;
+  result_ecd[tid] = (sizeof(testSizeDerived)%24 == 0);
```
yxsamliu: check size instead of modulo, same as below

tests/src/compiler/hipClassKernel.cpp
phani544: Hi @yxsamliu , I have refactored the kernels and reduced the number to 3.

tests/src/compiler/hipClassKernel.h
phani544: Hi @yxsamliu , Yes. the test fails with both HCC and HIP-Clang, but with different options. Will enable it for HIP-Clang alone when the noopts issue is fixed.

PR #568, opened 2018-07-12
Adding a few more struct scenarions and a ResultValidation()

Srinivasuluch:
Adding more struct scenarions and a ResultValidation()

### DIFF
--- a/tests/src/kernel/hipLaunchParm.cpp
+++ b/tests/src/kernel/hipLaunchParm.cpp
@@ -26,14 +26,60 @@ THE SOFTWARE.
 #include "hip/hip_runtime.h"
 #include "test_common.h"
 
-// Struct with dummy variables to check the hipLaunchKernel()
-typedef struct hipLaunchKernelStruct {
-    int i;
-    float f;
-} hipLaunchKernelStruct_t;
+static const int  STRUCT_SIZE = 1024;
+
+// This test is to verify Struct with variables to check the hipLaunchKernel() support, read and write into the same struct
+typedef struct hipLaunchKernelStruct1 {
+  int li;  // local int
+  float lf;  // local float
+  bool result;  // default is false, will be set to true if the condition is met
+} hipLaunchKernelStruct_t1;
+
+// This test is to verify struct with padding, read and write into the same struct
+typedef struct hipLaunchKernelStruct2 {
+  char c1;  // local char
+  long l1;  // local long
+  char c2;  // local char
+  long l2;  // local long
+  bool result;  // default is false, will be set to true if the condition is met
+} hipLaunchKernelStruct_t2;
+
+typedef struct hipLaunchKernelStruct3 {
+  char bf1;
+  char bf2;
+  long l1;
+  char bf3;
+  bool result;  // default is false, will be set to true if the condition is met
+} hipLaunchKernelStruct_t3;
+
+
+// Passing struct to a hipLaunchKernel(), read and write into the same struct
+__global__ void hipLaunchKernelStructFunc1(hipLaunchParm lp, hipLaunchKernelStruct_t1* hipLaunchKernelStruct_) {
+    int x = blockIdx.x * blockDim.x + threadIdx.x;
+
+    // set the result to true if the condition met
+    hipLaunchKernelStruct_[x].result =  ((hipLaunchKernelStruct_[x].li == 1) && (hipLaunchKernelStruct_[x].lf == 1.0)) ? true : false;
+}
+
+// Passing struct to a hipLaunchKernel(), checks padding, read and write into the same struct
+__global__ void hipLaunchKernelStructFunc2(hipLaunchParm lp, hipLaunchKernelStruct_t2* hipLaunchKernelStruct_) {
+    int x = blockIdx.x * blockDim.x + threadIdx.x;
+
+    // set the result to true if the condition met
+    hipLaunchKernelStruct_[x].result =  ((hipLaunchKernelStruct_[x].c1 == 'a') && (hipLaunchKernelStruct_[x].l1 == 1.0)
+                                          && (hipLaunchKernelStruct_[x].c2 == 'b') && (hipLaunchKernelStruct_[x].l2 == 2.0) ) ? true : false;
+}
+
+// Passing struct to a hipLaunchKernel(), checks padding, read and write into the same struct
+__global__ void hipLaunchKernelStructFunc3(hipLaunchParm lp, hipLaunchKernelStruct_t3* hipLaunchKernelStruct_) {
+    int x = blockIdx.x * blockDim.x + threadIdx.x;
+
+    // set the result to true if the condition met
+    hipLaunchKernelStruct_[x].result =  ((hipLaunchKernelStruct_[x].bf1 == 1) && (hipLaunchKernelStruct_[x].bf2 == 1)
+                                           && (hipLaunchKernelStruct_[x].l1 == 1.0) && (hipLaunchKernelStruct_[x].bf3 == 1) ) ? true : false;
+}
 
 __global__ void vAdd(hipLaunchParm lp, float* a) {}
-__global__ void hipLaunchKernelStructFunc(hipLaunchParm lp, hipLaunchKernelStruct_t* hipLaunchKernelStruct_) {}
 
 //---
 // Some wrapper macro for testing:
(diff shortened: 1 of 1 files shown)


tests/src/kernel/hipLaunchParm.cpp
```diff
@@ -23,75 +23,768 @@ THE SOFTWARE.
...
+    hipLaunchKernelStruct_h7.si = 1;
+    #if ENABLE_ALIGNMENT_TEST  // This is broken on small bar
+    hipLaunchKernelGGL(HIP_KERNEL_NAME(hipLaunchKernelStructFunc7),
+                    dim3(BLOCK_DIM_SIZE),
+                    dim3(1), 0, 0, hipLaunchKernelStruct_h7,
+                    result_d);
+    ResultValidation();
+    #endif
+
+    // Test: Passing struct with packed aligned to 6Bytes
```
yxsamliu: aligned to 4 bytes

tests/src/kernel/hipLaunchParm.cpp
```diff
@@ -23,75 +23,768 @@ THE SOFTWARE.
...
+                    bool* result_d8) {
+    int x = blockIdx.x * blockDim.x + threadIdx.x;
+
+    // set the result to true if the condition met
+    // get the address of the xth element, struct[x],
+    // size_t(p)%4 will be 0 if aligned to 4Byte address space
+    int *p = (int*)(&hipLaunchKernelStruct_);
+    result_d8[x] =  ((hipLaunchKernelStruct_.c1 == 'c')
+                      && (hipLaunchKernelStruct_.si == 1)
+                      && ((size_t(p))%4 ==0) );
```
yxsamliu: also need to check sizeof(hipLaunchKernelStruct_)==6

tests/src/kernel/hipLaunchParm.cpp
```diff
@@ -23,75 +23,768 @@ THE SOFTWARE.
...
+    #if ENABLE_ALIGNMENT_TEST  // This is broken on small bar
+    hipLaunchKernelGGL(HIP_KERNEL_NAME(hipLaunchKernelStructFunc7),
+                    dim3(BLOCK_DIM_SIZE),
+                    dim3(1), 0, 0, hipLaunchKernelStruct_h7,
+                    result_d);
+    ResultValidation();
+    #endif
+
+    // Test: Passing struct with packed aligned to 6Bytes
+    hipLaunchKernelStruct_t8 hipLaunchKernelStruct_h8;
```
yxsamliu: can you add another test just to check packed struct size is expected? You can reuse this structure type, just remove the aligned attribute but keep the packed attribute.

yxsamliu: can you also add another test which has struct type argument with an aligned data member?

tests/src/kernel/hipLaunchParm.cpp
```diff
@@ -23,75 +23,768 @@ THE SOFTWARE.
...
+    // allocating memory for char pointer on device
+    HIPCHECK(hipMalloc((void**)&ip_d9, sizeof(uint32_t)));
+    HIPCHECK(hipMemset(ip_d9, 1, sizeof(uint32_t)));
+    // ip_d9 passed as pointer to struct member, struct.ip = &ip_d9
+    const hipLaunchKernelStruct_t9 hipLaunchKernelStruct_h9 = {'c', ip_d9};
+    hipLaunchKernelGGL(HIP_KERNEL_NAME(hipLaunchKernelStructFunc9),
+                    dim3(BLOCK_DIM_SIZE),
+                    dim3(1), 0, 0, hipLaunchKernelStruct_h9,
+                    result_d);
+    #if ENABLE_DECLARE_INITIALIZATION_POINTER
```
yxsamliu: does this test fail with hip-clang? This is unexpected.

tests/src/kernel/hipLaunchParm.cpp
```diff
@@ -23,75 +23,768 @@ THE SOFTWARE.
...
 
+#include <cstdint>
 #include "hip/hip_runtime.h"
 #include "test_common.h"
-#include "hip/hip_runtime_api.h"
-#include <iostream>
 
-__global__ void vAdd(hipLaunchParm lp, float* a) {}
+// Memory alignment is broken
+#define ENABLE_ALIGNMENT_TEST 0
```
yxsamliu: Are the failed tests fail with both hcc and hip-clang?
If not, you can use `#if __HCC__` to differentiate them. Turn off the the test for only hcc or hip-clang.

tests/src/kernel/hipLaunchParm.cpp
```diff
@@ -23,75 +23,768 @@ THE SOFTWARE.
...
+typedef struct hipLaunchKernelStruct14 {
+  int readint;
+  int writeint[BLOCK_DIM_SIZE];  // will write to this from device
+} hipLaunchKernelStruct_t14;
+
+// This test is to verify struct with dynamic memory, new int
+// the heap memory will be accessed from device
+typedef struct hipLaunchKernelStruct15 {
+  char c1;
+  int* heapmem = new int[BLOCK_DIM_SIZE];
```
yxsamliu: device code cannot access host heap memory. This memory has to be allocated by hipMalloc.

tests/src/kernel/hipLaunchParm.cpp
```diff
@@ -23,75 +23,768 @@ THE SOFTWARE.
...
+// This test is to verity user defined STL, simple stack implementation
+typedef struct stackNode {
+    int data;
+    stackNode* nextNode = NULL;
+} stackNode_t;
+typedef struct hipLaunchKernelStruct19 {
+  stackNode_t* stack = NULL;
+  unsigned int size_ = 0;
+  void pushMe(int value) {  // not a device function, setting from host
+    stackNode_t* newNode = new stackNode_t;
```
yxsamliu: device code cannot access host heap memory, so this does not work.
For HCC, there is device function `__hip_hc_malloc` and `__hip_hc_free`, which you may use to implement device function pushMe and popMe. For host code, you may want to use hipMalloc and hipFree.

tests/src/kernel/hipLaunchParm.cpp
```diff
@@ -23,75 +23,768 @@ THE SOFTWARE.
...
+}
+
+// Passing out of order initalized struct, access in-order
+__global__ void hipLaunchKernelStructFunc20(
+                    hipLaunchKernelStruct_t20 hipLaunchKernelStruct_,
+                    bool* result_d20) {
+    int x = blockIdx.x * blockDim.x + threadIdx.x;
+
+    // accessing struct members in order
+    result_d20[x] = (hipLaunchKernelStruct_.name = 'A'
```
yxsamliu: Should = be ==?

tests/src/kernel/hipLaunchParm.cpp
```diff
@@ -23,75 +23,768 @@ THE SOFTWARE.
...
+    // strcmp(hipLaunchKernelStruct_.name, "AMD") -> strcmp is not broken
```
yxsamliu: I am not quite get it. name is a member of char type, why is it compared with a string?

tests/src/kernel/hipLaunchParm.cpp
```diff
@@ -23,75 +23,768 @@ THE SOFTWARE.
...
+    MY_LAUNCH_WITH_PAREN(hipLaunchKernelGGL(vAdd, dim3(1024),
+                         dim3(1), 0, 0, Ad), true, "firstCall");
 #endif
 
+    HIPCHECK(hipHostFree(result_h));
+    HIPCHECK(hipFree(result_d));
+
     passed();
-}
+}
```
yxsamliu: miss newline at end of file

tests/src/kernel/hipLaunchParm.cpp
Srinivasuluch: Typo corrected

tests/src/kernel/hipLaunchParm.cpp
Srinivasuluch: i guess it's 4bytes not 6, sizeof(hipLaunchKernelStruct_)==4.
Done. Add the size validation condition to the kernel

tests/src/kernel/hipLaunchParm.cpp
Srinivasuluch: Done, added two more tests as you suggested, aligned(4) & packed. 
Test names:
hipLaunchKernelStructFunc8A for packed
hipLaunchKernelStructFunc8B for aligned(4)

Results: 
Packed broken
alignment working fine.

Srinivasuluch: To be more specific about aligment member:  with latest changes, alignment is working fine on large bar setup but still it's broken on small bar hence tests disabled

tests/src/kernel/hipLaunchParm.cpp
Srinivasuluch: Yes Sam, this is not failing.

tests/src/kernel/hipLaunchParm.cpp
Srinivasuluch: Yes, failing on both hcc and hip-clang?

tests/src/kernel/hipLaunchParm.cpp
Srinivasuluch: Done

tests/src/kernel/hipLaunchParm.cpp
Srinivasuluch: Done

tests/src/kernel/hipLaunchParm.cpp
Srinivasuluch: initially i tried with a char* name = "AMD" and called strcmp(). As it was not working i changed to a char instead a pointer

tests/src/kernel/hipLaunchParm.cpp
Srinivasuluch: Done

tests/src/kernel/hipLaunchParm.cpp
yxsamliu: strcmp is not a device function, therefore cannot be called by a device function.

tests/src/kernel/hipLaunchParm.cpp
Srinivasuluch: Done

Issue #542, opened 2018-06-28
hipLaunchKernelGGL can launch kernel multiple times


jszuppe:
hipLaunchKernelGGL can launch the same kernel multiple times. Example: [multiple_kernel_runs.zip](https://github.com/ROCm-Developer-Tools/HIP/files/2145466/multiple_kernel_runs.zip)

When the situation occurs:

1. There is a header library `bar` with `bar::bar_impl()` function that launches a kernel.
2.  `bar::bar_impl` is used in another header library `foo` to implement `FooClass::foo()`.
3.  There is dynamic lib `foobar` with `foobar()` func which calls `FooClass::foo()` function.
4. Main application both uses `foobar()` and directly `FooClass::foo()`. Only one call to those functions launches the kernel. In that call instead of one launch, kernel is run twice.

We suspect  that probably in [here](https://github.com/ROCm-Developer-Tools/HIP/blob/836627279fd5bd995910c471882743accef5ec58/src/functional_grid_launch.inl#L126) should be a break after `hipModuleLaunchKernel()`.

See https://github.com/ROCmSoftwarePlatform/rocPRIM/issues/13

AlexVlx:
Thank you for reporting this (and for providing a repro); orthogonally to this issue, the current code has a logical issue (my fault) in that it'd needlessly iterate through the vector even after having found a valid kernel.

PR #575, opened 2018-07-17
It is unclear what I was thinking when authoring the original code...

AlexVlx:


AlexVlx:
This addresses #542 as well.

AlexVlx:
@scchan that'd be a neat change but at this point it'd be much ado about nothing, with the up and coming better future; I'd also not try to brave those treacherous waters given that things sort of work at the moment, yuckiness notwithstanding.

PR #585, opened 2018-07-20
[HIPClang] Fix icmp asm and attributes

aaronenyeshi:
HIP llvm.amdgcn.icmp needs i64 return. This should fix recent llvm trunk changes causing this issue:

Intrinsic name not mangled correctly for type arguments! Should be: llvm.amdgcn.icmp.i64.i32
i64 (i32, i32, i32)* @llvm.amdgcn.icmp.i32
fatal error: error in backend: Broken function found, compilation aborted!
clang: error: clang frontend command failed with exit code 70 (use -v to see invocation)
clang version 4.0
Target: x86_64-unknown-linux-gnu
Thread model: posix

### DIFF
--- a/include/hip/hcc_detail/device_functions.h
+++ b/include/hip/hcc_detail/device_functions.h
@@ -689,13 +689,13 @@ int __any(int predicate) {
 __device__
 inline
 unsigned long long int __ballot(int predicate) {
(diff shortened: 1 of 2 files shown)


aaronenyeshi:
Closing for now

PR #594, opened 2018-07-25
[tests] Fix hipSimpleAtomicsTest on nvcc path

mangupta:
- Test uses c++11. Added it to nvcc options
- testKernelSub was missing __device__ keyword
- unsigned long long atomics don't seem to be supported by nvcc
- atomicAdd does not seem to be supported for double datatype


AlexVlx:
CUDA supports `unsigned long long` atomics just fine, the implementation was done starting from the spec, see: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#arithmetic-functions. Same thing with `atomicAdd` for `double`, so what is the actual issue here, except for the missing `__device__`?

PR #596, opened 2018-07-25
[tests] Fixed hipLaunchParm test on nvcc path

mangupta:
- Uses c++11 features. Added it to nvcc options
- Arguments for some kernels exceeded 4096 bytes which is the limit
  imposed by nvcc. Reduced BLOCK_DIM_SIZE to 512 to handle this
- Fixed compilation issues on nvcc path


Srinivasuluch:
Hi @mangupta  - I am fine with the PR. Thank you for the changes

PR #601, opened 2018-07-26
Remove device mapping from shareWithAll memory

sarunyap:
When shareWithAll memory (e.g., host memory) is allocated, set appId
in hc::AmPointerInfo to -1 to indicate that this memory is not mapped
to any device.  Peer checking in ihipStream_t::canSeeMemory is not
necessary if memory is shared with all devices.  Thus, it is skipped.

Note that earlier host memory is always mapped to device 0 and HIP
always performs peer checking for all kinds of hipMemcpy.  Since the
peer checking process requires context locking, hipMemcpy from/to host
memory always grabs device 0's context lock.  Therefore, if there is
another thread holding the context lock of device 0 (e.g.,
hipDeviceSynchronize on device 0), hipMemcpy will have to wait for the
lock until it can actually perform memcpy.  This can significantly
deteriorate execution performance.

Signed-off-by: Sarunya Pumma <sarunya.pumma@amd.com>

### DIFF
--- a/src/hip_hcc.cpp
+++ b/src/hip_hcc.cpp
@@ -1808,7 +1808,7 @@ bool ihipStream_t::canSeeMemory(const ihipCtx_t* copyEngineCtx, const hc::AmPoin
     // TODO - pointer-info stores a deviceID not a context,may have some unusual side-effects here:
     if (dstPtrInfo->_sizeBytes == 0) {
         return false;
-    } else {
+    } else if (dstPtrInfo->_appId != -1) {
 #if USE_APP_PTR_FOR_CTX
         ihipCtx_t* dstCtx = static_cast<ihipCtx_t*>(dstPtrInfo->_appPtr);
 #else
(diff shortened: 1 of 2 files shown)


src/hip_hcc.cpp
```diff
@@ -1935,9 +1935,9 @@ void ihipStream_t::resolveHcMemcpyDirection(unsigned hipMemKind,
             throw ihipException(hipErrorRuntimeOther);
     };
 
-    if (srcPtrInfo->_isInDeviceMem) {
+    if (srcPtrInfo->_isInDeviceMem && srcPtrInfo->_appId != -1) {
```
bensander: Is there a case where !_isInDeviceMem and appId != -1 ?  Wonder if the second comparison is redundant.

gargrahul: @bensander It is not possible to have _appId = -1 and _isInDeviceMem = true.

sarunyap: @bensander @gargrahul I will remove the second check and update the PR. Thanks.

sarunyap:
@bensander @gargrahul Hi, I have fixed my PR according to your comments.  Could you please review it again?  Thanks!

whchung:
If possible I’d like to have the PR reviewed and merged into upcoming ROCm 1.9 release candidates soon so we can enjoy perf improvements for certain models in mGPU setups :)

PR #603, opened 2018-07-27
Drop use of NVCC_COMPAT since it is always set

b-sumner:


yxsamliu:
Can you document this in the programming guide https://github.com/ROCm-Developer-Tools/HIP/blob/master/docs/markdown/hip_programming_guide.md

b-sumner:
@yxsamliu I'm not sure I follow.  NVCC_COMPAT is being removed and was not documented to begin with as far as I can tell.  What should be added to the programming guide?

yxsamliu:
Ben suggested to mention in programming guide that these functions are slow due to compatibility with CUDA and they have faster alternatives.

b-sumner:
Can we leave that for later or even an optimization guide?  The current doc doesn't mention other cases where faster alternatives are available.  Also, I have no information that the likelihood of the argument being 0 is much lower than the likelihood of other arguments.

b-sumner:
Who can merge this?  It appears to be approved.

aaronenyeshi:
I believe either Maneesh, Rahul or Ben

bensander:
Someone please open a ticket to track the performance recommendation.

On Jul 31, 2018, at 2:32 PM, Aaron Enye Shi <notifications@github.com<mailto:notifications@github.com>> wrote:


I believe either Maneesh, Rahul or Ben

PR #611, opened 2018-07-31
Enable HIP_PLATFORM option for HIP-clang as clang

aaronenyeshi:
We don't want to link libraries that are tied to HCC for the HIP-Clang path. This will help remove some of the flags that are passed in from linking to hc::hccrt such as -hc.


aaronenyeshi:
Let's not add clang as HIP_PLATFORM. We will stay with hcc and nvcc for now, likely move to rocm and nvcc in the future.

PR #614, opened 2018-08-01
Add overloading resolution functions for fma

yxsamliu:


b-sumner:
I'm just wondering what is wrong with reporting an error.  Can't we expect the developer to pass properly typed arguments?  Also, what will we do about half precision?

yxsamliu:
std::numeric_limits<half>::is_specialized is false, therefore the overloading template has no effect on function calls with half argument.

It is a nuance for user to manually cast the arguments to double. This feature has been requested before.

cuda-clang has similar overloading resolution mechanism.

b-sumner:
Are we sure we want the double function, even if all the arguments are bool?  Don't we want the float version if all the arguments can be losslessly cast to float?  Similarly, for fma(char, half, char) don't we want the half version?

yxsamliu:
I updated the PR for better control of promoting to float fma or double fma. If float can hold all three arguments, the float fma is used, otherwise double fma is used.

For half type, since std::numeric_limits is not defined, the template does not apply. However, the standard overloading resolution still applies, so it does not make the situation worse.

b-sumner:
Consuder these two cases:
    fma(float_var, (nt)3, float_var) --> fma(float, float, float)
    fma(float_var, (int)0x2000001, float_var) --> fma(double, double, double)  because 0x2000001 can't be exactly represented by a float

Now, a double can represent every [u]int, but not every [u]long, but with few enough non-zero bits, a float can exactly represent some [u]long.

I don't know how complex you want to make it, but the fma argument type should probably be chosen to exactly represent the original arguments as often as possible.

include/hip/hcc_detail/math_functions.h
```diff
@@ -22,6 +22,7 @@ THE SOFTWARE.
 
 #pragma once
 
+#include "hip_fp16_math_fwd.h"
```
mangupta: This causes build failures in the ```directed_tests/deviceLib/hipTestIncludeMath```. To fix this should either be included after ```hip/hcc_details/host_defines.h``` or we should modify ```hip_fp16_math_fwd.h``` to include ```host_defines.h```.

yxsamliu:
updated PR. conflicts resolved.

PR #619, opened 2018-08-03
Revert pr457

whchung:
HIP PR#457 has been identified not only breaks 1 test case in PaddlePaddle, but also at least 5 unit tests in TensorFlow. After some consideration it should be abandoned.

Propose to change the default HIP runtime behavior so kernel symbol resolution is deferred until 1st kernel launch, instead of initialization before main(). This would be filed as separate PR.

For benchmarking programs which requires low setup time, implement an env var to reinstate eager initialization. This would be filed as separate PR by @sunway513.

### DIFF
--- a/include/hip/hcc_detail/program_state.hpp
+++ b/include/hip/hcc_detail/program_state.hpp
@@ -93,12 +93,11 @@ public:
     }
 };
 
-const std::unordered_map<hsa_agent_t, std::vector<hsa_executable_t>>& executables(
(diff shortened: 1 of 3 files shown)


sunway513:
PR #621 is filed, with that HIP will disable the startup loader by default.

sunway513:
@gargrahul @mangupta as we've discussed on the meeting, could you help merge it to the master branch first?

PR #621, opened 2018-08-04
Disable startup loader by default and guard with env var

sunway513:
This PR changed the HIP default behavior and disabling the startup loader by default.
Users can turn it on by setting the environment variable as follows:
`export HIP_STARTUP_LOADER=1`
If the environment variable is not set or is set to other values, including '0', the startup loader will be disabled.

### DIFF
--- a/src/program_state.cpp
+++ b/src/program_state.cpp
@@ -516,16 +516,11 @@ hsa_executable_t load_executable(const string& file, hsa_executable_t executable
 // HIP startup kernel loader logic
 // When enabled HIP_STARTUP_LOADER, HIP will load the kernels and setup
 // the function symbol map on program startup
-class startup_kernel_loader {
-   private:
-    startup_kernel_loader() { functions(); }
-    startup_kernel_loader(const startup_kernel_loader&) = delete;
-    startup_kernel_loader& operator=(const startup_kernel_loader&) = delete;
-    static startup_kernel_loader skl;
-};
-
 extern "C" void __attribute__((constructor)) __startup_kernel_loader_init() {
-    if (atoi(std::getenv("HIP_STARTUP_LOADER")) == 1) functions();
+    int hip_startup_loader=0;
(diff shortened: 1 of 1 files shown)


sunway513:
@gargrahul @mangupta as we've discussed on the meeting, could you help merge it to master branch first?

src/program_state.cpp
```diff
@@ -513,15 +513,22 @@ hsa_executable_t load_executable(const string& file, hsa_executable_t executable
...
-// symbol map on program startup
+// HIP startup kernel loader logic
+// When enabled HIP_STARTUP_LOADER, HIP will load the kernels and setup
+// the function symbol map on program startup
 class startup_kernel_loader {
    private:
     startup_kernel_loader() { functions(); }
     startup_kernel_loader(const startup_kernel_loader&) = delete;
     startup_kernel_loader& operator=(const startup_kernel_loader&) = delete;
     static startup_kernel_loader skl;
```
whchung: where would `    static startup_kernel_loader skl;` be instantiated? if it's not instantiated anywhere should this be removed?

whchung: another idea is to let `skl` be a pointer, like: `static startup_kernel_loader*`.

and in the shared lib constructor, instead of calling `functions()` directly, use `new skl` to make sure the default constructor of `startup_kernel_loader()` is invoked, and `*skl` points to the instance of the startup kernel loader.

src/program_state.cpp
```diff
@@ -513,15 +513,21 @@ hsa_executable_t load_executable(const string& file, hsa_executable_t executable
     return executable;
 }
 
-// To force HIP to load the kernels and to setup the function
-// symbol map on program startup
+// HIP startup kernel loader logic
+// When enabled HIP_STARTUP_LOADER, HIP will load the kernels and setup
+// the function symbol map on program startup
 class startup_kernel_loader {
```
whchung: is it possible to remove `startup_kernel_loader` altogether?

PR #622, opened 2018-08-07
adding activity prof layer

eshcherb:
API callback/Activity profiling layer


eshcherb:
What is the reason for tests to fail, is it possible to restart the testing?

eshcherb:
How I can run the tests locally?

PR #635, opened 2018-08-17
Fix build failure of hipTestHalf and hipTestIncludeMath for hip-clang

yxsamliu:
This patch is needed for building TensorFlow 1.8 with hip-clang.

### DIFF
--- a/include/hip/hcc_detail/device_functions.h
+++ b/include/hip/hcc_detail/device_functions.h
@@ -27,11 +27,12 @@ THE SOFTWARE.
 #include "math_fwd.h"
 
 #include <hip/hip_runtime_api.h>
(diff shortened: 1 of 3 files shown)


include/hip/hcc_detail/hip_fp16.h
```diff
@@ -29,6 +29,8 @@ THE SOFTWARE.
     #include <utility>
 #endif
 
+#include "hip/hcc_detail/hip_fp16.h"
```
mangupta: This looks weird. The header is explicitly including itself. Maybe the wrong include here?

yxsamliu:
@mangupta issue fixed. Thanks.

PR #637, opened 2018-08-17
Let hipcc link -lpthread -lm by default

yxsamliu:
rocRAND assumes hipcc linking with -lpthread -lm by default.

### DIFF
--- a/bin/hipcc
+++ b/bin/hipcc
@@ -725,7 +725,7 @@ if ($needLDFLAGS and not $compileOnly) {
 }
 $CMD .= " $toolArgs";
 if ($needLDFLAGS and not $compileOnly and $HIP_PLATFORM eq "clang") {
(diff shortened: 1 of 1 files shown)


aaronenyeshi:
Thanks Sam, does this fix this PR https://github.com/ROCmSoftwarePlatform/rocRAND/pull/29

yxsamliu:
yes

ax3l:
This patch is very unstable and does not work with modern CMake, which directly uses `clang -x hip`.
rocRAND (and all other ROC/HIP packages) should publicly declare their dependencies instead of hoping that the `hipcc` wrapper is used to fix missing declarations of dependencies.

X-ref: https://github.com/ROCmSoftwarePlatform/rocRAND/pull/29#issuecomment-912815457

PR #639, opened 2018-08-17
Check for hipEnvVar at the same level if directed_tests location fails

saleelk:


saleelk:
Added more things in that PR

tests/src/runtimeApi/stream/hipStreamSync2.cpp
```diff
@@ -178,7 +178,7 @@ void runTests(int64_t numElements) {
     {
         test(0x01, C_d, C_h, numElements, syncNone, true /*expectMismatch*/);
         test(0x02, C_d, C_h, numElements, syncNullStream, false /*expectMismatch*/);
-        test(0x04, C_d, C_h, numElements, syncOtherStream, true /*expectMismatch*/);
+        test(0x04, C_d, C_h, numElements, syncOtherStream, false /*expectMismatch*/);
```
mangupta: This change causes the test to fail on both rocm-1.9.x and rocm-head.

saleelk: Can you leave this out and merge the rest of the tests?

PR #642, opened 2018-08-22
Let hipcc handle obj files in linker response file for hip-clang

yxsamliu:
If obj files in linker response file contains device code, pass them to hip-clang, otherwise keep them in the linker response file.

This patch is required to build TensorFlow 1.8 with hip-clang.


mangupta:
jenkins retest this please

PR #645, opened 2018-08-27
Roctracer hip frontend 180826

eshcherb:


src/hip_hcc_internal.h
```diff
@@ -299,6 +299,11 @@ extern uint64_t recordApiTrace(std::string* fullStr, const std::string& apiStr);
     HIP_INIT()                                                                                     \
     API_TRACE(0, __VA_ARGS__);
 
+#define HIP_INIT_CB_API(cid, ...)                                                                  \
```
mangupta: Was there any reason to create a new macro for this? It should have been possible to use ```HIP_INIT_API``` itself. Also ```cid``` is nothing but the function name in all the cases, you could have used ```__func__``` for this.

eshcherb: __func__ is expanding to the string which has the name of the function.
I'm using the macro 'cid' parameter for generating API ID and callbacks args data filling macro. The macro saves the API arguments.
Of cause I can use HIP_INIT_API, I can use __func__ string converting it to ID using some predefined map<string, int> and also I can stringify the arguments with variadic macro and safe them but it looks like requires some strings comparison, depending on number of arguments.
I tried to minimize overhead from tracing.
The using of cid macro parameter looks much simpler and more effective, maybe as an alternative I can just add the cid macro parameter to HIP_INIT_API macro itself, I have some script which I can use for this.

..../include/hip/hip_prof_str.h
// HIP API callbacks args data filling macros
#define INIT_hipHostFree_CB_ARGS_DATA(cb_data) { \
  cb_data.args.hipHostFree.ptr = (void*)ptr; \
};
#define INIT_hipMemcpyToSymbolAsync_CB_ARGS_DATA(cb_data) { \
  cb_data.args.hipMemcpyToSymbolAsync.symbolName = (const void*)symbolName; \
  cb_data.args.hipMemcpyToSymbolAsync.src = (const void*)src; \
  cb_data.args.hipMemcpyToSymbolAsync.sizeBytes = (size_t)count; \
  cb_data.args.hipMemcpyToSymbolAsync.offset = (size_t)offset; \
  cb_data.args.hipMemcpyToSymbolAsync.kind = (hipMemcpyKind)kind; \
  cb_data.args.hipMemcpyToSymbolAsync.stream = (hipStream_t)stream; \
};
#define INIT_hipMallocPitch_CB_ARGS_DATA(cb_data) { \
  cb_data.args.hipMallocPitch.ptr = (void**)ptr; \
  cb_data.args.hipMallocPitch.pitch = (size_t*)pitch; \
  cb_data.args.hipMallocPitch.width = (size_t)width; \
  cb_data.args.hipMallocPitch.height = (size_t)height; \
};
#define INIT_hipMalloc_CB_ARGS_DATA(cb_data) { \
  cb_data.args.hipMalloc.ptr = (void**)ptr; \
  cb_data.args.hipMalloc.size = (size_t)sizeBytes; \
};
. . .
#define INIT_CB_ARGS_DATA(cb_id, cb_data) INIT_##cb_id##_CB_ARGS_DATA(cb_data)

..../include/hip/hip_prof_api.h
// HIP API callbacks spawner object macro
#define HIP_CB_SPAWNER_OBJECT(CB_ID) \
  hip_api_data_t api_data{}; \
  INIT_CB_ARGS_DATA(CB_ID, api_data); \
  api_callbacks_spawner_t __api_tracer(HIP_API_ID_##CB_ID, api_data);

mangupta: Using __func__ does not make sense then. However, it is fine to keep using ```HIP_INIT_API``` instead of creating a new one.

include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -2528,6 +2528,24 @@ hipError_t hipLaunchByPtr(const void* func);
 } /* extern "c" */
 #endif
 
+#include <hip/hip_prof_api.h>
```
mangupta: Unless you plan for roctracer to support nvcc path, this should be ```hip/hcc_detail/hip_prof_api.h```.

include/hip/hip_prof_api.h
```diff
@@ -0,0 +1,199 @@
+// automatically generated sources
```
mangupta: Rename file: ```include/hip/hip_prof_api.h``` -> ```include/hip/hcc_detail/hip_prof_api.h```.

include/hip/hip_prof_api.h
```diff
@@ -0,0 +1,199 @@
...
+#ifndef _HIP_PROF_API_H
+#define _HIP_PROF_API_H
+
+#include <atomic>
+#include <iostream>
+#include <mutex>
+
+#include "hip/hip_prof_str.h"
```
mangupta: Again, assuming roctracer is only for rocm path, this should be ```hip/hcc_detail/hip_prof_str.h```.

include/hip/hip_prof_str.h
```diff
@@ -0,0 +1,2450 @@
+// automatically generated sources
```
mangupta: Rename file: ```include/hip/hip_prof_str.h``` -> ```include/hip/hcc_detail/hip_str_api.h```.

src/hip_intercept.cpp
```diff
@@ -0,0 +1,49 @@
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+#include "hip/hip_runtime.h"
+#include "hip/hip_prof_api.h"
```
mangupta: Should be ```hip/hcc_detail/hip_prof_api.h```

mangupta:
@eshcherb LGTM. However, it would be nice if you can add some comments to ```hip_prof_api.h``` or any other header file which documents what we need to do to add profiling for any new APIs that we will support in the future. For example there are already some new APIs that are in the master branch (such as ```hipStreamCreateWithPriority```) for which profiling support needs to be added.

eshcherb:
I have created rebased PR#760
For new APIs you will include teh API name in HIP_INIT_API macro as the first argument and I'm going to provide a python script for regenerating of hip_api_str.h fille with callbacks structures and macros.
I will add commrnts to hip_prof_(str/api).h

mangupta:
closing this since #760 is a rebase of this PR.

PR #651, opened 2018-08-29
Add clock() and clock64()

yxsamliu:
These functions are defined in CUDA SDK and used by Eigen.


b-sumner:
LGTM

yxsamliu:
I updated the PR to add a test for clock functions, and to make them work for current HCC.

Once device lib contains a generic clock function, we can update HIP clock functions to use the device lib clock function.

PR #652, opened 2018-08-29
Add environment variable HIP_AUTO_INCLUDE_HEADER for hip-clang

yxsamliu:
When HIP_AUTO_INCLUDE_HEADER is set to non-zero value, hip/hip_runtime.h is automatically include in
the beginning of each HIP source file.

This is required by TensorFlow 1.8.


mangupta:
/cc @bensander 
I don't think this is right. We discussed something similar for hcc path a long time back. The consensus was that the compiler should not auto-include any headers. The application should explicitly include hip header files. So TF 1.8 should fix their code to include the header file rather than us changing hipcc to accommodate this.

yxsamliu:
will attempt to add -include to TF crosstool instead of let hipcc do it.

PR #662, opened 2018-09-07
Disable __HIP_ARCH_HAS_DYNAMIC_PARALLEL__ for hip-clang

yxsamliu:


mangupta:
@yxsamliu So are you saying that there are some apps that rely on ```__HIP_ARCH_HAS_DYNAMIC_PARALLEL__``` being defined to use this feature? If so, they are already broken on hcc path as well since dynamic parallelism is not supported there. I fail to see what is different between hcc and hip-clang in the case of dynamic parallelism. Both don't support it. So any HIP application that tries to use this without the proper guards is broken. It is not a regression. And such apps need to fix their code.

include/hip/hcc_detail/hip_runtime.h
```diff
@@ -161,8 +161,7 @@ extern int HIP_TRACE_API;
 // misc
 #define __HIP_ARCH_HAS_SURFACE_FUNCS__ (0)
 #define __HIP_ARCH_HAS_3DGRID__ (1)
-#define __HIP_ARCH_HAS_DYNAMIC_PARALLEL__ (0)
```
mangupta: I don't think removing this is right either. Applications that were checking this flag is the correct way would break. Can you point me to any sample/application that is only checking for ```__HIP_ARCH_HAS_DYNAMIC_PARALLEL__``` being defined instead of checking it's value is 0?

yxsamliu:
It seems this change is no longer needed, so close it now.

PR #669, opened 2018-09-12
Remove potential for mismatch between runtime passed actuals and defined formals

AlexVlx:
This removes a rather populous family of nasty bugs that could arise when an explicit overload was passed alongside arguments that were of a different type which did not match the size / alignment of the formals. The resulting behaviour was rather asinine, with Kernarg becoming botched and nonsense being read. In hindsight, this solution was obvious (as per David Wheeler, a layer of indirection) - too bad it took me a year to realise that. An added benefit is that this will guard against unsupported behaviour (taking references through a `__global__` function), which we silently would have botched before. Finally, there's a strict mode wherein we guard against passing anything that is not `TriviallyCopyable`, which matches the CUDA spec - alas, apps seem to do this gleefully so it's too optimistic to try to enable this by default.

### DIFF
--- a/include/hip/hcc_detail/functional_grid_launch.hpp
+++ b/include/hip/hcc_detail/functional_grid_launch.hpp
@@ -51,30 +51,53 @@ inline T round_up_to_next_multiple_nonnegative(T x, T y) {
     return tmp - tmp % y;
 }
 
-inline std::vector<std::uint8_t> make_kernarg() { return {}; }
-
-inline std::vector<std::uint8_t> make_kernarg(std::vector<std::uint8_t> kernarg) { return kernarg; }
-
-template <typename T>
-inline std::vector<std::uint8_t> make_kernarg(std::vector<uint8_t> kernarg, T x) {
-    kernarg.resize(round_up_to_next_multiple_nonnegative(kernarg.size(), alignof(T)) + sizeof(T));
-
-    new (kernarg.data() + kernarg.size() - sizeof(T)) T{std::move(x)};
-
+template <
+    std::size_t n,
+    typename... Ts,
+    typename std::enable_if<n == sizeof...(Ts)>::type* = nullptr>
+inline std::vector<std::uint8_t> make_kernarg(
+    std::vector<std::uint8_t> kernarg, const std::tuple<Ts...>&) {
     return kernarg;
 }
 
-template <typename T, typename... Ts>
-inline std::vector<std::uint8_t> make_kernarg(std::vector<std::uint8_t> kernarg, T x, Ts... xs) {
-    return make_kernarg(make_kernarg(std::move(kernarg), std::move(x)), std::move(xs)...);
+template <
+    std::size_t n,
+    typename... Ts,
+    typename std::enable_if<n != sizeof...(Ts)>::type* = nullptr>
+inline std::vector<std::uint8_t> make_kernarg(
+    std::vector<std::uint8_t> kernarg, const std::tuple<Ts...>& formals) {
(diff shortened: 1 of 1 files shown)


include/hip/hcc_detail/functional_grid_launch.hpp
```diff
@@ -51,30 +51,53 @@ inline T round_up_to_next_multiple_nonnegative(T x, T y) {
...
+    return make_kernarg<n + 1>(std::move(kernarg), formals);
 }
 
-template <typename... Ts>
-inline std::vector<std::uint8_t> make_kernarg(Ts... xs) {
+template <typename... Formals, typename... Actuals>
+inline std::vector<std::uint8_t> make_kernarg(
+    void (*)(Formals...), std::tuple<Actuals...> actuals) {
+    static_assert(sizeof...(Formals) == sizeof...(Actuals),
+        "The count of formal arguments must match the count of actuals.");
```
scchan: Could you say more about how someone may end up in a situation where they get a size mismatch?  For example, is it like having a kernel that takes a float but the caller passes a double without a type cast?

include/hip/hcc_detail/functional_grid_launch.hpp
```diff
@@ -51,30 +51,53 @@ inline T round_up_to_next_multiple_nonnegative(T x, T y) {
...
+
+    std::tuple<Formals...> to_formals{std::move(actuals)};
```
scchan: Is this move necessary or a way to erase the type of actuals in case of a mismatch?

include/hip/hcc_detail/functional_grid_launch.hpp
AlexVlx: That is a relatively benign example, but yes. There are 2 (simplifying) cases: one passes a `foo(...)` that is unique or one passes a `foo(...)` which is part of a family of functions. For the latter the mechanism in hipLaunchKernelGGL will try do pick the overload based on the actuals, and will error out if there is a mismatch - this is good. However, there are many cases in which one explicitly picks an overload, e.g. via explicit template instantiation, for which the formals are mismatched with the actuals - this is bad as there's no mechanism guarding against it. Since currently the host lays out the Kernarg strictly based on actuals, what one can (and does) end up with is egregious misaligned accesses when you e.g. read the lower 4-bytes of a thing and the upper 4 bytes of the next thing and form a pointer. This has turned out inconvenient to diagnose and debug for users, as it involves taking a peek at the offsets from which we do Kernarg loads.

include/hip/hcc_detail/functional_grid_launch.hpp
AlexVlx: `move` doesn't do type erasure (or much at all really). `actuals` are sink parameters therefore it only makes sense to move from them as opposed to creating spurious copies.

scchan: sorry I mis-read and I thought you added an extra move into to_formals such that make_kernarg would get instantiated with types of formals rather than actuals since they could be mismatched.  This looks fine

AlexVlx: `make_kernarg` takes the tuple of `Actuals` cast to `Formals` as its argument. The cast is part of constructing the `tuple`. Observe it is a `tuple<Formals...>` that is constructed from a pack of rvalue references to `Actuals`.

scchan: I was referring to the subsequent call to make_kernarg which now takes a ref to the tuple of `Formals` so now the construction of the kernarg buffer is consistent with the kernel signature.

PR #672, opened 2018-09-13
Only LLVM6 and higher contain the necessary intrinsics.

iotamudelta:
This is required if we want to move to LLVM5 as a base compiler for PyTorch

### DIFF
--- a/include/hip/hcc_detail/hip_fp16.h
+++ b/include/hip/hcc_detail/hip_fp16.h
@@ -29,7 +29,7 @@ THE SOFTWARE.
     #include <utility>
 #endif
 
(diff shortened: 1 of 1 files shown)


iotamudelta:
@mangupta could you review this? thanks!

iotamudelta:
@mangupta no - the host compiler somewhere includes the HIP headers and chokes on the intrinsics.

mangupta:
So the host compiler is based on clang version below 5?

iotamudelta:
Host compiler is LLVM5 - the intrinsics are only available from LLVM6 on.

PR #678, opened 2018-09-17
Disable device code for gcc in hip_memory.h

yxsamliu:
These device code should only be seen by HCC or hip-clang. They causd build failure
for HIP-VDI runtime and should be disabled for gcc.


yxsamliu:
This will not end up as unresolved symbol. The device functions `__hip_malloc` and `__hip_free` are inline functions defined in this header file. The device code has to be compiled by either HCC or hip-clang, therefore these functions are always defined when compiling for device.

HIP/VDI runtime only contains host code, therefore it does not need these device functions.

PR #685, opened 2018-09-21
Improve hip_trig test case

aaronenyeshi:
Hip_trig is failing but returning a passing value on HIP-Clang.  Also update with debug output and free arrays used.


aaronenyeshi:
Thanks for reviewing, I had it in a header for some reason, just added it

PR #688, opened 2018-09-25
Use sinf and cosf from ocml device libs

aaronenyeshi:
Using llvm_amdgcn builtin fails to produce accurate values, we should move to using the ocml device library versions.

### DIFF
--- a/include/hip/hcc_detail/math_functions.h
+++ b/include/hip/hcc_detail/math_functions.h
@@ -507,7 +507,7 @@ float ynf(int n, float x)
 // BEGIN INTRINSICS
 __DEVICE__
 inline
-float __cosf(float x) { return __llvm_amdgcn_cos_f32(x); }
+float __cosf(float x) { return __ocml_cos_f32(x); }
 __DEVICE__
 inline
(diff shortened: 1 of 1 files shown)


b-sumner:
Just curious how you determined that the results were not accurate enough?   According to E.2 table 9 in the cuda C programming guide the accuracy of __cosf must be

For x in [-π,π], the maximum absolute error is 2-21.19, and larger otherwise.

But are we required to follow this?

aaronenyeshi:
@b-sumner I was using the test hip_trig.cpp , and llvm's implementation may be broken. When using __llvm_amdgcn_sin_f32 and __llvm_amdgcn_cos_f32, the results that didn't make sense: sin of 1 was 0, and cos of 1 was 1. Whereas the right answer should be around 0.841471 and 0.540302.

b-sumner:
Interesting.  Did you have a look at the ISA or could you post the relevant parts here?

b-sumner:
Thanks.  I had forgotten that those intrinsics require the argument to be scaled, so those implementations were quite broken.   __builtin_cosf and __builtin_sinf do the argument scaling; it would be interesting to know if using those instead of ocml causes the test to pass.

PR #692, opened 2018-09-26
HIP program state re-initialization logic (take 2)

whchung:
This PR tries to re-initialize HIP runtime data structures in program_state.cpp. In applications such as TensorFlow it was evident that HIP kernels within shared libraries may not be identified if they are loaded later after initialization by dlopen(). Instead of raising an exception immediately we re-initialize data structures within program_state.cpp with a rebuild flag.

In ROCm 1.9 the behavior of HIP runtime was changed so GPU kernel discovery is carried out upon the 1st GPU kernel invocation. That helped quite a few TensorFlow unit tests as well as PaddlePaddle tests. But it turns out there are additional TensorFlow-based applications and libraries (example: RCCL) which would load additional shared libraries *AFTER* TensorFlow is initialized, and thus logic in this PR is required to make sure GPU kernels in these shared libraries are properly discovered.

### DIFF
--- a/include/hip/hcc_detail/program_state.hpp
+++ b/include/hip/hcc_detail/program_state.hpp
@@ -93,11 +93,12 @@ public:
     }
 };
 
-const std::unordered_map<hsa_agent_t, std::vector<hsa_executable_t>>& executables();
+const std::unordered_map<hsa_agent_t, std::vector<hsa_executable_t>>& executables(
+    bool rebuild = false);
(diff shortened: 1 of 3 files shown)


mangupta:
@sunway513 thanks for confirming that it needs to go to ROCm 2.0. Also do you still want to keep the startup loader disabled by default? Or does that need to change as well?

sunway513:
Hi @mangupta , please keep the current startup loader setting, Paddle still need it to be disabled by default.

PR #696, opened 2018-10-01
Fixed image width for linear resource type texture

### LINKED ISSUE #671: hipResourceDesc.res.linear.sizeInBytes is passed to HSA incorreclty
https://github.com/ROCm-Developer-Tools/HIP/blob/master/src/hip_texture.cpp#L266

According to HSA specification, width is "Width of the image, in components", not in bytes:
http://www.hsafoundation.com/html/Content/Runtime/Topics/03_Extensions/hsa_ext_image_descriptor_t.htm

gargrahul:
Fixes #671

### DIFF
--- a/src/hip_texture.cpp
+++ b/src/hip_texture.cpp
@@ -198,7 +198,6 @@ bool getHipTextureObject(hipTextureObject_t* pTexObject, hsa_ext_image_t& image,
     return true;
 }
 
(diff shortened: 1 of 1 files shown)


ex-rzr:
I'm not sure but it seems that your solution ignores multiple channels.

It should be something like this (it needs checking though):
```
width = sizeInBytes / ((desc.x + desc.y + desc.z + desc.w) / 8)
```

gargrahul:
@ex-rzr Thanks. You are right. I will correct it.

PR #701, opened 2018-10-03
introduce tests for asynchronous streams

ssahasra:
The test is currently disabled, until the device implementation is
actually checked in. This only tests the synchronous aspects of the
stream.

### DIFF
--- a/tests/src/deviceLib/hipAsynchronousStreams.cpp
+++ b/tests/src/deviceLib/hipAsynchronousStreams.cpp
@@ -1,0 +1,713 @@
+/*
+Copyright (c) 2018 Advanced Micro Devices, Inc. All rights reserved.
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+/*
+ * HIT_START
+ * BUILD: %t %s ../test_common.cpp HIPCC_OPTIONS -std=c++11
+ * RUN: %t
+ * HIT_END
+ */
+
+// Includes HIP Runtime
+#include "hip/hip_runtime.h"
+#include <test_common.h>
+
+#include <set>
+#include <unordered_map>
+
+#define test_failed(test_name)                                                 \
+    printf("%s %s  FAILED!%s\n", KRED, test_name, KNRM);
+#define test_passed(test_name)                                                 \
+    printf("%s %s  PASSED!%s\n", KGRN, test_name, KNRM);
+
+/* BEGIN DUPLICATION */
+
+// TODO: The host code below is duplicated, and should ideally reside
+// in a common host-side library.
+
+typedef enum {
+    __OCKL_AS_PACKET_EMPTY = 0,
+    __OCKL_AS_PACKET_READY = 1
+} __ockl_as_packet_type_t;
+
+#define __OCKL_AS_PAYLOAD_ALIGNMENT 4
+#define __OCKL_AS_PAYLOAD_BYTES 48
+
+typedef enum {
+    __OCKL_AS_PACKET_HEADER_TYPE = 0, // corresponds to HSA_PACKET_HEADER_TYPE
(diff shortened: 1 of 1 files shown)


ssahasra:
@yxsamliu, @aaronenyeshi: I am unable to tag reviewers, probably because I am not a member of the project.
FYI: @b-sumner

tests/src/deviceLib/hipAsynchronousStreams.cpp
```diff
@@ -0,0 +1,607 @@
...
+    }
+
+    HIPCHECK(hipHostFree(buffer));
+    return success;
+}
+
+__global__ void multipleProducers(__ockl_as_stream_t* stream) {
+    const unsigned int tid = blockDim.x * blockIdx.x + threadIdx.x;
+    unsigned char data = (unsigned char)tid;
+    unsigned char len = 1;
```
b-sumner: Unused variable?

tests/src/deviceLib/hipAsynchronousStreams.cpp
```diff
@@ -0,0 +1,607 @@
...
+    }
+
+    HIPCHECK(hipHostFree(buffer));
+    return success;
+}
+
+__global__ void dropPackets(__ockl_as_stream_t* stream, unsigned int* status) {
+    const unsigned int tid = blockDim.x * blockIdx.x + threadIdx.x;
+    unsigned char data = (unsigned char)tid;
+    unsigned char len = 1;
```
b-sumner: Unused variable?

tests/src/deviceLib/hipAsynchronousStreams.cpp
```diff
@@ -0,0 +1,607 @@
...
+
+        if (*packet->payload >= 3) return false;
+    }
+
+    return true;
+}
+
+bool ocklAsDropPackets() {
+    bool success = true;
+    unsigned int numThreads = 4;
```
b-sumner: Again, these are too small.

tests/src/deviceLib/hipAsynchronousStreams.cpp
```diff
@@ -0,0 +1,607 @@
...
+
+        if (flags & __OCKL_AS_CONNECTION_END) break;
+    }
+
+    if (buffer.compare(STR27 STR64 STR40) != 0) return false;
+
+    return true;
+}
+
+bool ocklAsSplitMessage() {
```
b-sumner: Need a multiple split messages test with a non-tiny number of threads, and blocks.

ssahasra:
I've pushed another commit to address Brian's comments. The tests now use 3 blocks of 123 threads each (just to avoid powers of two). Also added extra strings where applicable.

ssahasra:
> I've pushed another commit to address Brian's comments. The tests now use 3 blocks of 123 threads each (just to avoid powers of two). Also added extra strings where applicable.

Ugh. Forgot to run clang-format. I am not sure how well github handles back-to-back commits in a review. Will ensure proper formatting before the change is merged.

b-sumner:
This set is a good start, but all have the form launch, synchronize, check stream.  But I think we also need, in cases where it is supported, tests where producers and consumer are running concurrently.

ssahasra:
Uploaded a fresh new change. All the tests are for synchronous mode only, and they all pass with the latest version of the device library implementation.

ssahasra:
The HIP test for asynchronous streams is now updated to reflect the changes in the HCC tests. This single test runs kernels that are equivalent to the HCC tests under tests/Unit/OCKL/AsyncStreams (in HCC sources), but only with synchronous mode.

The test should now run correctly since the async stream library is already part of rocdl. Need a review in order to push this change in.

tests/src/deviceLib/hipAsynchronousStreams.cpp
```diff
@@ -0,0 +1,707 @@
...
+
+__global__ void
+singlePacketSingleProducer(__ockl_as_stream_t *stream)
+{
+    uint len = STRLEN_HELLO_WORLD;
+
+    uint64_t connection_id;
+
+    __hip_as_write_block(stream, 42, &connection_id,
+                         (const uint8_t *)STR_HELLO_WORLD, 11,
```
mangupta: Probably need ```len``` here instead of ```11```.

tests/src/deviceLib/hipAsynchronousStreams.cpp
```diff
@@ -0,0 +1,707 @@
...
+}
+
+__global__ void
+singlePacketSingleProducer(__ockl_as_stream_t *stream)
+{
+    uint len = STRLEN_HELLO_WORLD;
+
+    uint64_t connection_id;
+
+    __hip_as_write_block(stream, 42, &connection_id,
```
mangupta: 42 maybe the answer to the Life, Universe and Everything, but not sure what it is doing here. Maybe you wanted ```__OCKL_AS_BUILTIN_SERVICE_PRINTF```?

tests/src/deviceLib/hipAsynchronousStreams.cpp
```diff
@@ -0,0 +1,707 @@
...
+
+__global__ void
+dropPackets(__ockl_as_stream_t *stream, unsigned int *status)
+{
+    const unsigned int tid = blockDim.x * blockIdx.x + threadIdx.x;
+
+    uint64_t connection_id;
+
+    status[tid] =
+        __hip_as_write_block(stream, 42, &connection_id,
```
mangupta: ```__OCKL_AS_BUILTIN_SERVICE_PRINTF``` again?

tests/src/deviceLib/hipAsynchronousStreams.cpp
ssahasra: Fixed. Thanks!

tests/src/deviceLib/hipAsynchronousStreams.cpp
ssahasra: The PRINTF enum is not required here. I have removed it entirely now. Also replaced the magic number 42 with a const integer TEST_SERVICE. It just needs to be some non-zero number for the test to check successful transmission.

tests/src/deviceLib/hipAsynchronousStreams.cpp
ssahasra: Fixed as mentioned earlier.

PR #705, opened 2018-10-10
Feature minimal changes for hc next

AlexVlx:
We have a bunch of changes coming up in HC next, including but not restricted to removing the decrepit grid_launch headers. This introduces the changes needed to keep HIP running smoothly.


tests/src/deviceLib/hipTestFMA.cpp
```diff
@@ -61,7 +62,7 @@ THE SOFTWARE.
     Check(fma(d, (double)f, (unsigned char)3) == doubleResult);
     Check(fma(d, (double)f, (short)3) == doubleResult);
     Check(fma(d, (double)f, (unsigned short)3) == doubleResult);
-    Check(fma(d, (double)f, (int)3) == doubleResult);
+    Check(fma(d, (double)f, (int)3) == doubleResult);
```
aaronenyeshi: Probably an accidental whitespace here

include/hip/hcc_detail/hip_runtime.h
```diff
@@ -109,9 +109,6 @@ extern int HIP_TRACE_API;
 #include <hip/hcc_detail/device_functions.h>
 #include <hip/hcc_detail/surface_functions.h>
 #include <hip/hcc_detail/texture_functions.h>
-#if __HCC__
-#include <hip/hcc_detail/math_functions.h>
```
aaronenyeshi: Does this mean that all HIP applications that use these math functions will also need to update and add `#include <hip/math_functions.h>`?

include/hip/hcc_detail/grid_launch.h
```diff
@@ -0,0 +1,69 @@
...
+
+
+// grid_launch_parm contains information used to launch the kernel.
+typedef struct grid_launch_parm
+{
+  //! Grid dimensions
+  gl_dim3      grid_dim;
+
+  //! Group dimensions
+  gl_dim3      group_dim;;
```
aaronenyeshi: extra semicolon

include/hip/hcc_detail/grid_launch.h
```diff
@@ -0,0 +1,69 @@
...
+
+  //! Amount of dynamic group memory to use with the kernel launch.
+  //! This memory is in addition to the amount used statically in the kernel.
+  unsigned int  dynamic_group_mem_bytes;;
```
aaronenyeshi: extra semicolon

include/hip/hcc_detail/grid_launch.h
```diff
@@ -0,0 +1,69 @@
...
+  //! see HSA_PACKET_HEADER_ACQUIRE_FENCE_SCOPE and hsa_fence_scope_t.
+  //! Set to -1 for conservative defaults.
+  //! Placeholder, is not used to control packet dispatch yet
+  unsigned int  launch_fence;
+
+  //! Pointer to the accelerator_view where the kernel should execute.
+  //! If NULL, the default view on the default accelerator is used.
+  hc::accelerator_view  *av;
+
+  //! Pointe to the completion_future used to track the status of the command.
```
aaronenyeshi: Small typo: `Pointer`

include/hip/hcc_detail/grid_launch.h
AlexVlx: The grid_launch files were just copied over from HCC, where we'd like to purge them from, so this thing has been there for years now... Good thing you spotted it:)

include/hip/hcc_detail/grid_launch.h
AlexVlx: Thanks.

include/hip/hcc_detail/grid_launch.h
AlexVlx: Thanks, see above.

include/hip/hcc_detail/hip_runtime.h
AlexVlx: Err no, that's a typo, thank you for spotting it.

david-salinas:
FYI, HCC upstream merge, builds/tests HIP, needs this PR with latest HCC (which has amp.h removed).

samples/0_Intro/hcc_dialects/vadd_hc_am.cpp
```diff
@@ -26,8 +26,8 @@ THE SOFTWARE.
 // which can only be used on the device.  The programmer has full control
 // over when data is copied.
 
-#include <hc.hpp>
-#include <hc_am.hpp>
+#include <hc/hc.hpp>
```
mangupta: Issue with this change is that if HCC is installed using ```make install```, there is no hc folder under hcc/include. So unless the only supported way of using HCC is to install using deb/rpm packages, this change is not acceptable. Alternatively HCC needs to fix ```make install``` to create hcc/include/hc and place all the include files under it.

tests/src/deviceLib/hip_trig.cpp
```diff
@@ -28,10 +28,11 @@ THE SOFTWARE.
 
 #include <hip/hip_runtime_api.h>
 #include <hip/hip_runtime.h>
-#include <iostream>
 #include "test_common.h"
 #include <hip/device_functions.h>
 
+#include <iostream>
```
mangupta: Why does the order of the includes need to be changed?

tests/src/kernel/hipLanguageExtensions.cpp
```diff
@@ -28,10 +28,11 @@ THE SOFTWARE.
  */
 
 #include "hip/hip_runtime.h"
+
 #include <test_common.h>
 
 #ifdef __HCC__
-#include <amp.h>
+    #include <hc.hpp>
```
mangupta: This change has already been submitted and merged since HCC broke this HIP test. So we can discard this change from this PR.

tests/src/kernel/hipLanguageExtensions.cpp
```diff
@@ -53,7 +54,7 @@ __device__ __forceinline__ int sum1_forceinline(int a) { return a + 1; };
 
 __device__ __host__ float PlusOne(float x) { return x + 1.0; }
 
-__global__ void MyKernel(const hipLaunchParm lp, const float* a, const float* b, float* c,
```
mangupta: This change also was submitted and merged as part of the PR that replaces hipLaunchKernel with hipLaunchKernelGGL in all tests and samples.

mangupta:
Also does this PR need to go into ROCm 2.0?

samples/0_Intro/hcc_dialects/vadd_hc_am.cpp
AlexVlx: Of course it is acceptable m8, it's just that these silly interop examples will be temporarily disabled or ignored. Nobody uses this anyway. This has nothing to do with debs (that creates different path anyway), but rather pertains to how the HC headers will be exposed going forward, i.e. they will be in hc/, with the legacy non-prefixed stuff existing only for HIP's perusal.

AlexVlx:
@mangupta yes it does need to go into 2.0

samples/0_Intro/hcc_dialects/vadd_hc_am.cpp
mangupta: Okay that is acceptable.

AlexVlx: I've refactored this to have the new includes under a guard. I do believe we may want to eventually remove the samples, but they'll stay usable during the transition.

mangupta:
@AlexVlx CI test failures seems to indicate that there is some conflict w.r.t. math functions between HIP's ```include/hip/hcc_detail/math_functions.h``` and HCC's ```include/kalmar_math.h```.

PR #714, opened 2018-10-19
Remove libhip_device.a static library

aaronenyeshi:
Move remaining function definitions from device_util.cpp to hip_runtime.h header. We can now remove the static library completely as device_util.cpp was the last file part of target hip::hip_device .

In this PR, I'm removing the device_util.cpp and associated links. Removing the hip:hip_device target, and also the package creation of libhip_device.a static lib.


aaronenyeshi:
Review comments from Maneesh (having connection issues):
CMakeLists.txt:
•	Interface library device is an alias for libhip_device.a. So we need to remove any reference to library device as well.

packaging/hip-targets.cmake:
•	Target hip::device is an alias for hip::hip_device. So we need to remove any reference to target hip::device as well.

tests/src/experimental/xcompile/*.sh:
•	These tests are outdated and are already broken. So removing reference to device_util.cpp from the build scripts does not have any meaningful impact. These tests need to be rewritten and are beyond the scope of this PR. So you might as well leave them unchanged.

aaronenyeshi:
I've removed the tests/src/experimental/xcompile/*.sh files AS-IS. I'm working on removing device alias now.

aaronenyeshi:
@mangupta I've removed hip::device completely, and we will replace its functionality using the hip::host target instead.

pfultz2:
hip::device should not be removed, otherwise we have no way to use hip for host-only or enable device compiling separately.

CMakeLists.txt
```diff
@@ -222,19 +219,16 @@ if(HIP_PLATFORM STREQUAL "hcc")
...
         target_include_directories(${TARGET} SYSTEM INTERFACE $<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>;${HSA_PATH}/include)
     endforeach()
     add_library(host INTERFACE)
-    target_link_libraries(host INTERFACE hip_hcc)
-    add_library(device INTERFACE)
     if(HIP_COMPILER STREQUAL "hcc")
-        target_link_libraries(device INTERFACE host hip_device hcc::hccrt hcc::hc_am)
-    elseif(HIP_COMPILER STREQUAL "clang")
-        target_link_libraries(device INTERFACE host hip_device)
+        target_link_libraries(host INTERFACE hip_hcc hcc::hccrt hcc::hc_am)
```
pfultz2: `hcc::hccrt` and `hcc::hc_am` require hcc compiler, but `hip::host` should work when compiling with gcc as well.

aaronenyeshi:
As discussed in the email, we will not remove hip::device target for now. We will keep hip::host as it was before. @pfultz2 @scchan @mangupta

aaronenyeshi:
@mangupta Maybe we can remove hip::device in a future PR after discussion. For now, this is good enough for 1.9 backport

PR #717, opened 2018-10-22
Check the returned hipDeviceptr_t from read_global_description()

fronteer:
What to check should be the content of dptr rather dptr itself.


mangupta:
Closing this since the change has been resubmitted as #721 which targets master branch.

PR #720, opened 2018-10-23
Make HIP functional again with HCC from ROCm 1.9.x

mangupta:


src/hip_stream.cpp
```diff
@@ -31,12 +31,21 @@ THE SOFTWARE.
 //-------------------------------------------------------------------------------------------------
 // Stream
 //
+#if (__hcc_minor__ < 3)
```
yxsamliu: HIP/VDI is compiled with gcc. This condition will always be true.

Should it be #if __HCC__ && (__hcc_minor__ < 3) ?

mangupta:
@yxsamliu I have updated the code. Hopefully this should address HIP/VDI concerns.

PR #723, opened 2018-10-24
Fix logic for double variants of __shfl*

mangupta:


b-sumner:
Are there any tests for these functions?  If so they should be improved.

b-sumner:
Otherwise the changes look fine.

mangupta:
I cross checked. shfl seems to have managed to evade all our the device library tests. So I need to create a new test for this. Will do that change as another PR. So going ahead and merging this in.

PR #726, opened 2018-10-25
hipLaunchKernel, hipLaunchParm are deprecated, and shall be removed.

AlexVlx:
Let's mark these as deprecated for now, so that users have time to adapt. I've just stumbled across internal examples being written in the present time and still using `hipLaunchParm`, which is outright ghastly, so the only way to get rid of these is to completely remove them in 2.x.


AlexVlx:
@bensander no, hence why the only function that is going to be removed is only marked deprecated for now, and the text says "completely remove them in 2.x".

PR #730, opened 2018-10-26
Add more checks and debug outputs for fat binary

yxsamliu:


b-sumner:
Should we document the new environment variable?

PR #731, opened 2018-10-28
Rely on code object metadata for kernarg arguments alignof and sizeof.

AlexVlx:
Due to assumptions in Clang about target independence it is possible to end up with cases in which for some type T the alignment and size obtained via `alignof` and `sizeof` are not what LC sees when it lowers to native ISA (e.g. implicit packing is opaque). This addresses such issues by way of using the code object metadata, which accurately reflects the quantities used by LC. The metadata parser itself is not ideal and should be seen as a placeholder, which the experts in the HIP team can either overhaul in the future, if they see fit, or completely replace with equivalent functionality in code object manager v3, once that is adopted. It is desirable to have this in ROCm 2.0.


AlexVlx:
@scchan if you want to implement that, sure, it would be glorious. Unfortunately, I'm not up for that task at the moment, even though it feels like a very good follow-up.

AlexVlx:
@mangupta can we accelerate reviewing & merging this please? I fear that on this occasion the usual speed will not do, given scheduling constraints. Thanks.

mangupta:
@AlexVlx Some of the tests are failing. See the snippet below-

```
The following tests FAILED:
         84 - directed_tests/deviceLib/hipTestIncludeMath.tst (OTHER_FAULT)
        104 - directed_tests/deviceLib/hipFloatMathPrecise.tst (OTHER_FAULT)
        109 - directed_tests/kernel/hipPrintfKernel.tst (OTHER_FAULT)
        116 - directed_tests/kernel/hipTestGlobalVariable.tst (OTHER_FAULT)
```
Manually running one of them shows-
```
./directed_tests/deviceLib/hipTestIncludeMath
terminate called after throwing an instance of 'std::runtime_error'
  what():  Missing metadata for __global__ function: _Z16FloatMathPrecisev
Aborted (core dumped)
```
Either the code changes are incomplete or maybe the tests need modifications to work with these changes.

PR #748, opened 2018-11-04
Implement hipGetSymbolAddress and hipGetSymbolSize

mkuron:


mkuron:
@gargrahul, thanks for the comments. I have added a test case and introduced an `ihipModuleGetGlobal`. `hipMemcpy`(`To`|`From`)`Symbol`(|`Async`) were also calling `hipModuleGetGlobal`, which I have changed to the new function.

mkuron:
@gargrahul, is there anything else you need me to do before this can be meeged?

gargrahul:
Jenkins is failing due to build issue on NVCC path. I will try to fix that.

Otherwise LGTM, though I would like to have another test to use memcpy to/from symbol after getting symbol address. I have already tried that and can be added as a new PR.

gargrahul:
@mangupta I think we can merge this unless you have any review comment.

PR #750, opened 2018-11-05
Remove non-default-rounded math apis

aaronenyeshi:
In ROCm-Device-Libs, they have dropped the non-default-rounded versions of add, sub, mul, div, sqrt and fma. Therefore, ocml has removed the rte, rtp, rtn, and rtz counterparts. This will remove the same math APIs in HIP for _ru, _rd, _rn, and _rz.

Since ROCDL removes these ocml functions here: https://github.com/RadeonOpenCompute/ROCm-Device-Libs/commit/2fc04e10e1354edee331ce700f98a60f8255effb

I've also updated the doc to set these as unsupported.


aaronenyeshi:
I've also fixed some of the HIP math API references, with OCML matching to HIP as:
- rte - round to nearest even mode = _rn - round nearest even
- rtn - round towards negative infinity = _rd - round down
- rtp - round towards positive infinity = _ru - round up
- rtz - round towards zero = _rz - round zero

include/hip/hcc_detail/math_functions.h
```diff
@@ -514,69 +514,69 @@ float __exp10f(float x) { return __ocml_exp10_f32(x); }
...
-__DEVICE__
-inline
-float __fdiv_rn(float x, float y) { return x / y; }
-__DEVICE__
-inline
-float __fdiv_ru(float x, float y) { return x / y; }
-__DEVICE__
-inline
-float __fdiv_rz(float x, float y) { return x / y; }
+// __DEVICE__
```
b-sumner: Would it be better to use a guard here like #if defined OCML_BASIC_ROUNDED_OPERATIONS instead of comments?

aaronenyeshi: Okay I can use guards instead, its more flexible

tests/src/deviceLib/hipDoublePrecisionIntrinsics.cpp
```diff
@@ -34,34 +34,38 @@ THE SOFTWARE.
...
+    __ddiv_rd(0.0, 1.0);
+    __ddiv_rn(0.0, 1.0);
+    __ddiv_ru(0.0, 1.0);
+    __ddiv_rz(0.0, 1.0);
+    __dmul_rd(1.0, 2.0);
+    __dmul_rn(1.0, 2.0);
+    __dmul_ru(1.0, 2.0);
+    __dmul_rz(1.0, 2.0);
+#endif
     __drcp_rd(2.0);
```
b-sumner: Why is drcp not hidden?  I don't think there is a correct implementation

include/hip/hcc_detail/math_functions.h
```diff
@@ -1034,42 +1040,44 @@ double yn(int n, double x)
...
+inline
+double __dmul_rn(double x, double y) { return __ocml_mul_rte_f64(x, y); }
+__DEVICE__
+inline
+double __dmul_ru(double x, double y) { return __ocml_mul_rtp_f64(x, y); }
+__DEVICE__
+inline
+double __dmul_rz(double x, double y) { return __ocml_mul_rtz_f64(x, y); }
+#endif
 __DEVICE__
```
b-sumner: These rcp implementations are simply not correct..  They should be guarded too, and can use the same guard since they will use OCML for their implementation.

tests/src/deviceLib/hipDoublePrecisionIntrinsics.cpp
aaronenyeshi: Hi Brian, where can I find the ocml functions for rcp?

include/hip/hcc_detail/math_functions.h
aaronenyeshi: I'll guard the rcp as well, but I won't change which ocml function it calls since it doesn't seem like ROCDL has __ocml_rcp_rt[n/p/z/e]_f[32/64] functions yet.

tests/src/deviceLib/hipDoublePrecisionIntrinsics.cpp
b-sumner: There isn't one.  A standard reciprocal is simply 1 / x.  There is no way to get a non-default-rounded reciprocal at all (just like there is no way to get a non-default-rounded add).

b-sumner: I.e., when OCML implements non-default-rounded division functions, then these reciprocal functions should be implemented using it.

b-sumner:
Looks good to me.

PR #760, opened 2018-11-13
Roctracer hip frontend 181113

eshcherb:
Rebase of PR#645


mangupta:
@eshcherb does this replace #645?

eshcherb:
Yes, this replace #645

CMakeLists.txt
```diff
@@ -150,6 +150,19 @@ if(NOT DEFINED COMPILE_HIP_ATP_MARKER)
 endif()
 add_to_config(_buildInfo COMPILE_HIP_ATP_MARKER)
 
+################
+# Detect profiling API
+################
+if (USE_PROF_API EQUAL 1)
+add_definitions(-DUSE_PROF_API=1)
```
mangupta: Didn't notice this before. But we should add the definition only if we find roctracer. Otherwise if we tried enabling USE_PROF_API on ROCm 1.9.x, it would fail. So this line should be moved to just before ```include_directories ( ${PROF_API_HEADER_DIR} )```.

PR #761, opened 2018-11-14
Let hip-clang support --genco

yxsamliu:


src/hip_fatbin.cpp
```diff
@@ -0,0 +1,89 @@
...
+    if (HIP_DUMP_CODE_OBJECT)
+      __hipDumpCodeObject(std::string{codeobj, desc->size});
+
+    tprintf(DB_FB, "__hipExtractCodeObjectFromFatBinary succeeds and returns %p\n",
+        codeobj);
+    return codeobj;
+  }
+
+  fprintf(stderr, "No device code bundle for %s\n", agent_name);
+  abort();
```
mangupta: Why do we need an abort here? We are calling ```__hipExtractCodeObjectFromFatBinary``` unconditionally for both hip-clang as well as hcc. When using hcc, obviously there is no device code bundle that can be found. So we end up aborting instead of continuing.

src/hip_module.cpp
```diff
@@ -546,6 +547,12 @@ hipError_t ihipModuleLoadData(hipModule_t* module, const void* image) {
     auto ctx = ihipGetTlsDefaultCtx();
     if (!ctx) return hipErrorInvalidContext;
 
+    // try extracting code object from image as fatbin.
+    char name[64] = {};
+    hsa_agent_get_info(this_agent(), HSA_AGENT_INFO_NAME, name);
+    if (auto *code_obj = __hipExtractCodeObjectFromFatBinary(image, name))
```
mangupta: Alternatively if we still need the above abort, we can call this piece of code only when hip-clang is used as the compiler.

src/hip_fatbin.cpp
yxsamliu: when using hcc and loading a code object, this function will return at line 56, since a code object is not a fat binary and does not contain the magic number of fat binary.
If the control flow reaches this point, that means the app is trying to load an invalid fat binary, therefore abort.

src/hip_module.cpp
yxsamliu: As commented above, apps compiled by hcc will try to load a code object, therefore will not abort. Abort only happens when apps compiled by hip-clang tries to load a corrupted fat binary.

src/hip_fatbin.cpp
mangupta: The code object for HCC is also a fat binary. The generation is handled in https://github.com/ROCm-Developer-Tools/HIP/blob/master/lpl_ca/lpl.hpp#L94-L117.

When using HCC if you run ```make vcpy_kernel.code``` in https://github.com/ROCm-Developer-Tools/HIP/tree/master/samples/0_Intro/module_api and look at the contents of the generated code object, you can see that it starts with ```__CLANG_OFFLOAD_BUNDLE__```. However, even though both hip-clang and hcc generate a fat binary, the format is slightly different. Hence when using this PR with HCC, we see ```"No device code bundle for gfx<NNN>``` followed by an abort.

yxsamliu: updated the PR. Removed abort().

src/hip_module.cpp
yxsamliu: The host code could be compiled by gcc, therefore there is no way to know the fat bin is intended for HCC or hip-clang. I have removed abort from code above so that fat bin for HCC can be handled in other places.

PR #762, opened 2018-11-15
[HIPIFY][LLVMCompat] Support of upcoming LLVM 8.0

emankov:
+ StringRef issue, which is finally moved to LLVM from Clang
+ Renamed getBeginLoc() and getEndLoc() resolution for Expr and TypeLoc classes
+ Support all the previous LLVM versions via LLVCompat


holyprince:
Sorry, I want to ask a question that my hipcc using clang version 7.0 and LLVM is 7.0. Can I just compile LLVM 8.0 and set the hipcc using the new LLVM? Because I do not have sudo to update the rocm and hipcc version.

emankov:
This PR is for hipify-clang, not for hipcc.

PR #797, opened 2018-12-07
Fixed hipPointerGetAttributes for hostmalloced ptr

gargrahul:
This was caused due to #601


ezyang:
Shouldn't there be a test for this? :)

PR #800, opened 2018-12-11
[HIPIFY][tests][fix] Add correct CUDA version detection while testing

emankov:
+ Call load_config for lit.site.cfg earlier in lit.cfg to get initialized cuda_version
+ Exclude CUDA 9.0 tests if CUDA < 9.0
+ Report CUDA version before running tests

### DIFF
--- a/hipify-clang/CMakeLists.txt
+++ b/hipify-clang/CMakeLists.txt
@@ -111,12 +111,6 @@ if (HIPIFY_CLANG_TESTS)
         @ONLY
     )
 
-    configure_file(
-        ${CMAKE_CURRENT_LIST_DIR}/../tests/hipify-clang/lit.cfg.in
-        ${CMAKE_CURRENT_LIST_DIR}/../tests/hipify-clang/lit.cfg
-        @ONLY
-    )
-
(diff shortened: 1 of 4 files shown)


hipify-clang/CMakeLists.txt
```diff
@@ -111,6 +111,12 @@ if (HIPIFY_CLANG_TESTS)
         @ONLY
     )
 
+    configure_file(
+        ${CMAKE_CURRENT_LIST_DIR}/../tests/hipify-clang/lit.cfg.in
+        ${CMAKE_CURRENT_LIST_DIR}/../tests/hipify-clang/lit.cfg
```
dfukalov:
```suggestion
        ${CMAKE_CURRENT_BINARY_DIR}/../tests/hipify-clang/lit.cfg
```
you should not create parsed in source directory tree

emankov: Agree. The request is updated. I've just moved load_config call for lit.site.cfg a bit earlier in lit.cfg to get initialized cuda_version and cuda_version_major. So there is no any need in lit.cfg.in (removed).

PR #810, opened 2018-12-13
Use hsa specialization for a fast path to memset.

iotamudelta:
This improves ROCm performance on an important customer benchmark by 20%.

Developed with @AlexVlx .

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -1509,6 +1509,13 @@ __global__ void hip_copy2d_n(T* dst, const T* src, size_t width, size_t height, 
 
 template <typename T>
 void ihipMemsetKernel(hipStream_t stream, T* ptr, T val, size_t sizeBytes) {
+
+    if(sizeof(T) == sizeof(uint32_t)){
+        // for all sizeof(uint32_t) use the fast path using hsa specialization
+        hsa_amd_memory_fill(ptr, reinterpret_cast<const std::uint32_t&>(val), sizeBytes);
+        return;
+    }
+
     static constexpr uint32_t block_dim = 256;
 
     const uint32_t grid_dim = clamp_integer<size_t>(sizeBytes / block_dim, 1, UINT32_MAX);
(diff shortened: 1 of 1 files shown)


iotamudelta:
@mangupta the test failures seem unrelated to this PR?

src/hip_memory.cpp
```diff
@@ -1509,6 +1509,13 @@ __global__ void hip_copy2d_n(T* dst, const T* src, size_t width, size_t height,
 
 template <typename T>
 void ihipMemsetKernel(hipStream_t stream, T* ptr, T val, size_t sizeBytes) {
+
+    if(sizeof(T) == sizeof(uint32_t)){
+        // for all sizeof(uint32_t) use the fast path using hsa specialization
+        hsa_amd_memory_fill(ptr, reinterpret_cast<const std::uint32_t&>(val), sizeBytes);
```
mangupta: Failure is in one of the memset tests. I need to check the hsa spec, but I think the ```sizeBytes``` probably needs to be divided by ```sizeof(uint32_t)```.

iotamudelta: According to https://github.com/RadeonOpenCompute/ROCR-Runtime/blob/d823418d0500b30af775efb3b17b28e164958f3a/src/inc/hsa_ext_amd.h#L1282 you are right! Fixing.

iotamudelta: @mangupta looks like that's not sufficient to pass tests - mind having a look? Thanks!

mangupta:
@iotamudelta The failures are still in all the memset related tests. Do all the HIP unit tests pass on your development system?
While I haven't tried digging into why the failures are occurring, there are few possible causes:
* Maybe you are hitting into either of the 2 failure cases @ https://github.com/RadeonOpenCompute/ROCR-Runtime/blob/master/src/inc/hsa_ext_amd.h#L1275-L1279.
* The call to hsa_amd_memory_fill ignores the stream i.e. the hsa_queue (which is important for the async memset cases). While it is doubtful that this can cause an issue, maybe it is worth checking.

PR #811, opened 2018-12-13
Add logic to hipconfig to read from rocm_agent_enumerator

sunway513:
This PR adds logic in hipconfig, to set HIP_PLATFORM as hcc if rocm_agent_enumerator returns additional ROCm agents other than the default gfx000.


mangupta:
closing this since this logic is going to be shortly rewritten.

PR #813, opened 2018-12-14
 Add HIP runtime compilation API

lmoriche:


gargrahul:
@lmoriche There are too many commits in this PR for this change. Where are the implementation of these APIs?

PR #814, opened 2018-12-14
Add HIP runtime compilation API

lmoriche:


mangupta:
We also have hiprtc implementation inside HIP source which should work on both hcc and hip-clang. So do we still need this?

mangupta:
Closing this since there is no response from the author and I think this is no longer required.

Issue #812, opened 2018-12-14
[HIPIFY] Place for temporary files


kasaurov:
Currently hipify tries to create temporary files in the source folder. It goes ugly if someone tries to hipify an example from read-only folder. It is more reasonable to create temporary files in whether system temporary folder or have a way to set the folder for them.

error: error reading '/usr/local/cuda/NVIDIA_CUDA-9.2_Samples/0_Simple/matrixMul/matrixMul.cu.hipify-tmp'
1 error generated when compiling for host.

PR #816, opened 2018-12-17
[HIPIFY] Temporary files in system temps introducing

emankov:
+ Files are not being placed near original source before hipification anymore, system tmp is used for them now.
+ Additional include (-I) to absolute original source dir is added
+ Partially implements #812
+ Tested on windows/linux, single/multiple inputs, with and without includes to user header files


emankov:
Ping

PR #818, opened 2018-12-18
Add default arguments for hipConfigureCall

yxsamliu:


include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -2573,7 +2573,7 @@ hipError_t hipIpcCloseMemHandle(void* devPtr);
  * @returns hipSuccess, hipInvalidDevice, hipErrorNotInitialized, hipErrorInvalidValue
  *
  */
-hipError_t hipConfigureCall(dim3 gridDim, dim3 blockDim, size_t sharedMem, hipStream_t stream);
+hipError_t hipConfigureCall(dim3 gridDim, dim3 blockDim, size_t sharedMem = 0, hipStream_t stream = 0);
```
mangupta: Default arguments work only on C++. To allow HIP runtime header to work with C compilers, we need to wrap the default value inside ```__dparm()```. So please replace the above with ```size_t sharedMem __dparm(0), hipStream_t stream __dparm(0)```

yxsamliu: Fixed. Thanks.

PR #819, opened 2018-12-19
Improvements to hipify-perl and HIP porting guide for HIP_SYMBOL

stanleytsang-amd:
Currently the the hipify-perl tool does not correctly hipify cudaMemcpyToSymbol, cudaMemcpyFromSymbol, cudaGetSymbolAddress, and cudaGetSymbolSize, as it does not insert the HIP_SYMBOL macro required around the symbol name for the HIP equivalents of these functions.  I've updated the appropriate regexes to generate correct HIP code, accounting for whitespaces and newlines.  Additionally I've updated the HIP porting guide to make stronger mention of the HIP_SYMBOL macro, as currently the only place I can find online where the HIP_SYMBOL macro is even mentioned is in the code snippet for hipMemcpyToSymbol in the porting guide, which is quite easy to overlook.


gargrahul:
@stanleytsang-amd , commit 418af22 doesn't look correct for changes needed in this PR.

Issue #824, opened 2018-12-20
[HIPIFY] Option -inplace stopped working after #816


emankov:
With introducing temporary files in system temps by createTemporaryFile() option -inplace stopped working due to system error: "system cannot move the file to another disk".

PR #825, opened 2018-12-20
[HIPIFY][fix][#824] Fix the bug with -inplace option

emankov:
+ Do temp file copying instead of renaming
+ Replace all file routine functions with LLVM's
+ Do not output hipified file in case of errors


emankov:
Ping

PR #829, opened 2018-12-21
[HIPIFY][#812] Finishing with temps

emankov:
1. Option -temp-dir for temporary directory:
  + if not specified system temp is used;
  + creates the directory if the directory doesn't exist (only one level in a tree);
  + reports an error in case of a wrong directory specified, in case of necessity of creating a tree of subfolders, or in case of a filename specified.
2. Option -save-temps for preserving temporary files:
  + if specified temporary files are not being deleted from system temps and user temps as well.
3. Work with files in terms of calculated absolute paths by collapsing all '.' and '..' patterns, resolving symlinks and expanding '~' expression to the user's home directory:
  + to produce correct include paths;
  + to avoid possible errors on file routines.


emankov:
ping

PR #827, opened 2018-12-20
Fix issues with prior iteration of 731

AlexVlx:
Also, this is rebased against the 2.0 branch.

PR #791, opened 2018-12-07
Start re-working 731.

AlexVlx:
**DO NOT MERGE**

PR #841, opened 2019-01-01
Fix issues with prior iteration of 731

mangupta:
Rebased to master branch


mangupta:
@AlexVlx This is #827 rebased to master since this fix needs to go into 2.1 and not 2.0. Let me know if the cherry-picks look fine. I will merge this in and close #827 and #791.

PR #853, opened 2019-01-08
Introduce hash key to HIP module implementation

whchung:
A hash calculated via FNV-1a algorithm is introduced in ihipModule_t, the
internal of hipModule_t. The hash is used by HIP module APIs such as

- read_agent_global_from_module

to determine whether the agent-scope globals for a module have been iterated.

This commit fixes one issue that applications which load / unload modules
frequently would occasionally fail. After deep investigation of the issue it
turns out the old implementation in read_agent_global_from_module uses
hipModule_t as the key, which is not robust enough, as hipModule_t instances
are allocated dynamically so there are cases that one memory address may be
used by multiple hipModule_t instances. The real solution is to introduce a
uniquely identifiable hash for the code object associated with the HIP module.
And that's the rationale behind this commit.

### DIFF
--- a/src/hip_hcc_internal.h
+++ b/src/hip_hcc_internal.h
@@ -374,6 +374,7 @@ struct ihipModule_t {
     std::string fileName;
     hsa_executable_t executable = {};
     hsa_code_object_reader_t coReader = {};
+    std::string hash;
 
     ~ihipModule_t() {
         if (executable.handle) hsa_executable_destroy(executable);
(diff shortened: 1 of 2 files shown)


sunway513:
@mangupta @gargrahul , could you help expedite the review of this PR, TF-XLA path requires this to be included in ROCm2.1 release.

gargrahul:
LGTM, @whchung has already verified it with TF XLA

sunway513:
@gargrahul @mangupta could you confirm if this PR will be included in ROCm2.1 release?
cc @whchung @dagamayank

gargrahul:
@sunway513 Yes I see it is part of the release branch.

sunway513:
Thanks @gargrahul !

PR #856, opened 2019-01-09
[HIPIFY] Introduce CUDA installation path option '--cuda-path'

emankov:
Repeats clang's '--cuda-path' option.

[Reason]
In case of absence of any other clang's options setting '-cuda-path' allows not to specify separator '--' before clang's '--cuda-path'.

+ Tests and scripts are updated accordingly.


emankov:
Ping.

PR #863, opened 2019-01-11
Add Dot functions as amd_mixed_dot function

aaronenyeshi:
Introduce Dot functions which are available in the ROCm device library. Forward those prototypes, and introduce HIP API to expose the usage of the dot functions.

HIP Tests are passing.

### DIFF
--- a/include/hip/hcc_detail/math_functions.h
+++ b/include/hip/hcc_detail/math_functions.h
@@ -23,6 +23,7 @@ THE SOFTWARE.
 #pragma once
 
 #include "hip_fp16_math_fwd.h"
+#include "hip_vector_types.h"
 #include "math_fwd.h"
 
 #include <hip/hcc_detail/host_defines.h>
@@ -117,6 +118,43 @@ uint64_t __make_mantissa(const char* tagp)
     }
 
     return __make_mantissa_base10(tagp);
+}
+
+// DOT FUNCTIONS
+__DEVICE__
+inline
+float amd_mixed_dot(__2f16 a, __2f16 b, float c, bool saturate) {
+    return __ockl_fdot2(a, b, c, saturate);
+}
+__DEVICE__
+inline
+int amd_mixed_dot(short2 a, short2 b, int c, bool saturate) {
+    return __ockl_sdot2(a, b, c, saturate);
+}
+__DEVICE__
+inline
+uint amd_mixed_dot(ushort2 a, ushort2 b, uint c, bool saturate) {
+    return __ockl_udot2(a, b, c, saturate);
+}
+__DEVICE__
+inline
+int amd_mixed_dot(char4 a, char4 b, int c, bool saturate) {
+    return __ockl_sdot4(a, b, c, saturate);
+}
+__DEVICE__
+inline
+uint amd_mixed_dot(uchar4 a, uchar4 b, uint c, bool saturate) {
+    return __ockl_udot4(a, b, c, saturate);
+}
+__DEVICE__
+inline
+int amd_mixed_dot(int a, int b, int c, bool saturate) {
+    return __ockl_sdot8(a, b, c, saturate);
+}
+__DEVICE__
+inline
+uint amd_mixed_dot(uint a, uint b, uint c, bool saturate) {
+    return __ockl_udot8(a, b, c, saturate);
 }
 
 // BEGIN FLOAT
--- a/include/hip/hcc_detail/math_functions.h
+++ b/include/hip/hcc_detail/math_functions.h
@@ -123,11 +123,6 @@ uint64_t __make_mantissa(const char* tagp)
 // DOT FUNCTIONS
 __DEVICE__
 inline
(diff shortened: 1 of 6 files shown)


include/hip/hcc_detail/math_functions.h
```diff
@@ -119,6 +120,43 @@ uint64_t __make_mantissa(const char* tagp)
     return __make_mantissa_base10(tagp);
 }
 
+// DOT FUNCTIONS
+__DEVICE__
+inline
+float amd_mixed_dot(__2f16 a, __2f16 b, float c, bool saturate) {
```
b-sumner: I'm not sure this is the best argument type.  Developers will probably want to use half2.  This function may need to bitcast a and b to the ockl types.

aaronenyeshi: @AlexVlx I noticed that half2 is being defined in hip_fp16.h as a struct. Do we want to change half2 into ockl type half2? or keep it as is? What are your thoughts

b-sumner: I think the types of the and and b arguments should be whatever developers are likely to use, and that we should translate (i.e. bitcast) it to the type ockl expects in this function.  I think they are more likely to use half2 than __2f16.

AlexVlx: @aaronenyeshi You absolutely do not want to touch the types that are part of the HIP public interface, which must match CUDA behaviourally. This is part of HIP's contract, those are law, you do not want to turn them into ockl anything. You want the AMD specific stuff to interoperate seamlessly, so you want the public interface to take HIP types. `__half2_raw::data` gives you exactly what you want.

aaronenyeshi:
I've moved the fp16 versions of dot functions into HIP's hip_fp16.h and hip_fp16_math_fwds.h, so that its enabled only when user is using HIP's fp16 . Also changed to using __half2, and conversion into ockl version of _Float16 __attribute__((ext_vector_type(2))) type.

aaronenyeshi:
I've updated this PR as per Alex's and Brian's review. Please let me know if this is okay. @AlexVlx  @b-sumner

b-sumner:
This should probably include an update to the half2 tests in HIP/tests/src/deviceLib/hipTestNativeHalf.cpp.  It would be good to see in the IR and ISA that the casts in the half2 variant don't actually add ISA instructions or require the linker to generate additional bitcasts.

aaronenyeshi:
@mangupta can you help this PR? I am not sure why Jenkins is failing. It is passing on my machine.

mangupta:
@aaronenyeshi HIP codebase needs to be able to build+run on latest public ROCm (which is 2.0.x right now) as well as the latest internal ROCm codebase. So HIP CI builds and tests with both ROCm versions.
Looking at the logs I see that on rocm-2.0.x the following 2 tests are not building:
```
	 directed_tests/deviceLib/hipTestNativeHalf.tst (Not Run)
	 directed_tests/deviceLib/hipTestDotFunctions.tst (Not Run)
```
Also the following test is not building on nvcc path:
```
	 directed_tests/deviceLib/hipTestDotFunctions.tst (Not Run)
```
I am not sure why there is build failure on rocm-2.0.x since we have the hcc workweek guard in place. Maybe we are missing the guard in some place. I will re-review the code and see what I can find. However on nvcc, you probably need a ```NVCC_OPTIONS -std=c++11``` in the build command of hipTestDotFunctions.cpp due to the use of ```nullptr```.

b-sumner:
The dot tests should not be run or pass trivially on nvcc path.

aaronenyeshi:
Thank you @mangupta . I was able to fix the tests for both cuda path and rocm 2.0.x path. Cuda path will be excluded now, and rocm 2.0.x path needed fix in HIP-Clang checking. Now it show pass all Jenkins: rocm_head, rocm_2_0, and cuda_9__x.

aaronenyeshi:
```
[cuda_9_x] 100% tests passed, 0 tests failed out of 106
[cuda_9_x] 
[cuda_9_x] Total Test time (real) = 141.22 sec

[rocm_2_1] 99% tests passed, 1 tests failed out of 120
[rocm_2_1] 
[rocm_2_1] Total Test time (real) = 114.49 sec
[rocm_2_1] 
[rocm_2_1] The following tests FAILED:
[rocm_2_1] 	 92 - directed_tests/runtimeApi/multiThread/hipMultiThreadDevice-pyramid.tst (SEGFAULT)

[rocm_2_0] 100% tests passed, 0 tests failed out of 120
[rocm_2_0] 
[rocm_2_0] Total Test time (real) = 114.42 sec

[rocm_head] 100% tests passed, 0 tests failed out of 120
[rocm_head] 
[rocm_head] Total Test time (real) = 114.65 sec
```

b-sumner:
Looks ready to go to me...

aaronenyeshi:
@mangupta Please help review and merge, so it makes it into next release. Thank you

PR #864, opened 2019-01-14
Switch to faster ocml variants for trig instrinsics

mangupta:


b-sumner:
Looks fine to me.  There are other functions there that we could consider speeding up using OCML's native variants.

b-sumner:
Looks good to me, thanks!

PR #866, opened 2019-01-14
[HIPIFY] Introduce '-D' and '-I' options as hipify-clang's ones

emankov:
+ '-D' and '-I' might be set as hipify-clang options (before separator '--' or without specifying separator at all);
+ '-D' or '-I' as a clang options might be specified as well (after options separator '--');
+ All defines and include directories specified as hipify-clang's (before options separator '--') will be set after clang's ones and will redefine the same clang's ones.


hipify-clang/src/ArgParse.h
```diff
@@ -33,6 +33,8 @@ extern cl::opt<std::string> OutputFilename;
 extern cl::opt<std::string> OutputDir;
 extern cl::opt<std::string> TemporaryDir;
 extern cl::opt<std::string> CudaPath;
+extern cl::list<std::string> I;
```
dfukalov: I suggest to rename these two variables to something like IncludeDirs and MacroNames since they are global and one-letter names can intersect with local variables

hipify-clang/src/ArgParse.cpp
```diff
@@ -85,8 +85,22 @@ cl::opt<bool> Examine("examine",
...
+cl::list<std::string> I("I",
+  cl::desc("Add directory to include search path"),
+  cl::value_desc("directory"),
+  cl::ZeroOrMore,
+  cl::Prefix,
+  cl::cat(ToolTemplateCategory));
+
+cl::list<std::string> D("D",
+  cl::desc("Define <macro> to <value> or 1 if <value> omitted"),
+  cl::value_desc("macro>=<value"),
```
dfukalov: typo?

emankov: `macro>=<value` ? No, it is not, outter angle brackets will be substituted by clang tooling - it might be seen in hipify-clang output by setting '-help' option.

hipify-clang/src/ArgParse.h
emankov: Well, the idea was to mimic to clang's options in order not to set options separator '--' for specifying those options ('-I' and '-D') for clang. Moreover, those options become clang's options. As for local variables, they will differ anyway.

emankov: Ok, I got what you meant. Let's they be IncludeDirs and MacroNames.

hipify-clang/src/main.cpp
```diff
@@ -199,6 +199,18 @@ int main(int argc, const char **argv) {
 #if defined(HIPIFY_CLANG_RES)
     Tool.appendArgumentsAdjuster(ct::getInsertArgumentAdjuster("-resource-dir=" HIPIFY_CLANG_RES));
 #endif
+    if (!MacroNames.empty()) {
```
dfukalov: possible this and second checks are redundant
otherwise, lgtm

Issue #878, opened 2019-01-23
Segfault caused by hipMemcpy2DAsync


xiaocenxiaocen:
- My test code is as follows:
memcpy_test.cpp:
```
#define __HIP_PLATFORM_HCC__

#include <hip/hip_runtime.h>
#include <hip/hip_runtime_api.h>

int main(int argc, char* argv[]) {
    float* d_src;
    float* d_dst;
    int M = 25600;
    int N = 1;
    int spitch = 9;
    int dpitch = 1;
    hipMalloc(&d_src, M*spitch*sizeof(float));
    hipMalloc(&d_dst, M*dpitch*sizeof(float));

    hipStream_t stream;
    hipStreamCreateWithFlags(&stream, hipStreamNonBlocking);
    hipMemcpy2DAsync(d_dst, dpitch * sizeof(float), d_src, spitch * sizeof(float),
                     N * sizeof(float), M, hipMemcpyDeviceToDevice, stream);
    hipFree(d_src);
    hipFree(d_dst);
}
```
**Compile the code with the following command**
```
hipcc -Wall -std=c++14 memcpy_test.cpp -L/opt/rocm/lib -lhip_hcc
```
- hip version is 1.5.18494

Run this code with
```
./a.out
```
The program crashed with segmentation fault.
The backtrace of gdb is 
```
#0  0x00007ffff66ae262 in ?? () from /opt/rocm/hsa/lib/libhsa-runtime64.so.1
#1  0x00007fffef469ccb in waitComplete () at /home/jenkins/jenkins-root/workspace/compute-rocm-rel-2.0/external/hcc-tot/lib/hsa/mcwamp_hsa.cpp:5128
#2  0x00007fffef455372 in ~HSACopy () at /home/jenkins/jenkins-root/workspace/compute-rocm-rel-2.0/external/hcc-tot/lib/hsa/mcwamp_hsa.cpp:828
#3  0x00007fffef46fc1b in _M_release () at /usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/bits/shared_ptr_base.h:150
#4  ~__shared_count () at /usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/bits/shared_ptr_base.h:659
[...]

PR #881, opened 2019-01-23
Add env var for 2D memcpy through SDMA

gargrahul:
Default path uses 2D Memcpy kernel

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -1637,7 +1637,15 @@ hipError_t ihipMemcpy2D(void* dst, size_t dpitch, const void* src, size_t spitch
                     stream->locked_copySync((unsigned char*)dst + i * dpitch,
                                     (unsigned char*)src + i * spitch, width, kind);
             } else {
-                stream->locked_copy2DSync(dst, src, width, height, spitch, dpitch, kind);
+                int HIP_MEMCPY2D_FORCE_SDMA = 0;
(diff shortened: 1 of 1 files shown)


src/hip_memory.cpp
```diff
@@ -1637,7 +1637,15 @@ hipError_t ihipMemcpy2D(void* dst, size_t dpitch, const void* src, size_t spitch
                     stream->locked_copySync((unsigned char*)dst + i * dpitch,
                                     (unsigned char*)src + i * spitch, width, kind);
             } else {
-                stream->locked_copy2DSync(dst, src, width, height, spitch, dpitch, kind);
+                int HIP_MEMCPY2D_FORCE_SDMA = 0;
+                ihipReadEnv_I(&HIP_MEMCPY2D_FORCE_SDMA, "HIP_MEMCPY2D_FORCE_SDMA", "HIP_MEMCPY2D_FORCE_SDMA",
```
mangupta: Why are we reading this env variable in every memcpy2D? Is it a requirement for the user to be able to turn this on or off while the process is still running? We should probably handle it like the rest of the env variables that are read once during init.
Also we seem to be changing the default from SDMA enabled to SDMA disabled.

gargrahul: Yes your comments are valid though I created this PR just to help checking and unblocking #878 . If we keep this as long term change then I will update this before you merge it. Let us wait for 2.1 test results on stability of SDMA. If SDMA issues persist then we can keep default path using 2D kernel for now.

gargrahul:
Closing this since https://github.com/ROCm-Developer-Tools/HIP/pull/947 should help when SDMA path doesn't work.

PR #882, opened 2019-01-24
enabling PROF_API pkgconfig

eshcherb:


CMakeLists.txt
```diff
@@ -155,7 +155,12 @@ add_to_config(_buildInfo COMPILE_HIP_ATP_MARKER)
 ################
 if(USE_PROF_API EQUAL 1)
     if(NOT DEFINED PROF_API_HEADER_PATH)
-        set(PROF_API_HEADER_PATH /opt/rocm/roctracer/include/ext)
+        if(PROF_API_PKG_CONFIG EQUAL 1)
```
mangupta: I don't like the idea of adding one more argument to the cmake commandline. Maybe a better approach is to have something like ```pkg_check_modules(ROCTRACERPROTO QUIET roctracer-proto)```. Then we can use ```ROCTRACERPROTO_FOUND``` to determine if we want to use ```pkg_get_variable``` or explicitlly set the ```PROF_API_HEADER_PATH```.

PR #884, opened 2019-01-25
roctracer-proto header find_path

eshcherb:


CMakeLists.txt
```diff
@@ -154,19 +154,16 @@ add_to_config(_buildInfo COMPILE_HIP_ATP_MARKER)
...
+            ${PROF_API_HEADER_PATH}
+        PATHS
+            /opt/rocm/roctracer
+        PATH_SUFFIXES
+            include/ext
+    )
     if(NOT PROF_API_HEADER_DIR)
-        MESSAGE(STATUS "PROF_API_HEADER_PATH = ${PROF_API_HEADER_PATH}")
-        MESSAGE(WARNING "Profiling API header not found. Disabling roctracer integration. Use -DPROF_API_HEADER_PATH=<path to prof_protocol.h header>")
+        MESSAGE(FATAL_ERROR "Profiling API header not found. Disabling roctracer integration. Use -DPROF_API_HEADER_PATH=<path to prof_protocol.h header>")
```
mangupta: I don't think this should be a fatal error. We can still build HIP runtime albeit without roctracer integration. So it is sufficient to give a warning that the profiling API header is not found and roctracer integration is disabled. This is important for CI until 2.1.x is released. Also we still have folks building tip of HIP on 2.0.x.

eshcherb: 'pkg-config' is removed with find_path.
I just worry that if the roctracer-proto header will be installed in wrong place or not installed by mistake it will be easy to check with the error.
Could they build 2.0.x version without USE_PROF_API enabled?

mangupta:
@eshcherb There are merge conflicts with this PR. Please resolve them in case this PR is still needed. Otherwise we can go ahead and close this.

PR #885, opened 2019-01-29
roctracer-proto pkg-config fix to includedir

eshcherb:


mangupta:
@eshcherb I guessing that this implies that #884 is no longer required? Some of the changes are the same in both the PRs. Anyways, going ahead and merging this in. In case the other PR is required, it needs to be rebased due to conflicts.

PR #891, opened 2019-02-01
hipcc defaults to HIP-Clang if built with HIP-Clang

aaronenyeshi:
Add HIP_COMPILER to hipConfig, so that HIP packages built with HIP_COMPILER for HIP-Clang will be known during runtime by hipcc. Also add default locations to check for llvm and device-libs.

### DIFF
--- a/bin/hipcc
+++ b/bin/hipcc
@@ -58,8 +58,8 @@ $HIPCC_LINK_FLAGS_APPEND=$ENV{'HIPCC_LINK_FLAGS_APPEND'};
 
 $HIP_PATH=$ENV{'HIP_PATH'} // dirname (dirname $0); # use parent directory of hipcc
 $HIP_VDI_HOME=$ENV{'HIP_VDI_HOME'};
-$HIP_CLANG_PATH=$ENV{'HIP_CLANG_PATH'};
-$DEVICE_LIB_PATH=$ENV{'DEVICE_LIB_PATH'};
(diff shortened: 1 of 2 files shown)


bin/hipcc
```diff
@@ -58,8 +58,8 @@ $HIPCC_LINK_FLAGS_APPEND=$ENV{'HIPCC_LINK_FLAGS_APPEND'};
 
 $HIP_PATH=$ENV{'HIP_PATH'} // dirname (dirname $0); # use parent directory of hipcc
 $HIP_VDI_HOME=$ENV{'HIP_VDI_HOME'};
-$HIP_CLANG_PATH=$ENV{'HIP_CLANG_PATH'};
-$DEVICE_LIB_PATH=$ENV{'DEVICE_LIB_PATH'};
+$HIP_CLANG_PATH=$ENV{'HIP_CLANG_PATH'} // "/opt/rocm/llvm/bin";
```
yxsamliu: these two lines will break HIP_VDI_HOME

we want HIP_VDI_HOME still works if it is set. see line 99-110 for its logic.

basically, HIP_VDI_HOME should override default clang path and dev lib path
HIP_CLANG_PATH and DEVICE_LIB_PATH will override HIP_VDI_HOME

aaronenyeshi: Okay I just changed it, thank you for catching that

aaronenyeshi:
Sorry, I missed a semi colon, I force pushed it back in for Jenkins tests. @yxsamliu Please review again, thanks

PR #910, opened 2019-02-08
hipcc should consume -mcode-object-v3 flag

aaronenyeshi:
Allow hipcc to accept the code-object-v3 flag, and pass it to clang. Some changes in clang upstream as well.


aaronenyeshi:
Please help merge this, as HIP Toolchain changes were made upstream: https://reviews.llvm.org/D57977

PR #911, opened 2019-02-08
Allow HIP_LIB_PATH to be overriden at the command line. Use HIP_LIB_P…

lmoriche:
…ATH for rpath instead of HIP_PATH/lib


aaronenyeshi:
I'm running into error:
`Use of uninitialized value $HIP_LIB_PATH in concatenation (.) or string at /opt/rocm/hip/bin/hipcc line 770.` 
In regular HIP/HCC use-case. Need to set a default value after VDI doesn't set one.

bin/hipcc
```diff
@@ -754,9 +757,9 @@ if ($setStdLib eq 0 and $HIP_PLATFORM eq 'hcc')
 
 if ($needHipHcc) {
     if ($linkType eq 0) {
```
yxsamliu: check if HIP_LIB_PATH is defined, if not set it to $HIP_PATH/lib here

PR #912, opened 2019-02-09
Remove C++ function templates for hipMalloc and hipHostMalloc

whchung:
This allows hipMalloc and hipHostMalloc be true C functions from application
perspective at compile time, so decltype(hipMalloc) and decltype(hipHostMalloc)
can be computed.

All HIP applications shall now correctly use C-style cast or reinterpret_cast
when invoking hipMalloc and hipHostMalloc.

### DIFF
--- a/include/hip/hip_runtime_api.h
+++ b/include/hip/hip_runtime_api.h
@@ -318,9 +318,13 @@ enum hipComputeMode {
  *
  * Perform automatic type conversion to eliminate need for excessive typecasting (ie void**)
  *
+ * __HIP_DISABLE_CPP_FUNCTIONS__ macro can be defined to suppress these
+ * wrappers. It is useful for applications which need to obtain decltypes of
+ * HIP runtime APIs.
+ *
  * @see hipMalloc
  */
-#ifdef __cplusplus
+#if defined(__cplusplus) && !defined(__HIP_DISABLE_CPP_FUNCTIONS__)
 template <class T>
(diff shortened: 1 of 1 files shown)


whchung:
This proposed patch is to address SWDEV-178514.

pfultz2:
I think this might break some users who are not using the `(void**)` casts. How about surrounding the code with `#ifdef __cplusplus` instead of deleting?

gargrahul:
@pfultz2 Current code already has `#ifdef __cplusplus' https://github.com/ROCm-Developer-Tools/HIP/blame/master/include/hip/hip_runtime_api.h#L323 and the purpose of this code was to avoid the void** casts as mentioned at https://github.com/ROCm-Developer-Tools/HIP/blame/master/include/hip/hip_runtime_api.h#L319

pfultz2:
I see.

> the purpose of this code was to avoid the void** casts as mentioned

Yes, but isn't it possible some users rely on this?

Perhaps, hip could do something like `#if defined(__cplusplus) && !defined(HIP_DISABLE_CPP_FUNCTIONS)` and then tensorflow/hip could add `-DHIP_DISABLE_CPP_FUNCTIONS` to the build to avoid these overloads?

whchung:
Will update this PR and add `HIP_DISABLE_CPP_FUNCTIONS` to the header file.

whchung:
ping

PR #915, opened 2019-02-13
consolidate thread local storage

jeffdaily:
Attempt to reduce the cost of `__tls_get_addr`.  Thread local storage can be expensive when used within a shared library, e.g., libhip_hcc.so.  Combine all thread local storage (TLS) variables into a single TLS struct accessed at most once per public HIP function.  Some internal functions are now macros.  Some internal functions now explicitly passed the TLS pointer.


src/hip_texture.cpp
```diff
@@ -381,7 +381,7 @@ hipError_t hipGetTextureObjectTextureDesc(hipTextureDesc* pTexDesc,
 }
 
 // Texture Reference APIs
-hipError_t ihipBindTextureImpl(int dim, enum hipTextureReadMode readMode, size_t* offset,
+hipError_t ihipBindTextureImpl(TlsData *tls, int dim, enum hipTextureReadMode readMode, size_t* offset,
```
mangupta: ```ihipBindTextureImpl``` is used by the cplusplus variants of ```hipBindTexture``` defined in include/hip/hcc_detail/hip_runtime_api.h. Hence this change breaks the unit test directed_test/texture/hipBindTexRef1DFetch.

src/hip_texture.cpp
```diff
@@ -519,12 +519,12 @@ hipError_t hipBindTexture2D(size_t* offset, textureReference* tex, const void* d
...
     HIP_INIT_API(hipBindTexture2D, offset, tex, devPtr, desc, width, height, pitch);
     hipError_t hip_status = hipSuccess;
-    hip_status = ihipBindTexture2DImpl(hipTextureType2D, hipReadModeElementType, offset, devPtr,
+    hip_status = ihipBindTexture2DImpl(tls, hipTextureType2D, hipReadModeElementType, offset, devPtr,
                                        desc, width, height, tex);
     return ihipLogStatus(hip_status);
 }
 
-hipError_t ihipBindTextureToArrayImpl(int dim, enum hipTextureReadMode readMode,
+hipError_t ihipBindTextureToArrayImpl(TlsData *tls, int dim, enum hipTextureReadMode readMode,
```
mangupta: ```ihipBindTextureToArrayImpl``` is used by the cplusplus variants of ```hipBindTextureToArrray``` defined in include/hip/hcc_detail/hip_runtime_api.h. Hence this change breaks the unit test directed_test/texture/hipTextureRef2D.

jeffdaily:
@mangupta related question.  For the cplusplus variants of these functions, why do they not call HIP_INIT_API?

mangupta:
> @mangupta related question. For the cplusplus variants of these functions, why do they not call HIP_INIT_API?
Definitely a bug. I'll raise a PR to fix that. Thanks for noticing it though.

jeffdaily:
@mangupta Let me fix the build/merge with latest master.

PR #921, opened 2019-02-18
Avoid including hc.hpp within program_state.hpp

whchung:
hc.hpp would open a plethora of weird issues in certain HIP applications such
as TensorFlow. Try stick with vanilla ROCR runtime API calls in this commit.


whchung:
The issue with `hc.hpp` seems a bit more bizarre and less easy to illustrate.

There are some other issues with HIP headers after changes proposed in this PR, this is easier to document:

TensorFlow is built with `-fno-exceptions`, and in `ELFIO` and `program_state.hpp` there are places where `try/catch` are used. We'll probably need to change them as macros which can be turned on/off at compile time.

PR #929, opened 2019-02-25
dlopen() fixes

AlexVlx:
This makes various internal structures which make up HIP's state be unique per shared object, and interposition free; it also moves initialisation to run time, cleaning up the kludgy static init related hacks. This makes HIP robust for cases where either its libraries or a library which includes HIP kernel dispatches is loaded at run time via e.g. `dlopen`.


whchung:
Perhaps a better subject than "Re-sync with upstream"?

AlexVlx:
@whchung that is a truly insightful review, I will not lie...

include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -35,11 +35,14 @@ THE SOFTWARE.
 #define GENERIC_GRID_LAUNCH 1
 #endif
 
+#include <hsa/hsa.h>
```
whchung: we are getting weird issues with this line running TF on CI. I'll try get more info on this.

include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -1363,6 +1371,61 @@ hipError_t hipMemcpyDtoHAsync(void* dst, hipDeviceptr_t src, size_t sizeBytes, h
 hipError_t hipMemcpyDtoDAsync(hipDeviceptr_t dst, hipDeviceptr_t src, size_t sizeBytes,
                               hipStream_t stream);
 
+__attribute__((visibility("hidden")))
```
whchung: inline?

include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -1363,6 +1371,61 @@ hipError_t hipMemcpyDtoHAsync(void* dst, hipDeviceptr_t src, size_t sizeBytes, h
...
+ * @param[out] devPtr - Pointer to a pointer to the memory referred to by the symbol
+ * @return #hipSuccess, #hipErrorNotInitialized, #hipErrorNotFound
+ *
+ *  @see hipGetSymbolSize, hipMemcpyToSymbol, hipMemcpyFromSymbol, hipMemcpyToSymbolAsync,
+ * hipMemcpyFromSymbolAsync
+ */
+inline
+__attribute__((visibility("hidden")))
+hipError_t hipGetSymbolAddress(void** devPtr, const void* symbolName) {
+    //HIP_INIT_API(hipGetSymbolAddress, devPtr, symbolName);
```
whchung: do we need `HIP_INIT_API` here? if not perhaps this line can be removed? same goes to the one in `hipGetSymbolSize`

src/hip_memory.cpp
```diff
@@ -955,37 +955,14 @@ hipError_t hipHostUnregister(void* hostPtr) {
     return ihipLogStatus(hip_status);
 }
 
-namespace {
-inline hipDeviceptr_t agent_address_for_symbol(const char* symbolName) {
```
whchung: @AlexVlx `agent_address_for_symbol()` is now removed, and we use `hipGetSymbolAddress()` instead, right?

include/hip/hcc_detail/hip_runtime_api.h
AlexVlx: I have no idea, since I have no idea what those macros are supposed to do (beyond being very noisy), but probably? However, I'm not going to sort out how to deal with all that noise in this particular iteration (and someone more familiar with the thing should do it anyway). So this should stay there as a reminder it's disabled for now.

include/hip/hcc_detail/hip_runtime_api.h
AlexVlx: I was under the impression that this was verified, which was the reason for asking for a PR...

include/hip/hcc_detail/program_state.hpp
```diff
@@ -93,16 +128,517 @@ class Kernel_descriptor {
...
+inline
+ELFIO::section* find_section_if(ELFIO::elfio& reader, P p) {
+    const auto it = std::find_if(
+        reader.sections.begin(), reader.sections.end(), std::move(p));
+
+    return it != reader.sections.end() ? *it : nullptr;
+}
+
+inline
+__attribute__((visibility("hidden")))
```
whchung: @AlexVlx could you say more about the rule for functions inside `hip_impl` namespace to carry `hidden` visibility or not? My understanding is that as long as they carry static data structures they are marked as hidden. Is this understanding correct?

AlexVlx: Yes, it is.

include/hip/hcc_detail/hip_runtime_api.h
whchung: it could be an environment issue. don't worry about it too much (yet)

include/hip/hcc_detail/hip_runtime_api.h
yxsamliu: I think the inline implementation of these functions are for HIP/HCC runtime only. Since this header file is shared between HIP/VDI runtime and HIP/HCC runtime, it is better to move them to a separate header file, e.g. hip_module_hcc.hpp. Then add conditions, e.g.

`
#if __hcc__
#include "hip_module_hcc.hpp"
#else
// keep original declarations 
#endif
`

include/hip/hcc_detail/hip_runtime_api.h
gargrahul: HIP_INIT_API is called from all hip (public) APIs for initializing hip. This behavior is needed to match CUDA runtime APIs.

include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -1387,35 +1450,36 @@ hipError_t hipMemcpyDtoDAsync(hipDeviceptr_t dst, hipDeviceptr_t src, size_t siz
...
-                             size_t offset __dparm(0), hipMemcpyKind kind __dparm(hipMemcpyHostToDevice));
+inline
+__attribute__((visibility("hidden")))
+hipError_t hipMemcpyToSymbol(const void* symbolName, const void* src,
+                             size_t sizeBytes, size_t offset __dparm(0),
+                             hipMemcpyKind kind __dparm(hipMemcpyHostToDevice)) {
+    if (!symbolName) return hipErrorInvalidSymbol;
 
+    hipDeviceptr_t dst = NULL;
+    hipGetSymbolAddress(&dst, (const char*)symbolName);
```
gargrahul: HIP API calling another HIP API isn't a good idea. This should be modified to use an ihipGetSymbolAddress.

include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -1387,35 +1450,36 @@ hipError_t hipMemcpyDtoDAsync(hipDeviceptr_t dst, hipDeviceptr_t src, size_t siz
...
  * hipMemcpyFromSymbolAsync
  */
-hipError_t hipMemcpyToSymbol(const void* symbolName, const void* src, size_t sizeBytes,
-                             size_t offset __dparm(0), hipMemcpyKind kind __dparm(hipMemcpyHostToDevice));
+inline
+__attribute__((visibility("hidden")))
+hipError_t hipMemcpyToSymbol(const void* symbolName, const void* src,
+                             size_t sizeBytes, size_t offset __dparm(0),
+                             hipMemcpyKind kind __dparm(hipMemcpyHostToDevice)) {
+    if (!symbolName) return hipErrorInvalidSymbol;
```
gargrahul: HIP_INIT_API should be used here?

include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -2379,11 +2576,20 @@ hipError_t hipFuncGetAttributes(hipFuncAttributes* attr, const void* func);
  *
  * @returns hipSuccess, hipErrorInvalidValue, hipErrorNotInitialized
  */
-hipError_t hipModuleGetGlobal(hipDeviceptr_t* dptr, size_t* bytes, hipModule_t hmod,
-                              const char* name);
+inline
+__attribute__((visibility("hidden")))
+hipError_t hipModuleGetGlobal(hipDeviceptr_t* dptr, size_t* bytes,
+                              hipModule_t hmod, const char* name) {
+    if (!dptr || !bytes) return hipErrorInvalidValue;
```
mangupta: HIP_INIT_API()...

include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -2368,6 +2468,103 @@ hipError_t hipModuleGetFunction(hipFunction_t* function, hipModule_t module, con
...
+hsa_agent_t this_agent();
+
+inline
+__attribute__((visibility("hidden")))
+hipError_t read_agent_global_from_module(hipDeviceptr_t* dptr, size_t* bytes,
+                                         hipModule_t hmod, const char* name) {
+    // the key of the map would the hash of code object associated with the
+    // hipModule_t instance
+    static std::unordered_map<
+        std::string, std::vector<Agent_global>> agent_globals;
```
scchan: Could we only expose a static object that manages the lifetime of these globals in this header and move/hide all the implementation details in a .cpp file?  One motivation is shielding the algorithm from hazardous C++ environment of some users because this code would get generated during HIP's build rather than in the user environment.  As a bonus it would reduce compilation time of user applications and produce smaller code size.

src/hip_hcc.cpp
```diff
@@ -2460,3 +2461,28 @@ hipError_t hipHccGetAcceleratorView(hipStream_t stream, hc::accelerator_view** a
 
 //// TODO - add identifier numbers for streams and devices to help with debugging.
 // TODO - add a contect sequence number for debug. Print operator<< ctx:0.1 (device.ctx)
+
+namespace hip_impl {
+    std::vector<hsa_agent_t> all_hsa_agents() {
```
scchan: The caller of this function is in a header and it could be problematic to some user environment if the header and the source have a different implementation of std::vector for some reason...

include/hip/hcc_detail/hip_runtime_api.h
AlexVlx: That's a brilliant idea and you should absolutely pursue it as a further development. Lots of bonuses, as you've observed!

src/hip_hcc.cpp
AlexVlx: You have more problems than that if for some reason you are mixing implementations of std::vector...

include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -1442,14 +1506,50 @@ hipError_t hipGetSymbolSize(size_t* size, const void* symbolName);
  * hipMemcpyFromArrayAsync, hipMemcpy2DFromArrayAsync, hipMemcpyToSymbolAsync,
  * hipMemcpyFromSymbolAsync
  */
-hipError_t hipMemcpyToSymbolAsync(const void* symbolName, const void* src, size_t sizeBytes,
-                                  size_t offset, hipMemcpyKind kind, hipStream_t stream __dparm(0));
+inline
+__attribute__((visibility("hidden")))
+hipError_t hipMemcpyToSymbolAsync(const void* symbolName, const void* src,
```
scchan: Implementing the hipMemcpyToSymbol* APIs as inline functions in a header won't work with the dlsym() use case that some users are asking for, can this be moved into a .cpp file?

include/hip/hcc_detail/hip_runtime_api.h
scchan: I'm not suggesting we do this in this PR but is there any architecture assumption made that would prevent moving all the machinery into a .cpp file impossible?

mangupta:
Merging this in with the following caveats:
* We need a follow on patch that fixes issues this PR introduces in HIP/VDI (if any).
* We need a follow on patch that addresses the HIP init issue and nested calls to HIP APIs.

include/hip/hcc_detail/hip_runtime_api.h
scchan: `hipLaunchKernelGGL` not calling `HIP_INIT_API` is serious issue.  I think it might be reason why HIP is broken if the HCC runtime is set to lazy init.  Could we prioritize on fixing the `HIP_INIT_API` issue?

PR #296, opened 2017-12-14
hipMemset refactoring

gargrahul:
- Added hipMemset2DAsync
- Added hipMemsetD8Async

PR #933, opened 2019-02-25
Add HIP memset APIs to cope with non-zero initial values of integer types

whchung:
`hipMemsetAsync` and `hipMemset` takes `int` as input so we really should use the copy kernel which copies `int`.

This PR fixes some failed cases in TensorFlow where non-zero values are passed to `hipMemsetAsync`.

Also it seems there is no test case for `hipMemsetAsync`, I'll try amend this PR to provide one.

### DIFF
--- a/include/hip/hcc_detail/hip_runtime_api.h
+++ b/include/hip/hcc_detail/hip_runtime_api.h
@@ -1505,6 +1505,17 @@ hipError_t hipMemset(void* dst, int value, size_t sizeBytes);
 hipError_t hipMemsetD8(hipDeviceptr_t dest, unsigned char value, size_t sizeBytes);
 
 /**
+ *  @brief Fills the memory area pointed to by dest with the constant integer
+ * value for specified number of times.
+ *
+ *  @param[out] dst Data being filled
+ *  @param[in]  constant value to be set
+ *  @param[in]  number of values to be set
+ *  @return #hipSuccess, #hipErrorInvalidValue, #hipErrorNotInitialized
+ */
+hipError_t hipMemsetD32(hipDeviceptr_t dest, int value, size_t count);
+
+/**
  *  @brief Fills the first sizeBytes bytes of the memory area pointed to by dev with the constant
  * byte value value.
  *
@@ -1520,6 +1531,23 @@ hipError_t hipMemsetD8(hipDeviceptr_t dest, unsigned char value, size_t sizeByte
  *  @return #hipSuccess, #hipErrorInvalidValue, #hipErrorMemoryFree
  */
 hipError_t hipMemsetAsync(void* dst, int value, size_t sizeBytes, hipStream_t stream __dparm(0));
+
+/**
+ *  @brief Fills the memory area pointed to by dev with the constant integer
+ * value for specified number of times.
+ *
+ *  hipMemsetD32Async() is asynchronous with respect to the host, so the call may return before the
+ * memset is complete. The operation can optionally be associated to a stream by passing a non-zero
+ * stream argument. If stream is non-zero, the operation may overlap with operations in other
+ * streams.
+ *
+ *  @param[out] dst Pointer to device memory
+ *  @param[in]  value - Value to set for each byte of specified memory
+ *  @param[in]  count - number of values to be set
+ *  @param[in]  stream - Stream identifier
+ *  @return #hipSuccess, #hipErrorInvalidValue, #hipErrorMemoryFree
+ */
+hipError_t hipMemsetD32Async(void* dst, int value, size_t count, hipStream_t stream __dparm(0));
 
 /**
  *  @brief Fills the memory area pointed to by dst with the constant value.
--- a/include/hip/hcc_detail/hip_runtime_api.h
+++ b/include/hip/hcc_detail/hip_runtime_api.h
@@ -1547,7 +1547,8 @@ hipError_t hipMemsetAsync(void* dst, int value, size_t sizeBytes, hipStream_t st
  *  @param[in]  stream - Stream identifier
  *  @return #hipSuccess, #hipErrorInvalidValue, #hipErrorMemoryFree
  */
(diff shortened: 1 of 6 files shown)


whchung:
Please hold reviewing this PR for now. It seems in TensorFlow XLA some kernels would try fill the memory with non-zero initial values, some with unsigned char types, some with integer types. More detailed analysis would be delivered to help better understand the desired semantic for hipMemset and hipMemsetAsync.

whchung:
Instead of changing the semantic of existing HIP APIs it may be better to introduce new APIs and let client applications to adopt the new APIs. I'll revise this PR.

gargrahul:
@whchung I think you are looking for equivalent of cuMemsetD32 ?

whchung:
Yes in my working repo I’ve got hipMemsetD32 and hipMemsetD32Async. Going to polish it up and submit over this PR

whchung:
@gargrahul the first pass has been completed. please help review.

also while trying to work on a test case, I found it surprising that all 3 tests related to `hipMemset` / `hipMemset2D` / `hipMemset3D` always return `passed()` no matter what. I think the original tests should be rectified first.

gargrahul:
@whchung Sure. I had added hipMemsetDxx functions with https://github.com/ROCm-Developer-Tools/HIP/pull/296/files but that PR was not merged since I messed up git commits there and looks like we didn't add them later.

Yes, this always passing issue has been there in few tests since beginning. We have rectified few already. I can rectify hipMemset test.

gargrahul:
@whchung #939 is a quick attempt. It includes hipMemsetAsync test too. This test can be further modified to accommodate testing of hipMemsetDxx APIs.

whchung:
jenkins: retest this please

whchung:
jenkins: retest this please

whchung:
jenkins: retest this please

whchung:
@gargrahul Thanks for your PR #939 . I've add unit tests for the new APIs accordingly.

gargrahul:
@whchung You haven't mapped hipMemsetD32/D32Async on HIP/NVCC. Those should get mapped to cuMemsetD32/D32Async respectively.

whchung:
Got it. Let me fix that

whchung:
jenkins: retest this please

whchung:
@gargrahul tried to add implementation for NVCC path. I don't really have access to a NV machine at this moment so I'm not exactly sure if the commit is correct.

include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -1521,6 +1532,23 @@ hipError_t hipMemsetD8(hipDeviceptr_t dest, unsigned char value, size_t sizeByte
...
+ * stream argument. If stream is non-zero, the operation may overlap with operations in other
+ * streams.
+ *
+ *  @param[out] dst Pointer to device memory
+ *  @param[in]  value - Value to set for each byte of specified memory
+ *  @param[in]  count - number of values to be set
+ *  @param[in]  stream - Stream identifier
+ *  @return #hipSuccess, #hipErrorInvalidValue, #hipErrorMemoryFree
+ */
+hipError_t hipMemsetD32Async(void* dst, int value, size_t count, hipStream_t stream __dparm(0));
```
mangupta: I believe this should be hipDeviceptr_t dst here.

include/hip/nvcc_detail/hip_runtime_api.h
```diff
@@ -655,11 +655,20 @@ inline static hipError_t hipMemset(void* devPtr, int value, size_t count) {
     return hipCUDAErrorTohipError(cudaMemset(devPtr, value, count));
 }
 
+inline static hipError_t hipMemsetD32(void* devPtr, int value, size_t count) {
```
mangupta: Breaks build on NVCC path. This should be hipDeviceptr_t devPtr.

include/hip/nvcc_detail/hip_runtime_api.h
```diff
@@ -655,11 +655,20 @@ inline static hipError_t hipMemset(void* devPtr, int value, size_t count) {
...
+    return hipCUDAErrorTohipError(cuMemsetD32(devPtr, value, count));
+}
+
 inline static hipError_t hipMemsetAsync(void* devPtr, int value, size_t count,
                                         hipStream_t stream __dparm(0)) {
     return hipCUDAErrorTohipError(cudaMemsetAsync(devPtr, value, count, stream));
 }
 
+inline static hipError_t hipMemsetD32Async(void* devPtr, int value, size_t count,
```
mangupta: Breaks build on NVCC path. This should be hipDeviceptr_t devPtr.

include/hip/hcc_detail/hip_runtime_api.h
whchung: @mangupta could you help clarify when `void*` be used, and when `hipDeviceptr_t` be used? Looking at the header it seems they are used somewhat interchangeably.

whchung: meanwhile I just pushed a commit to use `hipDeviceptr_t`

include/hip/nvcc_detail/hip_runtime_api.h
whchung: done

include/hip/nvcc_detail/hip_runtime_api.h
whchung: done

include/hip/hcc_detail/hip_runtime_api.h
mangupta: Depends on what is the equivalent CUDA API signature. Typically CUDA runtime APIs use void* while CUDA driver APIs use CUdeviceptr* (which gets replaced by hipDeviceptr_t in the case of HIP APIs).

gargrahul:
@whchung you may need to modify the test too to use hipDeviceptr_t . But we can wait for the jenkins test result.

whchung:
@mangupta thanks for getting this PR in!

PR #936, opened 2019-02-27
[hipconfig] Update HIP_PLATFORM detection logic

mangupta:
HIP_PLATFORM detection logic relied on finding a working KFD. If it was
found, the platform was set as hcc else as nvcc.
However this logic is flawed since it is possible for the development
system to only have the user mode bits to build HIP application code.
Hence the better logic is to rely on finding a suitable compiler.
The new logic is as follows:
- look for a working HCC. If found, platform is set as hcc.
- else look for a working NVCC. If found, platform is set as nvcc.
- else the platform defaults to hcc for now.

### DIFF
--- a/bin/hipconfig
+++ b/bin/hipconfig
@@ -59,6 +59,18 @@ sub parse_config_file {
     }
 }
 
+#---
+# Function to check if executable can be run
+sub can_run {
+    my ($exe) = @_;
+    `$exe --version 2>&1`;
+    if ($? == 0) {
+        return 1;
+    } else {
+        return 0;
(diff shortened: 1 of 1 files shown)


sunway513:
@mangupta , any ETA on merging this PR?

mangupta:
@sunway513 I was awaiting your confirmation that it resolves TF issues. Since it appears that it does fix it, I am merging it in.

sunway513:
Thanks @mangupta , the change LGTM, QA has confirmed it separately as well.

PR #943, opened 2019-02-28
Fix PyTorch build failure

mhbliao:
- Duplicated `inline` as PyTorch uses rocRand, which official release
  introducs another `inline`.
- Quick fix but needs more comprehensive solution if this `inline` is
  necessary.


mhbliao:
abortion

PR #944, opened 2019-02-28
Fix PyTorch build failure

mhbliao:
- Duplicated `inline` as PyTorch uses rocRand, which official release
  introducs another `inline`.
- Quick fix but needs more comprehensive solution if this `inline` is
  necessary.

### DIFF
--- a/include/hip/hcc_detail/host_defines.h
+++ b/include/hip/hcc_detail/host_defines.h
@@ -79,6 +79,8 @@ THE SOFTWARE.
 /**
  * Function and kernel markers
  */
+#if defined(__clang__)
+
 #define __host__
 #define __device__
 
@@ -89,6 +91,21 @@ THE SOFTWARE.
 
 #define __shared__
 #define __constant__
+
+#else
+
+#define __host__
+#define __device__
+
+#define __global__
+
+#define __noinline__
+#define __forceinline__
+
+#define __shared__
+#define __constant__
+
+#endif
 
 #endif
 


mhbliao:
@gargrahul could you review this and justify why we need `inline`? This's quick fix but would need revisiting if anything is reasonable.

iotamudelta:
@mhbliao which rocRAND release fails with this? Unless `-Werror` is set, complain should only complain about a redundant definition of inline.

`__forceinline__` should imply `inline`

mhbliao:
yeah, PyTorch build has -Werror set

iotamudelta:
Which functions is it complaining about? It would seem more prudent to correct them to use only `__forceinline__`

mhbliao:
All functions declared in, says, /opt/rocm/rocrand/include/rocrand_common.h

mhbliao:
That's header in the official release of rocRand. Even if it's fixed, the issue won't be fixed until the next release. We need to bring CI back to green.

iotamudelta:
So just remove `-Werror` from the PyTorch compilation?

mhbliao:
Why we add that change? Which app/workload requires that?

iotamudelta:
PyTorch

yxsamliu:
gcc emits an error instead of warning for duplicate inline. There is no -Werror in the first place.

mhbliao:
Could you elaborate which part of the PyTorch? What kind of gain from that change?

mhbliao:
yeah, Sam is right, duplicate 'inline' is an error instead of warning

PR #958, opened 2019-03-12
CXX11 ABI Mismatch Workaround

aaronenyeshi:


aaronenyeshi:
This patch passes all rocBLAS unit tests.

include/hip/hcc_detail/code_object_bundle.hpp
```diff
@@ -35,7 +35,6 @@ THE SOFTWARE.
 namespace hip_impl {
 
 inline
-__attribute__((visibility("hidden")))
```
mangupta: Why do we need to remove this attribute?

whchung: @scchan I would recommend we take a second thought on removing this hidden attributes. Iirc they were added by @alexvlx to make sure no symbol collisions in a ELF with multiple objects invoking HIP kernels after changes made in #929

AlexVlx: These were removed as a consequence of my advice, and they'd serve no use here. Ritualistically spamming `__attribute__((visibility("hidden")))` is something we probably want to avoid. The functions that #929 messed with were special due to the static local state, whereas these do not have it.

PR #966, opened 2019-03-15
reimplement HIP_INIT as hip_impl::hip_init(), add hip_init() to some of the inlined API

scchan:
This is to fix some HIP APIs not initializing HIP.  This causes some of the HIP unit tests to fail when HCC's lazy initialization is enabled.

### DIFF
--- a/src/hip_hcc.cpp
+++ b/src/hip_hcc.cpp
@@ -1567,7 +1567,6 @@ void ihipPrintKernelLaunch(const char* kernelName, const grid_launch_parm* lp,
 // Allows runtime to track some information about the stream.
 hipStream_t ihipPreLaunchKernel(hipStream_t stream, dim3 grid, dim3 block, grid_launch_parm* lp,
                                 const char* kernelNameStr) {
-    hip_impl::hip_init();
     stream = ihipSyncAndResolveStream(stream);
     lp->grid_dim.x = grid.x;
     lp->grid_dim.y = grid.y;
@@ -1589,7 +1588,6 @@ hipStream_t ihipPreLaunchKernel(hipStream_t stream, dim3 grid, dim3 block, grid_
 
 hipStream_t ihipPreLaunchKernel(hipStream_t stream, size_t grid, dim3 block, grid_launch_parm* lp,
                                 const char* kernelNameStr) {
-    hip_impl::hip_init();
     stream = ihipSyncAndResolveStream(stream);
     lp->grid_dim.x = grid;
     lp->grid_dim.y = 1;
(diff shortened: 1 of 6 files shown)


src/hip_hcc.cpp
```diff
@@ -1561,7 +1567,7 @@ void ihipPrintKernelLaunch(const char* kernelName, const grid_launch_parm* lp,
 // Allows runtime to track some information about the stream.
 hipStream_t ihipPreLaunchKernel(hipStream_t stream, dim3 grid, dim3 block, grid_launch_parm* lp,
                                 const char* kernelNameStr) {
-    HIP_INIT();
+    hip_impl::hip_init();
```
gargrahul: Internal functions shouldn't call hip_init(), caller of this function should have already initialized the HIP.

src/hip_hcc.cpp
```diff
@@ -1583,7 +1589,7 @@ hipStream_t ihipPreLaunchKernel(hipStream_t stream, dim3 grid, dim3 block, grid_
 
 hipStream_t ihipPreLaunchKernel(hipStream_t stream, size_t grid, dim3 block, grid_launch_parm* lp,
                                 const char* kernelNameStr) {
-    HIP_INIT();
+    hip_impl::hip_init();
```
gargrahul: Internal functions shouldn't call hip_init(), caller of this function should have already initialized the HIP.

src/hip_hcc.cpp
```diff
@@ -1604,7 +1610,7 @@ hipStream_t ihipPreLaunchKernel(hipStream_t stream, size_t grid, dim3 block, gri
 
 hipStream_t ihipPreLaunchKernel(hipStream_t stream, dim3 grid, size_t block, grid_launch_parm* lp,
                                 const char* kernelNameStr) {
-    HIP_INIT();
+    hip_impl::hip_init();
```
gargrahul: Internal functions shouldn't call hip_init(), caller of this function should have already initialized the HIP.

src/hip_hcc.cpp
```diff
@@ -1625,7 +1631,7 @@ hipStream_t ihipPreLaunchKernel(hipStream_t stream, dim3 grid, size_t block, gri
 
 hipStream_t ihipPreLaunchKernel(hipStream_t stream, size_t grid, size_t block, grid_launch_parm* lp,
                                 const char* kernelNameStr) {
-    HIP_INIT();
+    hip_impl::hip_init();
```
gargrahul: Internal functions shouldn't call hip_init(), caller of this function should have already initialized the HIP.

src/hip_hcc.cpp
scchan: removed

src/hip_hcc.cpp
scchan: removed

src/hip_hcc.cpp
scchan: removed

src/hip_hcc.cpp
scchan: removed

PR #971, opened 2019-03-19
[hip] `hipModuleGetGlobal` should be exposed.

mhbliao:
This fix the build issue in TC.


mhbliao:
better PR is prepared @ https://github.com/ROCm-Developer-Tools/HIP/pull/972

PR #976, opened 2019-03-20
look for HSA in /opt/rocm instead of /opt/rocm/hsa

ssahasra:
The directory /opt/rocm/hsa/lib does not contain a symlink named libhsa-runtime64.so, but it is present in /opt/rocm/lib. General opinion seems to be that /opt/rocm should be treated as the base for all components of ROCm, including HSA.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,7 +117,7 @@ if(HIP_PLATFORM STREQUAL "hcc")
     # Determine HSA_PATH
     if(NOT DEFINED HSA_PATH)
         if(NOT DEFINED ENV{HSA_PATH})
-            set(HSA_PATH "/opt/rocm/hsa" CACHE PATH "Path to which HSA runtime has been installed")
+            set(HSA_PATH "/opt/rocm" CACHE PATH "Path to which HSA runtime has been installed")
         else()
             set(HSA_PATH $ENV{HSA_PATH} CACHE PATH "Path to which HSA runtime has been installed")
         endif()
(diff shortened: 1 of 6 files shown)


mangupta:
@ssahasra HIP has so far allowed the user to build with HCC from a non-standard location via env var HCC_HOME. This is useful when we say want to build HIP with a differently enabled HCC (maybe debug on, etc) or we want to build HIP with a newer HCC (without replacing the one from /opt/rocm).

If we introduce this change, we cannot support building HIP using HCC from non-standard location since build system might end up picking HCC bits from "HSA_PATH".

bensander:
with frequent use of Docker containers, perhaps the *HOME and *PATH variables are no longer necessary?

mangupta:
> with frequent use of Docker containers, perhaps the *HOME and *PATH variables are no longer necessary?

I agree. Unfortunately ROCm CI does not use docker and currently relies on these variables.

ssahasra:
Uh, I would humbly submit my desire to see _less_ use of docker! I have never managed to wrap my head around that abstraction ... give me PATHs and HOMEs in the environment any day!

pfultz2:
In CMake, dependencies are found by setting `CMAKE_PREFIX_PATH`, which can be an env var or cmake build variable. Using cmake's `find_*` commands will ensure we are searching for dependencies in the correct directories(even when cross-compiling). If this is not the case it should considered a bug.

We can't hardcode the paths to dependencies. Package managers will install dependencies into local directories, and distro maintainers will want to put everything into `/usr`. 

Cmake will take care of all these scenarios if we utilize the tools it provides instead of relying on our own custom variables.

ssahasra:
+1 for that. The ideal solution is to fix the CMake files in HIP sources so that they do a package search for HCC and HSA instead of using a default path.

PR #981, opened 2019-03-21
Make hipModuleGetGlobal be in HIP runtime so it can be discovered at runtime

whchung:
In HIP PR #929, quite a few HIP public APIs were made as inline functions with
hidden visibility. It was necessary to support applications with shared
libraries with GPU kernels launched via hipLaunchKernelGGL(), after HIP runtime
is initialized.

In empirical tests, the implementation has been proved to be a bit too
excessive, especially for hipModuleGetGlobal(). The function is used by another
type of client applications which relies on the existence of this function
within HIP runtime so global symbols from HSA code objects loaded dynamically
at runtime can be retrieved programmtically.

This commit moves hipModuleGetGlobal() back to src/hip_module.cpp, and makes it
visible and not inline, to fulfill requirements for applications
aforementioned. It does not change the behavior of applications depending on
hipLaunchKernelGGL().

### DIFF
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -282,6 +282,19 @@ hipError_t hipHccModuleLaunchKernel(hipFunction_t f, uint32_t globalWorkSizeX,
         localWorkSizeZ, sharedMemBytes, hStream, kernelParams, extra, startEvent, stopEvent, 0));
 }
 
+hipError_t hipModuleGetGlobal(hipDeviceptr_t* dptr, size_t* bytes,
+                              hipModule_t hmod, const char* name) {
+    if (!dptr || !bytes) return hipErrorInvalidValue;
+
+    if (!name) return hipErrorNotInitialized;
+
+    const auto r = hmod ?
+        hip_impl::read_agent_global_from_module(dptr, bytes, hmod, name) :
+        hip_impl::read_agent_global_from_process(dptr, bytes, name);
+
+    return r;
+}
+
 namespace hip_impl {
     hsa_executable_t executable_for(hipModule_t hmod) {
         return hmod->executable;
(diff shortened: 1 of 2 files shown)


src/hip_module.cpp
```diff
@@ -282,6 +282,19 @@ hipError_t hipHccModuleLaunchKernel(hipFunction_t f, uint32_t globalWorkSizeX,
         localWorkSizeZ, sharedMemBytes, hStream, kernelParams, extra, startEvent, stopEvent, 0));
 }
 
+hipError_t hipModuleGetGlobal(hipDeviceptr_t* dptr, size_t* bytes,
+                              hipModule_t hmod, const char* name) {
+    if (!dptr || !bytes) return hipErrorInvalidValue;
```
mangupta: Since this code has been moved back to hip source, we should be calling ```HIP_INIT_API(hipModuleGetGlobal, dptr, bytes, hmod, name);``` here.

whchung: thanks. it's done.

whchung:
a gentle ping

mangupta:
Two HIP unit tests consistently fail (throw an exception) with the changes in this PR:
> directed_tests/kernel/hipTestConstant.tst
> directed_tests/deviceLib/hipTestDeviceSymbol.tst

Can you verify that these 2 tests pass you on system?

whchung:
@mangupta Let me look into it. Indeed these 2 tests are passing on `master` but failing on this branch.

whchung:
jenkins: retest this please

whchung:
@mangupta I figured out the issue with the failing unit tests. I split the implementation so `hipModuleGetGlobal` is now decoupled from other symbol-related HIP APIs. Now all tests are passing. Could you help review once again? Thanks.

whchung:
@mangupta thanks for taking this PR. Could you also help make sure the PR gets into ROCm 2.3 release branch?

PR #982, opened 2019-03-21
HACK for SWDEV-173477

whchung:
For code objects with global symbols of length 0, ROCR runtime would
ignore them even though they exist in the symbol table. Therefore the
result from read_agent_globals() can't be trusted entirely.

As a workaround to tame applications which depend on the existence of
global symbols with length 0, always return hipSuccess here.

This behavior shall be reverted once ROCR runtime has been fixed to
address SWDEV-173477


gargrahul:
LGTM, this PR reverses changes partially  made ion #721

PR #983, opened 2019-03-21
Add one test case where HIP APIs are retrieved via dlsym

whchung:
Introduce one test where HIP runtime is loaded via dlopen(), and some HIP
APIs are retrieved via dlsym(), and then 2 end-to-end tests are condutcted.

The logic is largely adopted from HIP samples "module_api_global".

It is expected HIP test framework be modified so the test can be executed with
HIP build_tests target.


whchung:
Notice this test would fail on current HIP master as `hipModuleGetGlobal` is marked as an inline function with hidden visibility, so it's not identifiable with `dlsym`, and #981 tries to address this issue.

PR #985, opened 2019-03-22
program state now obeys visibility settings

jeffdaily:


src/hip_hcc.cpp
```diff
@@ -2467,11 +2467,17 @@ hipError_t hipHccGetAcceleratorView(hipStream_t stream, hc::accelerator_view** a
 namespace hip_impl {
     std::vector<hsa_agent_t> all_hsa_agents() {
         std::vector<hsa_agent_t> r{};
-        for (auto&& acc : hc::accelerator::get_all()) {
+        auto accelerators = hc::accelerator::get_all();
```
AlexVlx: Why is this necessary? Just in order to get an indexed loop?

src/hip_hcc.cpp
```diff
@@ -2467,11 +2467,17 @@ hipError_t hipHccGetAcceleratorView(hipStream_t stream, hc::accelerator_view** a
...
+        for (int i = 0; i < accelerators.size(); i++) {
+            auto&& acc = accelerators[i];
             const auto agent = acc.get_hsa_agent();
 
             if (!agent || !acc.is_hsa_accelerator()) continue;
 
+            // If device is not in visible devices list, ignore
+            if (std::find(g_hip_visible_devices.begin(), g_hip_visible_devices.end(), (i - 1)) ==
```
AlexVlx: What happens if accelerators[0] happens to be a HSA accelerator? What's the value of i - 1 in that case?

jeffdaily: This code is duplicated from:

https://github.com/ROCm-Developer-Tools/HIP/blob/30b5c02ec48c04a4221385ad83a81afb2084b999/src/hip_hcc.cpp#L1410-L1422

Admittedly, this violates the Don't Repeat Yourself principle.  The assumption appears to be that accelerators[0] will be the CPU device and/or skipped.  If it made it as far as the std::find(), it would be looking for a visible ID of -1, which will never be in the visible device request list since it is out of range in:

https://github.com/ROCm-Developer-Tools/HIP/blob/30b5c02ec48c04a4221385ad83a81afb2084b999/src/hip_hcc.cpp#L1393-L1400

Would you suggest creating a static inline function that returns the visible device list to be used in these two places?  The logic in such a function would be updated to not assume anything, but rather it should iterate over the returned accelerators and enumerate only the GPUs, starting with an enumerated index of 0.  Something like the following?

```C++
static inline std::vector<hsa_agent_t> visible_hsa_agents() {
    std::vector<hsa_agent_t> r{};
    int accelerator_index = 0;
    for (auto&& acc : hc::accelerator::get_all()) {
        const auto agent = acc.get_hsa_agent();

        if (!agent || !acc.is_hsa_accelerator()) continue;

        // We know agent is an accelerator.
        // If device is not in visible devices list, ignore.
        // Post increment accelerator_index.
        if (std::find(g_hip_visible_devices.begin(), g_hip_visible_devices.end(), accelerator_index++) ==
                g_hip_visible_devices.end()) continue;

        r.emplace_back(*static_cast<hsa_agent_t*>(agent));
    }

    return r;
}
```

jeffdaily:
Besides the changes requested by @AlexVlx , this PR causes TensorFlow to fail if using HIP_VISIBLE_DEVICES=1 (anything non-zero) due to HIP's default stream for the default context being for device 0 regardless of users' requested visibility.  I am investigating the new issue caused by this PR and will amend ASAP.

src/hip_hcc.cpp
AlexVlx: Strictly from the standpoint of "I'd like to understand this code months / years later" the above would be highly preferable IMHO (even though it still pre-assumes that there's a bijection from index in the vector returned by get_all() and visibility info as reflected in g_hip_visible_devices, which induces implicit coupling at a distance). 

Separately, it's unclear to me if the access to g_hip_visible_devices is safe - could a situation arise in which someone is adjusting visibility in a separate thread, concurrently with this invocation of visible_hsa_agents? Or is visibility constant per process?

bensander:
The visibility env vars are read once (when hip is initialized) in a call_once routine.

jeffdaily:
The correct solution was not to have program state use visible devices, but to use the TLS current device.

@AlexVlx please review again.

@whchung please review.

AlexVlx:
This is wrong. Executables pre-populates for all GPUs in the process, once. With this change if you ever feel like using anything but the current agent it will not have a loaded executable i.e. this breaks quite a few things.

jeffdaily:
The current logic of pre-populating for all GPUs in the process is what is breaking Horovod.  I thought the code was doing a lazy init, and if the needed kernel wasn't found, it was rebuilding the in-memory kernels.  Besides using an env var, e.g., HIP_VISIBLE_DEVICES, we need to support this use case.  There must be some solution that works for all use cases.

AlexVlx:
There must be but this isn't it.

jeffdaily:
I hadn't been following development close enough.  It seems #929 got rid of the rebuild/reinit logic.  What was the reason for the rebuild logic removal?  If we are opposed to reviving it, what might be a possible alternative that would only load for 1 GPU at a time, on demand?

include/hip/hcc_detail/program_state.hpp
```diff
@@ -386,7 +386,7 @@ const std::unordered_map<
     static std::once_flag f;
 
     std::call_once(f, []() {
-        for (auto&& agent : hip_impl::all_hsa_agents()) {
+            auto agent = current_hsa_agent();
```
whchung: @jeffdaily I don't think this is correct. This lambda would only be invoked once per shared lib loaded. Using `current_hsa_agent` here would make the executables loaded only on 1 HSA agent, ignoring all other available agents, or those specified with `HIP_VISIBLE_DEVICES`.

Would it be a good idea to change `all_hsa_agents()` to `visible_hsa_agents()`, and leverage the implementation of `visible_accelerators`?

jeffdaily:
This PR is obsolete; replaced by #991.

PR #986, opened 2019-03-25
[hip] Fix typo in macro `hipLaunchKernel`

mhbliao:


mhbliao:
Need to fix HCC compilation due to the immediate arg support.

PR #990, opened 2019-03-26
SWDEV-184380 Fix hcc compilation

mhbliao:
- `hcc` has no builtin. Need to invoke LLVM intrinsic directly.

### DIFF
--- a/include/hip/hcc_detail/device_functions.h
+++ b/include/hip/hcc_detail/device_functions.h
@@ -254,14 +254,22 @@ __device__ static inline float __hip_ds_permutef(int index, float src) {
 template <int pattern>
 __device__ static inline unsigned __hip_ds_swizzle_N(unsigned int src) {
     union { int i; unsigned u; float f; } tmp; tmp.u = src;
+#if defined(__HCC__)
+    tmp.i = __llvm_amdgcn_ds_swizzle(tmp.i, pattern);
+#else
     tmp.i = __builtin_amdgcn_ds_swizzle(tmp.i, pattern);
+#endif
     return tmp.u;
 }
 
 template <int pattern>
 __device__ static inline float __hip_ds_swizzlef_N(float src) {
     union { int i; unsigned u; float f; } tmp; tmp.f = src;
+#if defined(__HCC__)
+    tmp.i = __llvm_amdgcn_ds_swizzle(tmp.i, pattern);
+#else
     tmp.i = __builtin_amdgcn_ds_swizzle(tmp.i, pattern);
+#endif
     return tmp.f;
 }
 


mhbliao:
Need to fix HCC compilation due to immediate argument support.

mhbliao:
Need to fix HCC compilation due to immediate argument support.

> The builtin to call is hc::__amdgcn_ds_swizzle

The HCC path just is reverted to its original code. I cannot find where `hc::__amdgcn_ds_swizzle` is defined. Could you elaborate more details?

> Similar issues exist for ds_bpermute

ds_bpermute is not changed in the original patch

scchan:
It's in `hc.hpp`
https://github.com/RadeonOpenCompute/hcc/blob/clang_tot_upgrade/include/hc.hpp#L3003

I mentioned `ds_bpermute` just because it has a similar pattern.  I haven't tried it to see if it works fine or now.  I'm also seeing a similar pattern with `move_dpp` so perhaps it would be more safer to update them as well.

mhbliao:
> It's in `hc.hpp`
> https://github.com/RadeonOpenCompute/hcc/blob/clang_tot_upgrade/include/hc.hpp#L3003
> 
> I mentioned `ds_bpermute` just because it has a similar pattern. I haven't tried it to see if it works fine or now. I'm also seeing a similar pattern with `move_dpp` so perhaps it would be more safer to update them as well.

In the orignal patch, only `__builtin_amdgcn_ds_swizzle` is introduced. All other device functions revised are only changed to support immediate argument restriction in LLVM. The compilation issue so far is that hcc doesn't have clang's builtin functions. Only __builtin_amdgcn_ds_swizzle needs revising to have HCC path. All other changes still invoke intrinsics using the original way by calling `__llvm_amdgcn_<XXX>`

scchan:
I was not saying that your patch introduced those other bugs and I wasn't aware of what your previous patch actually changed.  I was trying to point out that similar bugs exist and they should be fixed.

mhbliao:
> I was not saying that your patch introduced those other bugs and I wasn't aware of what your previous patch actually changed. I was trying to point out that similar bugs exist and they should be fixed.

Sorry for my words may confuse you. What my explanation is that, except `__builtin_amdgcn_ds_swizzle`, all other change doesn't change the way intrinsics are invoked, says `__llvm_amdgcn_ds_swizzle` is used in the original code and defined in HIP's header `include/hip/hcc_detail/llvm_intrinsics.h` and its definition is the same as the one in hc.hpp. As the compilation is only caused by the clang's builtins are not available in HCC. We only need to switch to the original code when being compiled by HCC to minimize the change.

pramenku:
Agree with scchan.  When HCC Clang version is 7, we are seeing below issues.
But, when HCC Clang version is 9, below issues are not observed.

So, it's good to have fix as sschan suggested. 

In file included from /home/roc-master/rocblas/build/release/Tensile/Kernels.cpp:28:
In file included from /home/roc-master/rocblas/build/release/Tensile/Kernels.h:32:
In file included from /opt/rocm/hip/include/hip/hip_runtime.h:56:
/opt/rocm/hip/include/hip/hcc_detail/hip_runtime.h:194:36: error: expected unqualified-id
__device__ float __hip_ds_swizzlef(float src, int pattern);
                                   ^
/opt/rocm/hip/include/hip/hcc_detail/hip_runtime.h:194:36: error: expected ')'
/opt/rocm/hip/include/hip/hcc_detail/hip_runtime.h:194:18: note: to match this '('
__device__ float __hip_ds_swizzlef(float src, int pattern);
                 ^
/opt/rocm/hip/include/hip/hcc_detail/device_functions.h:252:72: note: expanded from macro '__hip_ds_swizzlef'
#define __hip_ds_swizzlef(src, pattern) __hip_ds_swizzlef_N<(pattern)>((src))
                                                                       ^
In file included from /home/roc-master/rocblas/build/release/Tensile/Kernels.cpp:28:
In file included from /home/roc-master/rocblas/build/release/Tensile/Kernels.h:32:
In file included from /opt/rocm/hip/include/hip/hip_runtime.h:56:
/opt/rocm/hip/include/hip/hcc_detail/hip_runtime.h:196:44: error: expected ')'
__device__ int __hip_move_dpp(int src, int dpp_ctrl, int row_mask, int bank_mask, bool bound_ctrl);
                                           ^
/opt/rocm/hip/include/hip/hcc_detail/hip_runtime.h:196:16: note: to match this '('
__device__ int __hip_move_dpp(int src, int dpp_ctrl, int row_mask, int bank_mask, bool bound_ctrl);
               ^
/opt/rocm/hip/include/hip/hcc_detail/device_functions.h:277:20: note: expanded from macro '__hip_move_dpp'
  __hip_move_dpp_N<(dpp_ctrl), (row_mask), (bank_mask), (bound_ctrl)>((src))
                   ^
In file included from /home/roc-master/rocblas/build/release/Tensile/Kernels.cpp:28:
In file included from /home/roc-master/rocblas/build/release/Tensile/Kernels.h:32:
In file included from /opt/rocm/hip/include/hip/hip_runtime.h:56:
/opt/rocm/hip/include/hip/hcc_detail/hip_runtime.h:196:16: error: expected expression
__device__ int __hip_move_dpp(int src, int dpp_ctrl, int row_mask, int bank_mask, bool bound_ctrl);
               ^
/opt/rocm/hip/include/hip/hcc_detail/device_functions.h:277:30: note: expanded from macro '__hip_move_dpp'
  __hip_move_dpp_N<(dpp_ctrl), (row_mask), (bank_mask), (bound_ctrl)>((src))
                             ^
In file included from /home/roc-master/rocblas/build/release/Tensile/Kernels.cpp:28:
In file included from /home/roc-master/rocblas/build/release/Tensile/Kernels.h:32:
In file included from /opt/rocm/hip/include/hip/hip_runtime.h:56:
/opt/rocm/hip/include/hip/hcc_detail/hip_runtime.h:196:31: error: expected unqualified-id
__device__ int __hip_move_dpp(int src, int dpp_ctrl, int row_mask, int bank_mask, bool bound_ctrl);
                              ^
/opt/rocm/hip/include/hip/hcc_detail/hip_runtime.h:196:31: error: expected ')'
/opt/rocm/hip/include/hip/hcc_detail/hip_runtime.h:196:16: note: to match this '('
__device__ int __hip_move_dpp(int src, int dpp_ctrl, int row_mask, int bank_mask, bool bound_ctrl);
               ^

PR #994, opened 2019-03-29
Feature: launch kernel with status code

whchung:
To support client applications which requires understanding if a kernel has been dispatched, introducing `hipLaunchKernelGGLEx` in this PR, plus supplying a new test case to check this new interface.

It can be discussed whether to change existing `hipLaunchKernelGGL` to return `hipError_t`.

I add a new API so it becomes less intrusive to existing HIP clients.

### DIFF
--- a/include/hip/hcc_detail/functional_grid_launch.hpp
+++ b/include/hip/hcc_detail/functional_grid_launch.hpp
@@ -148,7 +148,7 @@ hsa_agent_t target_agent(hipStream_t stream);
 
 inline
 __attribute__((visibility("hidden")))
-void hipLaunchKernelGGLImpl(
+hipError_t hipLaunchKernelGGLImpl(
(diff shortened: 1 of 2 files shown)


gargrahul:
Thanks @whchung. You need to disable newly added test on NVCC or on HIP/NVCC do you want to map hipLaunchKernelGGlEX to cudaLaunchKernel, which returns cudaError_t ?

whchung:
@gargrahul I'd like to map `hipLaunchKernelGGLEx` to `cudaLaunchKernel`, could you show me relevant pointers?

whchung:
@gargrahul / @mangupta It seems the syntax of `cudaLaunchKernel` is a bit different than conventional `<<< >>>` and isn't easily re-mapped via a macro. I've disabled the test on HIP/NVCC path.

PR #997, opened 2019-03-30
hip-clang: fix kernel not found on multi-gpu

yxsamliu:
__hipRegisterFunction is called during by .init functions during program initialization.
It calls hipModuleGetFunction to locate kernel symbol in code objects. hipModuleGetFunction
assumes current device when locating kernel symbols. This works for HCC but not for hip-clang,
since hip-clang needs to locate kernel symbols for different devices without switching
between devices.

This patch introduces a new hsa agent parameter to ihipModuleGetFunction, which allows
__hipRegisterFunction to choose the correct hsa agent when locating kernel symbols. By
default it uses this_agent(), therefore this patch has no impact on HCC.

### DIFF
--- a/src/hip_clang.cpp
+++ b/src/hip_clang.cpp
@@ -132,11 +132,13 @@ extern "C" void __hipRegisterFunction(
   assert(modules && modules->size() >= g_deviceCnt);
   for (int deviceId = 0; deviceId < g_deviceCnt; ++deviceId) {
     hipFunction_t function;
-    if ((hipSuccess == hipModuleGetFunction(&function, modules->at(deviceId), deviceName) ||
+    hsa_agent_t agent = g_allAgents[deviceId + 1];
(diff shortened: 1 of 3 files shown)


estewart08:
I had one question about your change to hip_module.cpp on line 484:

  HIP_INIT_API(hipModuleGetFunctionEx, hfunc, hmod, name);

In order to accommodate your agent addition to the function, should that be changed to:
  HIP_INIT_API(hipModuleGetFunctionEx, hfunc, hmod, name, agent);

yxsamliu:
> I had one question about your change to hip_module.cpp on line 484:
> 
> HIP_INIT_API(hipModuleGetFunctionEx, hfunc, hmod, name);
> 
> In order to accommodate your agent addition to the function, should that be changed to:
> HIP_INIT_API(hipModuleGetFunctionEx, hfunc, hmod, name, agent);

Good point. Thanks. will fix.

PR #1004, opened 2019-04-03
Revert "HACK for SWDEV-173477"

whchung:
This reverts commit d941f19399e9b4f040aec5d41123e7073913cbc3.

Since relevant fixes in lower level component such as ROCR runtime and ld.lld has been introduced to ROCm master, filing this PR to remove this undesirable workaround in HIP.

Please do NOT merge the PR yet, as no HIP directed tests can cover this scenario just yet. I'd like to make a couple more local verification based on TensorFlow XLA first.


whchung:
lower-level components and associated TensorFlow XLA tests have been verified by CQE team. removing [DO NOT MERGE YET] tag.

whchung:
@gargrahul to my understanding the fix in ROCR would only be in ROCm 2.4, NOT ROCm 2.3. So could you make sure this PR does NOT get into release branch for 2.3?

gargrahul:
@whchung we have already branched for 2.3 and will be cherry-picking very strictly into release branch, so I believe this won't get in 2.3 release.

scchan:
2.3 hasn't been released yet so isn't it too early to revert this now?

whchung:
@scchan Yes so that's why I asked @gargrahul to ensure this patch doesn't enter ROCm 2.3. It should be ok to have it landed onto `master` as TF CI uses stock ROCm releases whenever possible (albeit sometimes we have to deviate from it), rather than HIP `mater`.

sunway513:
To clarify, we do need PR #981 to be landed in ROCm2.3 release package. 
Here's the expectations:
ROCm2.3: include HIP PR#981, not include PR#1004
ROCm2.4: include HIP PR#1004

scchan:
@whchung  there are users that use HIP  `master` on ROCm `2.2` or `2.3` so it seems that reverting this patching should be cherrypick into a future `2.4` branch of HIP and then lands in `master` with the release of ROCm `2.4`.

whchung:
@scchan I see. Then let me add `[DO NOT MERGE YET]` back. @gargrahul sorry for bring your confusion.

gargrahul:
@whchung Is it possible to guard this change with ROCr work week (or equivalent version) ?

gargrahul:
> 
> 
> To clarify, we do need PR #981 to be landed in ROCm2.3 release package.
> Here's the expectations:
> ROCm2.3: include HIP PR#981, not include PR#1004
> ROCm2.4: include HIP PR#1004

@sunway513 Regarding #981 in 2.3, @mangupta has already initiated the cherry-picking process. Let us follow up internally.

whchung:
@gargrahul Looking at ROCR CMake I don't believe it has workweek feature, and its version number seem to bump relatively slow:
https://github.com/RadeonOpenCompute/ROCR-Runtime/blob/master/src/CMakeLists.txt#L76

So I think it's not possible to do the checks you proposed based on ROCR version.

sunway513:
@gargrahul @mangupta 
We need this PR merged to roc-master ASAP. 
Could you make sure this PR would be landed on roc-master before ROCm2.4 is branched?

whchung:
Removing "[DO NOT MERGE]" prefix per request from @sunway513

PR #1006, opened 2019-04-04
Use COMgr to read Kernel Args Metadata

aaronenyeshi:
Replace the read_kernargs_metadata implementation with a COMGR implementation. This is in preparation to handling Code Object V3 as well.


aaronenyeshi:
+ @wkwchau

aaronenyeshi:
Jenkins job is failing because it cannot find location of COMGR lib and header. Is COMGR installed on the Jenkins machines? I set the default locations to check `/opt/rocm/include` and `/opt/rocm/lib`.
```
[rocm_head] CMake Warning at CMakeLists.txt:214 (MESSAGE):
[rocm_head]   Code Object Manager header not found.  Use
[rocm_head]   -DCODE_OBJ_HEADER_DIR=<path_to_amd_comgr.h>.
[rocm_head] 
[rocm_head] 
[rocm_head] CMake Error at CMakeLists.txt:224 (MESSAGE):
[rocm_head]   Code Object Manager not found.  Use
[rocm_head]   -DCODE_OBJ_MGR_DIR=<path_to_libamd_comgr.so>.
[rocm_head] 
[rocm_head] 
[rocm_head] -- Configuring incomplete, errors occurred!
```

include/hip/hcc_detail/program_state.hpp
```diff
@@ -558,92 +561,166 @@ const std::unordered_map<
...
+    status = amd_comgr_get_metadata_list_size(kernelList, &num_kernels);
+    checkError(status, "amd_comgr_get_metadata_list_size");
+    for (int i = 0; i < num_kernels; i++) {
+        status = amd_comgr_index_list_metadata(kernelList, i, &kernelMap);
+        checkError(status, "amd_comgr_index_list_metadata");
+
+        std::string kernel_name;
+        status = amd_comgr_metadata_lookup(kernelMap, "Name", &printNode);
+        checkError(status, "amd_comgr_metadata_lookup");
+        create_string_from_string_node_and_destroy(printNode, kernel_name);
```
scchan: I don't find hiding the extraction of the string into this function improves readability.  Actually, the fact that it would destroy the string node makes the code even more obscure.

scchan:
@aaronenyeshi I think you'll have to update docker script for the HIP CI to install the code object manager package

aaronenyeshi:
> @aaronenyeshi I think you'll have to update docker script for the HIP CI to install the code object manager package

Do you know which repo the Dockerfile is located?

include/hip/hcc_detail/program_state.hpp
aaronenyeshi: Do you suggest to put the contents of `create_string_from_string_node_and_destroy` into each instance that I grab a string?

scchan: or move the construction of the `std::string` and the metadata lookup into that this helper function (and rename it)?

mangupta:
> > @aaronenyeshi I think you'll have to update docker script for the HIP CI to install the code object manager package
> 
> Do you know which repo the Dockerfile is located?

I have fixed the rocm-head Docker image to install code object manager. However this PR breaks build on NVCC path as well since we are explictly requiring code object manager there as well. You need to guard the cmake changes against HIP_PLATFORM=nvcc.

CMakeLists.txt
```diff
@@ -193,6 +193,49 @@ if(USE_PROF_API EQUAL 1)
...
+        PATHS
+            /opt/rocm/lib
+    )
+    if (NOT CODE_OBJ_MGR)
+        MESSAGE(FATAL_ERROR "Code Object Manager not found. Use -DCODE_OBJ_MGR_DIR=<path_to_libamd_comgr.so>.")
+    else()
+        include_directories(${CODE_OBJ_HEADER})
+        add_library(amd_comgr SHARED IMPORTED GLOBAL)
+        set_target_properties(amd_comgr PROPERTIES
+            INTERFACE_INCLUDE_DIRECTORIES "${CODE_OBJ_HEADER}"
```
scott-linder: Comgr should already set this in amd_comgr-config.cmake; I don't think most of the changes here are required, as just `find_library` with an optional hint should be enough. If that doesn't work, I would consider it a bug in Comgr.

include/hip/hcc_detail/program_state.hpp
```diff
@@ -558,92 +561,156 @@ const std::unordered_map<
...
+    size_t num_kern_args = 0;
+
+    // Kernels is a list of MAPS!!
+    status = amd_comgr_metadata_lookup(blob_meta, "Kernels", &kernelList);
+    if(status != AMD_COMGR_STATUS_SUCCESS)
+        return;
+
+    status = amd_comgr_get_metadata_kind(kernelList, &mkindLookup);
+    if (mkindLookup != AMD_COMGR_METADATA_KIND_LIST) {
+        hip_throw(std::runtime_error{"Lookup of Kernels didn't return a list\n"});
```
scott-linder: Is there no way to recover from this throw? If there is, `kernelList` will leak here. Similar question for the other throws below.

include/hip/hcc_detail/program_state.hpp
```diff
@@ -558,92 +561,156 @@ const std::unordered_map<
...
+    amd_comgr_metadata_node_t kernelList, kernelMap;
+    amd_comgr_metadata_node_t kernArgList, kernArgMap;
+    std::string kernel_name;
+    size_t num_kernels = 0;
+    size_t num_kern_args = 0;
+
+    // Kernels is a list of MAPS!!
+    status = amd_comgr_metadata_lookup(blob_meta, "Kernels", &kernelList);
+    if(status != AMD_COMGR_STATUS_SUCCESS)
+        return;
```
scott-linder: Why `return` here, rather than throw? The "Kernels" key is required (see https://llvm.org/docs/AMDGPUUsage.html#amdgpu-amdhsa-code-object-metadata-map-table-v2).

include/hip/hcc_detail/program_state.hpp
aaronenyeshi: I could modify hip_throw to clean up all nodes that are created. When I pass a declared (but not used) node into amd_comgr_destroy_metadata, it will seg fault. Would it be good to keep a vector of created nodes, and destroy each of them on hip_throw?

include/hip/hcc_detail/program_state.hpp
aaronenyeshi: There seem to be some HIP memset kernels which don't have a "Kernels" key. So if I throw here, all those kernels that are missing this field will fail to run. @mangupta, I believe that makes them invalid kernels, should we create a different PR to fix those kernels?

include/hip/hcc_detail/program_state.hpp
scott-linder: Tracking each initialized object independently (either via a vector as you mention, or by wrapping them in a simple class containing the handle and a boolean) is the only way to mange this without the potential to leak. The Comgr API does not allow for inspecting/modifying the opaque handle itself, so it really has to be separate.

include/hip/hcc_detail/program_state.hpp
scott-linder: I would imagine that is a bug somewhere else, and allowing it here will probably just be papering over that? I'm not sure what other consumers do currently.

include/hip/hcc_detail/program_state.hpp
aaronenyeshi: I've changed the implementation to use simple class as you've suggested, and the destructor will destroy on hip_throw and leaving of scope.

CMakeLists.txt
aaronenyeshi: I've tried to do that, but ran into some issue. I believe that should be dealt with in a separate PR.

include/hip/hcc_detail/program_state.hpp
aaronenyeshi: I've added a FIXME for the HIP team to fix those kernels in the future. It should be fixed in a separate PR since its out of this PR's context. Thanks!

CMakeLists.txt
aaronenyeshi: Hi Scott, I've updated the CMAKE to use the find_package instead

scott-linder: LGTM, thank!

include/hip/hcc_detail/program_state.hpp
```diff
@@ -558,92 +561,185 @@ const std::unordered_map<
...
+
+    amd_comgr_status_t status;
+    comgr_data blob_data;
+    status = amd_comgr_create_data(AMD_COMGR_DATA_KIND_RELOCATABLE, &blob_data.data);
+    checkError(status, "amd_comgr_create_data");
+
+    status = amd_comgr_set_data(blob_data.data, blob_size, blob_buf);
+    if(status != AMD_COMGR_STATUS_SUCCESS)
+        return;
+    blob_data.set_active(true);
```
scott-linder: I think this `set_active` should be moved before the call to `set_data`; a data object will leak if it can be created but cannot be set. I don't see any other issues with the use of comgr in the review.

aaronenyeshi: Thanks for catching that!

PR #1009, opened 2019-04-07
Fix regression on multi-gpu due to PR#997

yxsamliu:
PR#997 caused memory access fault for HCC due to hipMemcpy to a device in a different root complex when running PyTorch mGPU.

This patch restores behavior for HCC as before PR#997 therefore fixes the regression.


gargrahul:
@yxsamliu To be clear, this PR resolves side effects caused by #997, which caused SWDEV-185566- Illegal instruction error while running TensorFlow,caffe2,rocblas,hipblas,rocprim,MIOpen Tests.

This PR doesn't resolve, SWDEV-183320- memory access fault. SWDEV-183320 was resolved by different PRs as mentioned in the tickets comments.

PR #1017, opened 2019-04-11
add little changes in hip_runtime_api.h to work with c language

Mr-LiuSw:
Update /hcc_detail/hip_runtime_api.h

when i try to use mpicc or gcc to compile a c language code which call some hip runtime api , error occurred as
> /path/to/hcc_detail/hip_runtime_api.h:2268:33: error: unknown type name ‘hipFuncAttributes’; 
> hipFuncGetAttributes(hipFuncAttributes* attr, const void* func);
 
add ' struct ' for the first parameter of hipFuncGetAttributes will get ride of this problem.


Mr-LiuSw:
i have fix the problem. now it can be build successfully

PR #1018, opened 2019-04-11
In hipFree, synchronize owner of memory

jeffdaily:
This changes the behavior from synchronizing the currently set TLS device.

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -1897,19 +1897,29 @@ hipError_t hipFree(void* ptr) {
         am_status_t status = hc::am_memtracker_getinfo(&amPointerInfo, ptr);
         if (status == AM_SUCCESS) {
             if (amPointerInfo._hostPointer == NULL) {
-                ihipCtx_t* ctx;
-                if (amPointerInfo._appId != -1) {
-#if USE_APP_PTR_FOR_CTX
-                    ctx = static_cast<ihipCtx_t*>(amPointerInfo._appPtr);
-#else
-                    ctx = ihipGetPrimaryCtx(amPointerInfo._appId);
-#endif
-                } else {
-                    ctx = ihipGetTlsDefaultCtx();
+                if (HIP_SYNC_FREE) {
+                    // Synchronize all devices, all streams
+                    // to ensure all work has finished on all devices.
+                    // This is disabled by default.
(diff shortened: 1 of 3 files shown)


gargrahul:
@jeffdaily I think we need to have an all devices level sync in hipFree. It is more critical for _appId=-1 case.

src/hip_memory.cpp
```diff
@@ -1901,6 +1897,22 @@ hipError_t hipFree(void* ptr) {
         am_status_t status = hc::am_memtracker_getinfo(&amPointerInfo, ptr);
         if (status == AM_SUCCESS) {
             if (amPointerInfo._hostPointer == NULL) {
+                if (amPointerInfo._appId != -1) {
```
gargrahul: Do we still need this check? Since we are synchronizing all devices for now.

gargrahul:
@jeffdaily Shall we synchronize all devices always in hipFree for now and optimize to synchronize only peer devices later separately?

src/hip_memory.cpp
jeffdaily: I thought this was differentiating between the case where the pointer is owned and not shared versus when it is owned but shared.  Perhaps I misunderstood HCC pointer management and what the _appId field is used for.  I don't want hipFree() to become a performance bottleneck if we can help it.

gargrahul: _appId= -1 is used for host allocations. I don't think hipFree should be performance critical code.

gargrahul: Synchronizing all devices may help avoiding future mGPU related issues when devices are sharing pointers and freeing across devices. Does blanket all device synchronization resolve 183319?

jeffdaily:
For now, we should synchronize the owning device.  The all-device-all-stream sync will also solve the current problem, but is too extreme of a solution at this time.  However, to help debug possible mGPU issues in the future, I have added the environment variable HIP_SYNC_FREE to enable the all-device-all-stream sync.

bensander:
Thanks Jeff!

gargrahul:
LGTM. We may need to revisit this when we see a real use case where HIP_SYNC_FREE helps to make all/peer devices sync as default.

gargrahul:
@mangupta This PR needs to get merged today as this is required for ROCm 2.4 release.

PR #1032, opened 2019-04-18
Fix for __ffsll() device functions.

kpyzhov:
Use __builtin_ctzll() intrinsic in the __ffsll() implementation instead of __builtin_ctzl().


kpyzhov:
There are other changes needed to this file. I'll upload a new request later today.

PR #1033, opened 2019-04-19
introduce hostcall

ssahasra:
This an early proof of concept for introducing hostcall in HIP.

The change introduces a global hostcall state as follows:

1. Launch a consumer thread in hip_init.
2. Maintain a list of mappings from HSA queue to hostcall buffers.
3. Retrieve or create a hostcall buffer for the HSA queue at the time of launching a kernel. This requires locking the HIP stream so that the HSA queue is not stolen until the kernel is launched.
4. The destructor for the global state is responsible for terminating the consumer thread.

The change also includes a temporary hack that assigns the hostcall buffer poiner to the first explicit argument of the kernel. This will be replaced by an implicit kernel argument with suitable changes in the HIP compiler.

A working testcase can be seen here:
http://gitlab1.amd.com/ssahasra/test-hostcall

It combines the following components:
1. The hostcall [device library](http://git.amd.com:8080/c/lightning/ec/device-libs/+/212864). 
2. The hostcall [host library](http://git.amd.com:8080/c/lightning/ec/support/+/216834).
3. This POC change to the HIP runtime.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,12 @@
 cmake_minimum_required(VERSION 3.4.3)
 project(hip)
 
+# Threads are are required by the hostcall tests. The check uses the
+# system compiler, but it is affected by CMAKE_CXX_FLAGS. Later parts
+# of this file add flags like "-hc", which cause the system compiler
+# to error out.
+find_package(Threads)
+
 #############################
 # Options
 #############################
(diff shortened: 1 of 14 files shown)


ssahasra:
New more stable version that can handle a hipDeviceReset() by discarding all hostcall buffers on the device that got reset.

ssahasra:
Closing this draft, to be replaced by an official PR

PR #1046, opened 2019-04-23
Add COMGR relative path for build machines

aaronenyeshi:
This should fix the jenkins build machines.


aaronenyeshi:
I've tested that this works in finding a default COMGR in `/opt/rocm`, and also works when you set `CMAKE_PREFIX_PATH` to a directory such as `out/ubuntu-16.04/16.04/` and it will find it at `out/ubuntu-16.04/16.04/lib/cmake/amd_comgr` . Thanks

PR #1052, opened 2019-04-25
Partial clean to avoid top-level apps bypass HIP

ghostplant:
PR https://github.com/ROCm-Developer-Tools/HIP/pull/929 introduces some library-link pollution problems from applications that are supported to call HIP only (happens since ROCm 2.3.x). This is 1 out of 2 steps to get over such pollution.

Signed-off-by: CUI Wei <ghostplant@qq.com>

### DIFF
--- a/include/hip/hcc_detail/code_object_bundle.hpp
+++ b/include/hip/hcc_detail/code_object_bundle.hpp
@@ -34,48 +34,6 @@ THE SOFTWARE.
 
 namespace hip_impl {
 
-inline
-std::string transmogrify_triple(const std::string& triple)
-{
-    static constexpr const char old_prefix[]{"hcc-amdgcn--amdhsa-gfx"};
-    static constexpr const char new_prefix[]{"hcc-amdgcn-amd-amdhsa--gfx"};
-
-    if (triple.find(old_prefix) == 0) {
-        return new_prefix + triple.substr(sizeof(old_prefix) - 1);
-    }
-
-    return (triple.find(new_prefix) == 0) ? triple : "";
-}
-
-inline
-std::string isa_name(std::string triple)
-{
-    static constexpr const char offload_prefix[]{"hcc-"};
-
-    triple = transmogrify_triple(triple);
-    if (triple.empty()) return {};
-
-    triple.erase(0, sizeof(offload_prefix) - 1);
-
-    return triple;
-}
-
-inline
-hsa_isa_t triple_to_hsa_isa(const std::string& triple) {
-    const std::string isa{isa_name(std::move(triple))};
-
-    if (isa.empty()) return hsa_isa_t({});
-
-    hsa_isa_t r{};
-
-    if(HSA_STATUS_SUCCESS != hsa_isa_from_name(isa.c_str(), &r)) {
-        r.handle = 0;
-    }
-
-    return r;
-}
-
-
 struct Bundled_code {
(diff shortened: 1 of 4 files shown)


include/hip/hcc_detail/functional_grid_launch.hpp
```diff
@@ -160,10 +151,15 @@ void hipLaunchKernelGGLImpl(
     auto it = functions(agent).find(function_address);
 
     if (it == functions(agent).cend()) {
+        int devId{0};
+        char devName[64]{};
+        hipGetDevice(&devId);
```
mangupta: Having HIP APIs invoking other HIP APIs is something that needs to be avoided. There are some instances of it happening in the HIP runtime which need to be cleaned up. But adding more mess is not the way to go.

ghostplant: Thanks, I'll do a small change then.

ghostplant:
@mangupta The change is done. I have already cleaned **ALL** in-proper calls from top-level apps to HSA-runtime, only expect for user-include-file `program_state.hpp` which contains quantities of low-level calls bypassing HIP libraries.

In order to fully cleaning such low-level calls, I suggest to create another PR apart from this one to mainly manage on the cleaning of that header.

include/hip/hcc_detail/functional_grid_launch.hpp
```diff
@@ -135,16 +135,8 @@ std::string name(std::uintptr_t function_address)
     return it->second;
 }
 
-inline
-std::string name(hsa_agent_t agent)
-{
-    char n[64]{};
-    hsa_agent_get_info(agent, HSA_AGENT_INFO_NAME, n);
-
```
ghostplant: Get over this low-level call from headers file included by top-level apps.

include/hip/hcc_detail/code_object_bundle.hpp
```diff
@@ -34,48 +34,6 @@ THE SOFTWARE.
...
-
-    if(HSA_STATUS_SUCCESS != hsa_isa_from_name(isa.c_str(), &r)) {
-        r.handle = 0;
-    }
-
-    return r;
-}
-
-
 struct Bundled_code {
```
ghostplant: Moving them to `hsa_helpers.hpp` which should be transparent to HIP apps.

scchan:
@ghostplant  Thanks for patch.  I've been working on something similar where I've moved most of the logic introduced by PR #929 into the HIP runtime shared library so I think that would address the problem you are trying to solve here.  My changes are almost already so at this point I'd prefer not merging this PR.  If you are interested, you can get a snapshot of my stuffs here: https://github.com/scchan/HIP/tree/lazy_binding-merged_program_state

ghostplant:
@scchan Thanks, I tried that, but it is likely to conflict with https://github.com/ROCm-Developer-Tools/HIP/pull/1057/files, as PR 1057 brings back hsa-runtime calls again...

ghostplant:
@scchan I'll close this PR then as yours is more completed.

scchan:
@ghostplant   @jeffdaily already gave me a heads-up and I'll work with him to adapt his patch into my change set

ghostplant:
@scchan OK. BTW, can you take a look at this PR: https://github.com/scchan/HIP/pull/1

PR #1056, opened 2019-04-25
Revert COMGR patches

aaronenyeshi:
Here is a back-up plan to revert COMGR patches, until we fix the COMGR installation issue.


aaronenyeshi:
Please do not merge until unless there is a Green light from Jenkins tests.

PR #1057, opened 2019-04-25
cache executables based on elf name, isa, and agent.

jeffdaily:
This avoids program state reloading executables after a shared library is dlopened.

### DIFF
--- a/include/hip/hcc_detail/program_state.hpp
+++ b/include/hip/hcc_detail/program_state.hpp
@@ -85,7 +85,7 @@ inline constexpr bool operator==(hsa_isa_t x, hsa_isa_t y) {
 namespace hip_impl {
 
 std::vector<hsa_agent_t> all_hsa_agents();
-void executables_cache(std::string, hsa_isa_t, hsa_agent_t, std::vector<hsa_executable_t>&, bool);
+std::vector<hsa_executable_t>& executables_cache(std::string, hsa_isa_t, hsa_agent_t);
 extern std::mutex executables_cache_mutex;
 
 class Kernel_descriptor {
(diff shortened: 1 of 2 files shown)


include/hip/hcc_detail/program_state.hpp
```diff
@@ -399,29 +397,46 @@ const std::vector<hsa_executable_t>& executables(hsa_agent_t agent) {
...
+                    // illustration.
+                    std::string blob_to_str{blob.cbegin(), blob.cend()};
+                    tmp = load_executable(blob_to_str, tmp, a);
+
+                    if (tmp.handle) {
+                        current_exes.push_back(tmp);
+                    }
+                }
+                // cache the newly loaded executables
+                hip_impl::executables_cache(elf, x, a, current_exes, true/*write, not read*/);
```
scchan: I think the retrieval of the cached map, the load of a new executable and the cache update should be all in a critical section protected by a global mutex (since the cache is global).  I don't think the call_once flag is sufficient here to protect against another thread running the same code but for a different module.

jeffdaily:
@scchan I applied the same requested fix to https://github.com/scchan/HIP/pull/2, as well.

src/hip_hcc.cpp
```diff
@@ -2521,4 +2521,19 @@ namespace hip_impl {
             std::terminate();
         #endif
     }
+
+    std::mutex executables_cache_mutex;
+
+    void executables_cache(
```
scchan: Since the lock is have moved outside of this function, I think it would be simpler and efficient to just return a ref to the std::vector<hsa_executable_t>?  That way, we don't need to make a new copy on read and we won't actually need a write operation.

sunway513:
@scchan @mangupta please aware we need this fix to be in ROCm2.5, thanks.

scchan:
This PR should be merged since it has been superseded by #1077

sunway513:
Thanks @scchan . 
@jeffdaily , could you help confirm?

jeffdaily:
This PR might be causing TensorFlow fused ops to fail.  I will investigate further.

jeffdaily:
@scchan perhaps I misunderstood your earlier comment.  If this PR has been superseded by #1077, should it be closed?

scchan:
@jeffdaily yes, this PR should be closed.  There was a typo in my earlier comment (fixed) sorry for the confusion.

gargrahul:
@scchan Could we retain this PR and modify #1077 ? QA team has already done basic testing with this PR as we wanted to avoid possible regressions.

This PR is must for 2.5.

scchan:
@gargrahul #1077 is a must have for the next release as well

jeffdaily:
superseded by #1077

PR #1058, opened 2019-04-26
[Device Function] Fix implementation

mhbliao:
- It's a common mistake by assuming 1 << shamt would be promoted to
  64-bit, if shamt is a 64-bit integer. That's not the case. Replace
  that left shift to a 64-bit one to ensure it won't fall into undefined
  behavior.
- Fix the host-side implementation as well for device function testing.


mhbliao:
@yxsamliu could you review this?

aaronenyeshi:
Thank you Michael, does hip_bitextract need any changes?

mhbliao:
> Thank you Michael, does hip_bitextract need any changes?
shift's LHS(s) in hip_bitextract are variables in 64-bit integer. They are OK.

PR #1062, opened 2019-04-29
[hip] Re-implement ballot using AMDGCN builtins

mhbliao:
- As the signature of `amdgcn.icmp` is changed for next-gen hip, using
  clang builtins is portable way to hide that details.


mhbliao:
s/hip/chip/g for the comment

PR #1064, opened 2019-04-30
Return hipErrorInsufficientDriver status when CPU device not found

wkwchau:
This change is to avoid HIP runtime causing TF ihipexception error when there's no access to amdgpu driver.

### DIFF
--- a/src/hip_hcc.cpp
+++ b/src/hip_hcc.cpp
@@ -130,6 +130,7 @@ std::vector<int> g_hip_visible_devices;
 hsa_agent_t g_cpu_agent;
 hsa_agent_t* g_allAgents;  // CPU agents + all the visible GPU agents.
 unsigned g_numLogicalThreads;
+bool g_initDeviceFound = false;
 
 std::atomic<int> g_lastShortTid(1);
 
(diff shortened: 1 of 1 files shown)


src/hip_hcc.cpp
```diff
@@ -1405,9 +1406,11 @@ void ihipInit() {
     hsa_status_t err = hsa_iterate_agents(findCpuAgent, &g_cpu_agent);
     if (err != HSA_STATUS_INFO_BREAK) {
         // didn't find a CPU.
-        throw ihipException(hipErrorRuntimeOther);
+        g_initDeviceFound = false;
+        return;
     }
 
+    g_initDeviceFound = true;
```
mangupta: We should probably set this just before returning from ihipInit(). This way we won't have to move this around in case ihipInit() changes and we have another point where an error can be thrown.

wkwchau: set g_initDeviceFound to true at the end of ihipInit()

sunway513:
Thanks @wkwchau @mangupta , will this change to be in ROCm2.5?

mangupta:
Yes it will make it to ROCm 2.5.

PR #1067, opened 2019-05-02
Implement COMGR V2 support with cmake -DUSE_COMGR=1 guard

aaronenyeshi:
Re-add the previous COMGR changes, but use a macro guard enabled by cmake option -DUSE_COMGR=1. 

We can work on fixing PSDB with this guard.


aaronenyeshi:
@wkwchau I fixed the bug we were seeing with hipcc built binaries not properly defining the macro USE_COMGR as 1.

aaronenyeshi:
@wkwchau has resolved the conflicts in his PR https://github.com/ROCm-Developer-Tools/HIP/pull/1100

PR #1070, opened 2019-05-02
[test] Replace variadic function usage to alternative ones.

mhbliao:


mhbliao:
there are too much inconsistency between HCC and HIP-Clang

PR #1071, opened 2019-05-03
Runtime changes to allocate implicit kernel arguments.

cdevadas:
Appended the implicit kernel argument buffer in the argument area. This buffer (48-bytes) appears at the end of the user argument area. The hostcall services can utilize this area. 
This change should go into the runtime mainline once the compiler enables the implicit kernel argument for HIP.

### DIFF
--- a/src/hip_clang.cpp
+++ b/src/hip_clang.cpp
@@ -233,6 +233,9 @@ hipError_t hipLaunchByPtr(const void *hostFunction)
         " for device %d!\n", hostFunction, deviceId);
     abort();
   } else {
+    // Insert 48-bytes at the end for implicit kernel arguments and fill with value zero.
+    size_t padSize = (~exec._arguments.size() + 1) & (HIP_KERNARG_ALIGNMENT - 1);
+    exec._arguments.insert(exec._arguments.end(), padSize + HIP_IMPLICIT_KERNARG_SIZE, 0);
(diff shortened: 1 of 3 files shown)


mangupta:
@cdevadas Is it intended that this change is only going to apply when using hip-clang compiler? or does it also need to apply when using hcc?

ssahasra:
@mangupta that's the main reason why this PR is still a draft. The current change will work only with hip-clang, and we should add changes that work with hcc. Also, we should have tests making sure the change works as expected in both cases.

src/hip_clang.cpp
```diff
@@ -232,7 +232,10 @@ hipError_t hipLaunchByPtr(const void *hostFunction)
     abort();
   } else {
     // Insert 48-bytes at the end for implicit kernel arguments and fill with value zero.
-    exec._arguments.insert(exec._arguments.end(), 48, 0);
+    // First align the argument layout to the word boundary.
```
ssahasra: "align the argument layout" seems misleading to me. You are not really aligning the existing arguments, right? Simply computing the number of padding bytes to be inserted before the new arguments. Also, "pad_size" may be a better name than just "pad", since it is not the pad itself but the number of bytes to be padded.

cdevadas:
closing this pull request

PR #1077, opened 2019-05-03
migrate program_state logic from header into shared library (phase I)

scchan:
This is an ongoing effort to migrate device code management logic introduced in PR#929 from header files into the HIP runtime shared library while maintaining the same functionality.  This would reduce a significant amount of symbols in the application objects and minimize exposure of HIP's internal data structures to hazardous environments.

This PR includes functionalities in #1057 adapted by @jeffdaily 
This PR should be merged before #1067 , which needs to be updated to adapted to this PR and will be submitted in a future PR.


scchan:
@jeffdaily  could you review?  I wasn't able to add you as a reviewer...

include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -2629,90 +2630,129 @@ std::vector<Agent_global> read_agent_globals(hsa_agent_t agent,
...
-    }
 
-    const auto it0 = agent_globals.find(key);
-    if (it0 == agent_globals.cend()) {
-        hip_throw(
-            std::runtime_error{"agent_globals data structure corrupted."});
+        std::tie(*dptr, *bytes) = read_global_description(it0->second.cbegin(),
+                it0->second.cend(), name);
+
+        // HACK for SWDEV-173477
```
jeffdaily: What is the status of this hack?  Is it still needed?  Is the code here representative of the hack or the commented-out reversal of the hack?

scchan: The hack is still in the master branch.  @whchung could confirm but I think it would be safe to remove it now since this is not needed in 2.4.  We could take this out in a different PR

whchung: we don't need the hack now.

PR #1097, opened 2019-05-10
Add HIPRTC, glorious ersatz for NVRTC

AlexVlx:


src/hiprtc.cpp
```diff
@@ -0,0 +1,601 @@
...
+hiprtcResult hiprtcCompileProgram(hiprtcProgram p, int n, const char** o)
+{
+    using namespace std;
+
+    if (n && !o) return HIPRTC_ERROR_INVALID_INPUT;
+    if (!isValidProgram(p)) return HIPRTC_ERROR_INVALID_PROGRAM;
+    if (p->compiled) return HIPRTC_ERROR_COMPILATION;
+
+    static const string hipcc{
+        getenv("HIP_PATH") ? (getenv("HIP_PATH") + string{"/bin/hipcc"})
```
mangupta: We should probably look for hipcc relative to libhiprtc.so. This is because HIP unit tests don't set env var HIP_PATH and we usually don't install to /opt/rocm when building unit tests. So these tests will fail to find the right hipcc.
Maybe we can do something like (sanity checks/error handling not included):
```cpp
#include <dlfcn.h>
Dl_info dl_info;
dladdr((void*)hiprtcCompileProgram, &dl_info);
string path(dl_info.dli_fname); // path = /path/to/hip/lib/libhiprtc.so
path = path.substr(0, path.find_last_of("/")).substr(0, path.find_last_of("/")); // path = /path/to/hip
static const string hipcc{ path + string{"/bin/hipcc"}};
```

AlexVlx: I am against this. HIPRTC should just work with any `hipcc`, there is no such thing as the right `hipcc`. Moreover, the user could decide to interpose a flavour of `libhiprtc.so` that lives wherever he / she chooses, and then bad things happen. The test itself should be changed to set HIP_PATH, this alternative is more like the tail wagging the dog.

mangupta: Technically you are right. However here is the issue. HIP runtime has a close relationship with the HCC version. To ensure we are using the right HCC, hipcc has a strict check to confirm that the HCC version that was used to build HIP runtime and the HCC version that is currently being used is the same. So hipcc in another install location could be expecting a different HCC version than the hipcc in the path relative to the current libhiprtc.so. When that happens, hipcc will throw a nice little warning about mismatching HCC versions and abort compilation. So the test will fail.
As regards setting HIP_PATH, https://cmake.org/cmake/help/latest/command/add_test.html does not allow that. So I need to set the path manually *before* I build HIP itself. I don't think that is right either.

AlexVlx: Yeah that's nice but irrelevant, since the HIPRTC thing is supposed to work with any hipcc, and the latter either works with the environmental hcc or doesn't. It doesn't expect anything in that regard. 

Here's the basic idea, since for some reason it was not crisp enough initially: the .so could be anywhere in the path, with no obvious relationship to the placement of `hipcc`, and that's a valid use case. You cannot rely on its placement to mean anything, and you should not enshrine any implied expectations by rapport with its meaning.

Finally, just for the sake of argument, it is trivial to set the environment for a test, via `set_tests_properties`  + `ENVIRONMENT`, see https://cmake.org/cmake/help/latest/command/set_tests_properties.html and https://cmake.org/cmake/help/latest/prop_test/ENVIRONMENT.html.

PR #1114, opened 2019-05-16
Runtime changes to append implicit kernel arguments.

cdevadas:
48 bytes of empty bytes are appended to the kernarg area at runtime. The implicit arguments are enabled primarily for the hostcall services
and it is completely abstracted from the user code. Enabled it for both hip-clang and hip-hcc.


src/hip_clang.cpp
```diff
@@ -233,6 +233,9 @@ hipError_t hipLaunchByPtr(const void *hostFunction)
         " for device %d!\n", hostFunction, deviceId);
     abort();
   } else {
+    // Insert 48-bytes at the end for implicit kernel arguments and fill with value zero.
```
yxsamliu: kernels may be launched by hipModuleLaunchKernel. This place is not generic enough to catch all kernel launching in hip-clang. You may want to make changes to ihipModuleLaunchKernel to handle implicit kernel arg for all kernel launches.

include/hip/hip_hcc.h
```diff
@@ -27,6 +27,10 @@ THE SOFTWARE.
 
 #include "hip/hip_runtime_api.h"
 
+// HIP implicit kernargs.
+#define HIP_IMPLICIT_KERNARG_SIZE 48
+#define HIP_KERNARG_ALIGNMENT 8
```
yxsamliu: better not introduce this macro since it causes confusion. You align up the offset of the first implicit arg because it has alignment of 8 itself, not because there is generic alignment requirement for kernel args.

include/hip/hcc_detail/functional_grid_launch.hpp
```diff
@@ -108,10 +108,14 @@ inline std::vector<std::uint8_t> make_kernarg(
...
 
     auto& ps = hip_impl::get_program_state();
-    return make_kernarg<0>(to_formals, 
+    auto kernargs =  make_kernarg<0>(to_formals,
                            ps.get_kernargs_size_align(
                                reinterpret_cast<std::uintptr_t>(kernel)),
                            std::move(kernarg));
+    // Insert 48-bytes at the end for implicit kernel arguments and fill with value zero.
+    size_t padSize = (~kernargs.size() + 1) & (HIP_KERNARG_ALIGNMENT - 1);
+    kernargs.insert(kernargs.end(), padSize + HIP_IMPLICIT_KERNARG_SIZE, 0);
```
scchan: Can we add this into kernargs_size_align so that we have all the logic for kernarg size calculation in one place?

cdevadas:
Moved the code to ihipModuleLaunchKernel. It adds the implicit kernargs before launching the kernel. This function is common for both HIP-Clang and HIP-HCC and hence it handles all the paths.

src/hip_module.cpp
```diff
@@ -230,7 +229,11 @@ hipError_t ihipModuleLaunchKernel(hipFunction_t f, uint32_t globalWorkSizeX,
 
         hc::completion_future cf;
 
-        lp.av->dispatch_hsa_kernel(&aql, config[1] /* kernarg*/, kernArgSize,
+        // Insert 48-bytes at the end for implicit kernel arguments and fill with value zero.
```
ssahasra: We should move these three lines earlier, closer to where the kernargs vector is being populated. The code just above this line is not related to kernargs, and it is cleaner to have all kernarg related bunched together.

cdevadas: It is done.

cdevadas:
> hipModuleLaunchKernel is not considered. this PR would likely cause all ML workloads to fail

It is done.

include/hip/hcc_detail/functional_grid_launch.hpp
cdevadas: Unified the fix into a single file and hence included the static const declarations at the top of the file. There is no need to define the constants/macros in a header.

PR #1117, opened 2019-05-16
Fix hipcc exit code when failing

aaronenyeshi:


aaronenyeshi:
This patch deals with the issue where failed compilations returned a success (exit code 0). Instead we should forward what the compiler returns.

psychocoderHPC:
@mangupta can this PR somehow find the way into the next ROCm release. The wrong return code makes developer the life very hard because CMake is not seeing errors and the ROCm linker is very relaxed and creates binaries even if the device side compile fails.

mangupta:
@psychocoderHPC Yes this should be in the upcoming ROCm 2.6 release.

PR #1130, opened 2019-05-23
Add support for code object v3

lmoriche:
Use the code object manager library to parse the code object metadata. Both
code object v2 and v3 formats are now supported for HCC generated binaries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -271,6 +271,21 @@ if(HIP_PLATFORM STREQUAL "hcc")
         target_link_libraries(hiprtc PUBLIC stdc++fs)
     endif()
 
+
+    if(HIP_PLATFORM STREQUAL "hcc")
+        find_package(amd_comgr REQUIRED CONFIG
+            PATHS
+                /opt/rocm/
+            PATH_SUFFIXES
+                cmake/amd_comgr
+                lib/cmake/amd_comgr
+        )
+        MESSAGE(STATUS "Code Object Manager found at ${amd_comgr_DIR}.")
+    endif()
+
+    target_link_libraries(hip_hcc PRIVATE amd_comgr)
+    target_link_libraries(hip_hcc_static PRIVATE amd_comgr)
+
     string(REPLACE " " ";" HCC_CXX_FLAGS_LIST ${HCC_CXX_FLAGS})
     foreach(TARGET hip_hcc hip_hcc_static)
         target_include_directories(${TARGET} SYSTEM INTERFACE $<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>;${HSA_PATH}/include)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -270,6 +270,8 @@ if(HIP_PLATFORM STREQUAL "hcc")
                 PRIVATE ${PROJECT_SOURCE_DIR}/include ${HSA_PATH}/include)
         target_link_libraries(hiprtc PUBLIC stdc++fs)
     endif()
+    target_link_libraries(hip_hcc PRIVATE /opt/rocm/lib/libamd_comgr.so)
```
aaronenyeshi: Some users, including CI machines won't install to this location.

aaronenyeshi: Better to add 
```
    if(HIP_PLATFORM STREQUAL "hcc")
        find_package(amd_comgr REQUIRED CONFIG
            PATHS
                /opt/rocm/
            PATH_SUFFIXES
                cmake/amd_comgr
                lib/cmake/amd_comgr
        )
        add_definitions(-DUSE_COMGR=1)
        MESSAGE(STATUS "Code Object Manager found at ${amd_comgr_DIR}.")
    endif()
```
Then use 
```
        target_link_libraries(hip_hcc PRIVATE amd_comgr)
        target_link_libraries(hip_hcc_static PRIVATE amd_comgr)
```

lmoriche: Done

yxsamliu:
I think we need more people familiar with v3 to review this.

wkwchau:
LGTM.   This change allows my implementation of hipOccupancyMaxPotentialBlockSize(), with minor modification, to use for the kernels with CO v3.  I had verified the codes produce the same results as the kernel with CO v2.   Laurent mentioned offline that it would be better to extract the information out of the metadata instead of using the kernel descriptor.  Given that HIP/HCC is going to be deprecated in the future, I will make the change on top of this only if it is ever needed.

aaronenyeshi:
+ @scott-linder and + @t-tye

scott-linder:
Also LGTM

aaronenyeshi:
It seems that this fails hiprtc tests on ROCm 2.4 release packages, but passes with compute-roc-master

gargrahul:
I think its other way, failing on rocm_head.
[rocm_head] The following tests FAILED:
[rocm_head] 	122 - directed_tests/hiprtc/hiprtcGetTypeName.tst (SEGFAULT)
[rocm_head] 	123 - directed_tests/hiprtc/hiprtcGetLoweredName.tst (SEGFAULT)
[rocm_head] 	124 - directed_tests/hiprtc/saxpy.tst (SEGFAULT)
[rocm_head] Errors while running CTest

gargrahul:
Moreover these HIPRTC test failures don't look like related to changes in this PR. Other PRs are also failing these tests.

PR #1139, opened 2019-05-24
Fix HIP-Clang missing libhiprtc.so packaging

aaronenyeshi:
HIP-Clang does not build libhiprtc.so, so the script needs to skip it. Sam already added the changes to skip it for build and install, but it seems `make package` is using a different script.

Since HIP-Clang does not build libhiprtc, using make package will error with missing file. Therefore, a workaround is needed to avoid this on hip-clang: `  6 install(FILES @PROJECT_BINARY_DIR@/libhiprtc.so DESTINATION lib)`.


yxsamliu:
It is better to condition it with @HIP_COMPILER@ instead of duplicate the file. It is easier to maintain.

aaronenyeshi:
> It is better to condition it with @HIP_COMPILER@ instead of duplicate the file. It is easier to maintain.

thank you, I've changed the workaround to use that reference instead

aaronenyeshi:
Hi @mangupta, could you please help review/merge? This is required for hip-clang packaging.

PR #1143, opened 2019-05-27
Fix hipcc for hip-clang.

yxsamliu:
If there is -g, do not add -O3 by default.
If HIP_VDI_HOME is not set, set HIP_VDI_HOME based on hipcc directory for HIP/VDI runtime.
For HIP/VDI runtime, set HIP_CLANG_PATH and DEVICE_LIB_PATH based on HIP_VDI_HOME only if they exist.

This allows using HIP/VDI runtime with hip-clang installed at /opt/rocm/llvm and device lib installed
at /opt/rocm/lib.

### DIFF
--- a/bin/hipcc
+++ b/bin/hipcc
@@ -107,24 +107,20 @@ $HIP_RUNTIME= $hipConfig{'HIP_RUNTIME'};
 
 # If using VDI runtime, need to find HIP_VDI_HOME
 if ($HIP_RUNTIME eq "VDI" and !defined $HIP_VDI_HOME) {
-    $HIP_VDI_HOME = "/opt/rocm/hip"
+    $HIP_VDI_HOME = abs_path(dirname($0) . "/..");
 }
 
 if (defined $HIP_VDI_HOME) {
-    my $bits = "";
-    if (-d "$HIP_VDI_HOME/bin/x86_64") {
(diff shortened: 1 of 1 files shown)


bin/hipcc
```diff
@@ -107,24 +107,20 @@ $HIP_RUNTIME= $hipConfig{'HIP_RUNTIME'};
 
 # If using VDI runtime, need to find HIP_VDI_HOME
 if ($HIP_RUNTIME eq "VDI" and !defined $HIP_VDI_HOME) {
-    $HIP_VDI_HOME = "/opt/rocm/hip"
+    $HIP_VDI_HOME = abs_path(dirname($0) . "/..");
```
mangupta: FYI: We have a symlink /opt/rocm/bin/hipcc that points to /opt/rocm/hip/bin/hipcc. Some users prefer to add /opt/rocm/bin to their PATH and directly execute hipcc. In which case HIP_VDI_HOME would point to /opt/rocm. This should be fine as long as the rest of the dependencies referred to by HIP_VDI_HOME are also symlinked correctly into /opt/rocm/bin, /opt/rocm/include and /opt/rocm/lib.

yxsamliu: fixed

PR #1147, opened 2019-05-28
fix compilation error when host compiler is clang

scchan:
The original `if defined` guard was there to provide an alternative code path when compiling with g++; however, this breaks when a plan vanilla clang is used for compiling host code.  The new guard fixes the problem.

### DIFF
--- a/include/hip/hcc_detail/hip_fp16.h
+++ b/include/hip/hcc_detail/hip_fp16.h
@@ -34,7 +34,7 @@ THE SOFTWARE.
     #include <utility>
 #endif
 
-#if __HCC__ || __HIP__ 
+#if __HCC_OR_HIP_CLANG__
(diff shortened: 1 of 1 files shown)


include/hip/hcc_detail/hip_fp16.h
```diff
@@ -34,7 +34,7 @@ THE SOFTWARE.
     #include <utility>
 #endif
 
-#if defined(__clang__) && (__clang_major__ > 5)
+#if __HCC__ || __HIP__
```
gargrahul: Should this include check for defined(__clang__) too as done in https://github.com/ROCm-Developer-Tools/HIP/blob/master/include/hip/hcc_detail/hip_common.h#L30?
OR could we use __HCC_OR_HIP_CLANG__ ?

scchan: replaced with HCC_OR_HIP_CLANG as suggested

PR #1150, opened 2019-05-29
[HIP][Tests] Added new testcases for Module API

ansurya:
Added new testcases for below module API's :

hipModuleGetGlobal
hipModuleGetTexRef	
hipModuleLoadData


tests/src/runtimeApi/module/global_kernel.cpp
```diff
@@ -0,0 +1,38 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+#include "hip/hip_runtime.h"
+
+#define ARRAY_SIZE (16)
+
+__device__ float myDeviceGlobal=0;
+extern float myDeviceGlobalArray[16];
```
yxsamliu: CUDA device functions and kernels cannot access host variable. This does not compile with nvcc or hip-clang.

ansurya: Agree updated test to work on CUDA device.

PR #1151, opened 2019-05-30
introduce hostcall

ssahasra:
The hostcall feature allows a kernel to request a service from the
host, and expect to block until it receives a response. It depends on
an external host library libamd_hostcall to provide the host-side
support.

When initializing, the HIP runtime now launches a single consumer
process that responds to hostcalls invoked by active kernels. A
separate hostcall buffer is managed for each HSA queue.

This initial implementation is an early preview, and may change in
unspecified ways in future releases. The first supported service is
the ability to invoke a host function identified by a pointer passed
to the kernel.

### DIFF
--- a/src/hip_hcc.cpp
+++ b/src/hip_hcc.cpp
@@ -40,6 +40,10 @@ THE SOFTWARE.
 #include <atomic>
 #include <mutex>
 
+#ifdef WITH_HOSTCALL
+#include <amd_hostcall.h>
+#endif // WITH_HOSTCALL
+
 #include <hc.hpp>
 #include <hc_am.hpp>
 #include "hsa/hsa_ext_amd.h"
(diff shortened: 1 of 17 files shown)


gregrodgers:
You should not use the hc::accelerator_view for assignHostcallBuffer because that is going away with HIP VDI.     

static void* assignHostcallBuffer(hc::accelerator_view *av, void *hwqueue)

ssahasra:
> 
> 
> You should not use the hc::accelerator_view for assignHostcallBuffer because that is going away with HIP VDI.
> 
> static void* assignHostcallBuffer(hc::accelerator_view *av, void *hwqueue)

The changes in this PR are specifically for the HCC implementation of HIP. This use of accelerator_view is a property of how the HIP streams are mapped to HSA queues in the HIP/HCC runtime, so there is no way to avoid the accelerator view. At the future point when VDI arrives, the mapping of hostcall buffers to underlying hardware queues will have to be reimplemented for that path.

src/hip_hcc.cpp
```diff
@@ -139,6 +143,103 @@ std::atomic<int> g_lastShortTid(1);
...
+        if (buffers.find(queue) == buffers.end())
+            return nullptr;
+        return buffers[queue];
+    }
+
+public:
+    hostcallState() : consumer(nullptr) {}
+
+    ~hostcallState() {
+        if (consumer) {
```
scchan: should we lock before destroy the consumer?

ssahasra: > 
> 
> should we lock before destroy the consumer?

Do you mean to say that some other HIP thread might be accessing the "consumer" variable when we destroy it? I guess it would be safer to do that, but not really necessary in this implementation. The global state is destroyed only when hip_hcc.so is unloaded, so it is safe to assume that no other thread is still accessing this global state.

gargrahul:
@ssahasra This PR is not building. Can you fix the build errors? You may need to resolve conflicts too.

gargrahul:
@ssahasra Why was this closed? Will you create a new PR to add this ?

ssahasra:
I think it got closed when I renamed a branch. The act of closing itself was not intentional, but let's leave it that way. I'll open a new PR when we really want to introduce hostcall in HIP/HCC.

gargrahul:
And when shall we introduce hostcall in HIP/HCC? I thought we were ready when this PR was submitted initially.

PR #1159, opened 2019-06-04
[hip] Make vector type's scalar conversion explicit.

mhbliao:


dereklstinson:
This threw a bunch of errors in gcc.

mhbliao:
> This threw a bunch of errors in gcc.

You need https://github.com/ROCm-Developer-Tools/HIP/pull/1160 to pacify them.

PR #1162, opened 2019-06-05
Implement the hipOccupancyMaxPotentialBlockSize function

wkwchau:
This version supports both CO v2 and v3.

### DIFF
--- a/include/hip/hcc_detail/hip_runtime.h
+++ b/include/hip/hcc_detail/hip_runtime.h
@@ -279,8 +279,6 @@ extern hipStream_t ihipPreLaunchKernel(hipStream_t stream, size_t grid, size_t b
                                        grid_launch_parm* lp, const char* kernelNameStr);
 extern void ihipPostLaunchKernel(const char* kernelName, hipStream_t stream, grid_launch_parm& lp);
 
-extern hipError_t ihipGetDeviceProperties(hipDeviceProp_t* props, int device);
-
 #if GENERIC_GRID_LAUNCH == 0
 //#warning "Original hipLaunchKernel defined"
 // Due to multiple overloaded versions of ihipPreLaunchKernel, the numBlocks3D and blockDim3D can be
--- a/include/hip/hcc_detail/hip_runtime.h
+++ b/include/hip/hcc_detail/hip_runtime.h
@@ -279,6 +279,8 @@ extern hipStream_t ihipPreLaunchKernel(hipStream_t stream, size_t grid, size_t b
                                        grid_launch_parm* lp, const char* kernelNameStr);
 extern void ihipPostLaunchKernel(const char* kernelName, hipStream_t stream, grid_launch_parm& lp);
 
+extern hipError_t ihipGetDeviceProperties(hipDeviceProp_t* props, int device);
+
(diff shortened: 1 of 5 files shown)


whchung:
Nice!  + @deven-amd / @jerryyin  for awareness.

src/hip_module.cpp
```diff
@@ -652,3 +652,145 @@ hipError_t hipModuleGetTexRef(textureReference** texRef, hipModule_t hmod, const
...
+
+    auto ctx = ihipGetTlsDefaultCtx();
+    hipError_t ret = hipSuccess;
+
+    if (ctx == nullptr) {
+        ret = hipErrorInvalidDevice;
+    }
+
+    hipDeviceProp_t prop{};
+    hipGetDeviceProperties(&prop, ihipGetTlsDefaultCtx()->getDevice()->_deviceId);
```
mangupta: Replace with ihipGetDeviceProperties() otherwise it will mess up the HIP trace.

wkwchau: Fixed.

gargrahul:
@wkwchau Could you add a test too?

gargrahul:
@wkwchau It needs ../../test_common.cpp.

Also isn't it possible to use this API without using Module APIs calls?

gargrahul:
@mangupta I have disabled the hipOccupancyMaxPotentialBlockSize dtest on NVCC for now. We need to discuss how to support both runtime and driver APIs support with single HIP API.

include/hip/hcc_detail/hip_runtime.h
```diff
@@ -279,6 +279,8 @@ extern hipStream_t ihipPreLaunchKernel(hipStream_t stream, size_t grid, size_t b
                                        grid_launch_parm* lp, const char* kernelNameStr);
 extern void ihipPostLaunchKernel(const char* kernelName, hipStream_t stream, grid_launch_parm& lp);
 
+extern hipError_t ihipGetDeviceProperties(hipDeviceProp_t* props, int device);
```
mangupta: Why do we need this here? This should go to src/hip_module.cpp or src/hip_hcc_internal.h.

wkwchau: Fixed by moving extern declaration of ihipGetDeviceProperties to hip_module.cpp.

src/hip_module.cpp
```diff
@@ -652,3 +654,147 @@ hipError_t hipModuleGetTexRef(textureReference** texRef, hipModule_t hmod, const
...
+        }
+
+        size_t maxWavesLDSLimited = min(wavefrontsLDS, maxWavefrontsPerCU);
+
+        size_t activeWavefronts = 0;
+        size_t tmp_min = (size_t)min(maxWavesLDSLimited, maxWavesWGLimited);
+        tmp_min = min(maxWavesSGPRSLimited, tmp_min);
+        activeWavefronts = min(maxWavesVGPRSLimited, tmp_min);
+
+        if (maxActivWaves < activeWavefronts) {
```
whchung: @wkwchau shall this line be `<=` instead of `<`?

PR #1171, opened 2019-06-14
[hipcc] Revise include path calculation.

mhbliao:
- Once HIP_VDI_HOME is defined but HIP_CLANG_INCLUDE_PATH is not,
  calculate it directly without HIP_CLANG_PATH is defined or not;
  Otherwise, we may leave HIP_CLANG_INCLUDE_PATH undefined, if clang is
  not installed following the official way (so far, HIP-Clang breaks
  that), we may leave HIP_CLANG_INCLUDE_PATH undefined before its uses.


aaronenyeshi:
This is required for both HIP-Samples and HIP-Examples for HIP-VDI

PR #1180, opened 2019-06-21
Put 3-wide vector types on a ketogenic diet.

AlexVlx:


AlexVlx:
@mangupta good catch, missed that one. Should be OK now methinks.

PR #1184, opened 2019-06-24
Fix HIP_VISIBLE_DEVICES order

gargrahul:
Fix HIP_VISIBLE_DEVICES order

### DIFF
--- a/src/hip_hcc.cpp
+++ b/src/hip_hcc.cpp
@@ -1399,7 +1399,7 @@ void ihipInit() {
 
     // Make sure the hip visible devices are within the deviceCnt range
     for (int i = 0; i < g_hip_visible_devices.size(); i++) {
-        if (g_hip_visible_devices[i] >= deviceCnt) {
+        if ((g_hip_visible_devices[i] >= deviceCnt) ||(g_hip_visible_devices[i] < 0)){
             // Make sure any DeviceID after invalid DeviceID will be erased.
             g_hip_visible_devices.resize(i);
             break;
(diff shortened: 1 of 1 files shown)


src/hip_hcc.cpp
```diff
@@ -1418,13 +1418,16 @@ void ihipInit() {
     for (int i = 0; i < accs.size(); i++) {
```
mangupta: Isn't it simpler if we simply run a loop over entries in ```g_hip_visible_devices``` rather than over ```accs```?

src/hip_hcc.cpp
```diff
@@ -1418,13 +1418,16 @@ void ihipInit() {
...
+            int index = i-1;
+            if (it == g_hip_visible_devices.end() && g_visible_device) {
                 // If device is not in visible devices list, ignore
                 continue;
+            } else {
+                 if(g_visible_device)
+                     index = std::distance(g_hip_visible_devices.begin(), it);
             }
-            g_deviceArray[g_deviceCnt] = new ihipDevice_t(g_deviceCnt, deviceCnt, accs[i]);
+            g_deviceArray[index] = new ihipDevice_t(index, deviceCnt, accs[i]);
```
mangupta: You might have to consider the case if ```index``` does not cover all entries in ```sizeof(g_deviceArray)```. Inverting the loop as suggested above should avoid this.

gargrahul:
Thanks @mangupta . I made changes as you have suggested. Please review again.

gargrahul:
Cool, thanks @mangupta , so we will fix the incomplete env variable case in a separate PR.

gargrahul:
@wenkaidu Could you please retest this PR with rccl-tests if not already done after recent changes?

PR #1188, opened 2019-06-26
Added missing NULL checks and corrected API return values

ansurya:
Fixes SWDEV-132169, SWDEV-132552, SWDEV-132794, SWDEV-132931, SWDEV-132970, SWDEV-135524, SWDEV-137442, SWDEV-137535, SWDEV-138479, SWDEV-138481, SWDEV-138799, SWDEV-141618.

### DIFF
--- a/src/hip_event.cpp
+++ b/src/hip_event.cpp
@@ -85,7 +85,7 @@ hipError_t ihipEventCreate(hipEvent_t* event, unsigned flags) {
         (flags & ~supportedFlags) ||             // can't set any unsupported flags.
         (flags & releaseFlags) == releaseFlags;  // can't set both release flags
 
-    if (!illegalFlags) {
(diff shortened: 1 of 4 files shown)


src/hip_event.cpp
```diff
@@ -256,18 +258,22 @@ hipError_t hipEventElapsedTime(float* ms, hipEvent_t start, hipEvent_t stop) {
 
 hipError_t hipEventQuery(hipEvent_t event) {
     HIP_INIT_SPECIAL_API(hipEventQuery, TRACE_QUERY, event);
+    if ( NULL == event)
+    {
+        return hipErrorInvalidResourceHandle;
```
mangupta: This should be ```return ihipLogStatus(hipErrorInvalidResourceHandle);```

PR #1190, opened 2019-06-27
Adding bounds check before hipMemset

cjatin:
Previously bounds were not checked before doing a hipMemset, which can result in undefined behavior, since a larger block than allocated memory can be overridden.
A bounds check is implemented which results in error in case bounds are breached.
This should be bring hipMemset and hipMemsetAsync to behave like cuda counterpart respectively.

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -1929,25 +1929,7 @@ hipError_t hipMemGetInfo(size_t* free, size_t* total) {
 hipError_t hipMemPtrGetInfo(void* ptr, size_t* size) {
     HIP_INIT_API(hipMemPtrGetInfo, ptr, size);
 
-    hipError_t e = hipSuccess;
-
-    if (ptr != nullptr && size != nullptr) {
-        hc::accelerator acc;
-#if (__hcc_workweek__ >= 17332)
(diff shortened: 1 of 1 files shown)


src/hip_memory.cpp
```diff
@@ -1510,6 +1510,15 @@ __global__ void hip_copy2d_n(T* dst, const T* src, size_t width, size_t height,
 }
 }  // namespace
 
+//Checks if operation is inbound, i.e. sizeBytes is no more than allocated size
+inline bool isMemoryOperationInbound(void* dst, size_t sizeBytes) {
+    size_t allocSize = 0;
+    if ((hipMemPtrGetInfo(dst, &allocSize) == hipSuccess) && (allocSize >= sizeBytes)) {
```
mangupta: HIP APIs should not call other HIP APIs. It messes up the HIP trace. You might need to move the implementation of ```hipMemPtrGetInfo``` to an internal function such as ```ihipMemPtrGetInfo``` and then use that both here as well as in ```hipMemPtrGetInfo```.

cjatin: Rewrote without hip trace APIs, please review.

PR #1192, opened 2019-06-27
Added support of HDP flush register memory location query

wkwchau:
The query of HDP flush register and memory location is performed through the hipDeviceGetAttribute() call with attribute of hipDeviceAttributeHdpMemFlushCntl and hipDeviceAttributeHdpRegFlushCntl, respectively.


wenkaidu:
Is there another pull request for this feature? RCCL still needs this patch.

PR #1202, opened 2019-07-03
SWDEV-193778 - Device side support for Cooperative Group feature

mshivama:
Implements device side data types and related APIs for following Cooperative Groups feature
  * Inter-Device CG (also called multi-grid-group)
  * Inter-Workgroup CG (also called grid-group)
  * Intra-Workgroup CG (thread-block, coalesced-group, etc)

Current Status:
  * Inter-Device CG:   Supported
  * Inter-Workgroup CG:   Supported
  * Intra-Workgroup CG:
       ** coalesced_group:   Supported
       ** thread_block:   Supported
       ** coalesced (dynamic) tiled partition:   Supported
       ** templated (static) tiled partition:   Supported
  * Unit Testing: The unit testcase file - hipCGDevice.cpp is added.

TODO Items:
  * tiled partitioning (both static and dynamic) of existing coalesced parent type: This requires some thinking since the size of coalesced type could be anywhere in the range [0, 64) as it is computed based on the current active waves within the wavefront. 
  * additional APIs support like shfl() need to be added: This probably seems to be straightforward to support.
  * grid_group::is_valid() API at present simply returns "true". We assume that there is nothing to check any API constraint here since otherwise kernel launch could not have happened. However we need to discuss it.

Overview of Implementation:
 * the header `hip/hip_cooperative_groups.h` implements the CG data types, and the functions defined within this header (both member and non-member functions) call wrappers implemented within the header `hcc_detail/hip_cooperative_groups_helper.h` in order to get the needed functionalities.
 * the wrappers in `hcc_detail/hip_cooperative_groups_helper.h` make use of device library functions implemented in the header `hcc_detail/device_functions.h`


mshivama:
@yxsamliu, @b-sumner, @gargrahul, @scchan, @mangupta,  @jlgreathouse, @arsenm

I kindly request you to review this pull request.

include/hip/hcc_detail/device_functions.h
```diff
@@ -756,26 +756,106 @@ unsigned long long int __ballot64(int predicate) {
...
 }
 
 __device__
 inline
-int64_t __lanemask_lt()
+uint64_t __lanemask_lt()
 {
-    int32_t activelane = __ockl_activelane_u32();
+    uint32_t lane = __ockl_lane_u32();
+    if (lane == 0)
```
b-sumner: Why is this needed?  If lane == 0, then mask == 0.

include/hip/hcc_detail/device_functions.h
```diff
@@ -756,26 +756,106 @@ unsigned long long int __ballot64(int predicate) {
...
+uint64_t  __lanemask_eq()
+{
+    uint32_t lane = __ockl_lane_u32();
+    int64_t mask = ((uint64_t)1 << lane);
+    return mask;
+}
+
+__device__
+inline
+uint64_t __activemask()
```
b-sumner: Why is this being added?

include/hip/hcc_detail/device_functions.h
```diff
@@ -756,26 +756,106 @@ unsigned long long int __ballot64(int predicate) {
...
+{
+    return __builtin_amdgcn_read_exec();
+}
+
+__device__
+inline
+uint32_t __lane()
```
b-sumner: Why is this being added?

include/hip/hcc_detail/device_library_decls.h
```diff
@@ -58,6 +58,15 @@ extern "C" __device__ __attribute__((const)) float __ocml_trunc_f32(float);
 extern "C" __device__ __attribute__((const)) float __ocml_fmin_f32(float, float);
 extern "C" __device__ __attribute__((const)) float __ocml_fmax_f32(float, float);
 
+extern "C" __device__ __attribute__((const)) uint __ockl_lane_u32();
```
b-sumner: You can use uint32_t here if desired, and remove the casts in the wrappers.

include/hip/hcc_detail/hip_cooperative_groups.h
```diff
@@ -0,0 +1,563 @@
...
+ *  \details User is not allowed to directly construct an object of type
+ *           `thread_block`. Instead, he should construct it through this API
+ *           function
+ */
+__CG_QUALIFIER__ thread_block
+this_thread_block() {
+  return thread_block(internal::workgroup::size());
+}
+
+/** \brief The coalesced cooperative group type
```
b-sumner: I would leave this completely out for now.  Better for the compiler to complain up front that the type does not exist.

include/hip/hcc_detail/hip_cooperative_groups.h
```diff
@@ -0,0 +1,563 @@
...
+  __CG_QUALIFIER__ thread_block_tile(internal::group_type type, uint64_t mask)
+    : thread_block_tile_base(type, mask) { }
+};
+
+/** \brief User exposed API interface to construct new coalesced (dynamic) tiled
+ *         partitioned group from an `exiting` intra-workgroup (and only
+ *         intra-workgroup) cooperative group types
+ */
+__CG_QUALIFIER__ thread_group
+tiled_partition(const thread_group& parent, uint32_t tile_sz) {
```
b-sumner: Again, I would leave this completely out for now.

include/hip/hcc_detail/device_library_decls.h
```diff
@@ -58,6 +58,15 @@ extern "C" __device__ __attribute__((const)) float __ocml_trunc_f32(float);
 extern "C" __device__ __attribute__((const)) float __ocml_fmin_f32(float, float);
 extern "C" __device__ __attribute__((const)) float __ocml_fmax_f32(float, float);
 
+extern "C" __device__ __attribute__((const)) uint32_t __ockl_lane_u32();
```
b-sumner: lane, both syncs and is_valid should only get convergent attribute.

arsenm: That's not strictly necessary as convergent is assumed by default

include/hip/hcc_detail/device_functions.h
```diff
@@ -748,26 +748,98 @@ unsigned long long int __ballot64(int predicate) {
...
+{
+    return __ockl_lane_u32();
+}
+
+/**
+ * grid and multi-grid cooperative groups related device library API wrappers
```
b-sumner: I'm not sure these need to be here.  Suggest calling OCKL functions directly from helper for the time being.

mshivama:
I have taken care of all the review comments by Brian. If this PR looks good, then, we should plan to merge it as it has to for 2.9

mshivama:
@b-sumner, If this PR is fine with you, please approve it so that Maneesh can merge it.

PR #1204, opened 2019-07-05
Fix even more typo.

darkbuck:
With this change, `hipVectorTypesDevice` passed


AlexVlx:
This is incorrect, you "fixed" the wrong thing. This'll return all ones on equality and all zeros on inequality. It's supposed to return all 0s on equality and all -1s, which is what it did. More granularly, it shoudl give -1 on true, not 1, since it follows the Clang extended vector type spec. See: https://gcc.godbolt.org/z/y722FK.

Issue #808, opened 2018-12-13
hipCommander examples are broken.


baryluk:
```
root@debian:/opt/rocm/hip/samples/1_Utils/hipCommander# ./hipCommander -v -f c.cmd 
Device:Fiji [Radeon R9 FURY / NANO Series] Mem=4.0GB #CUs=64 Freq=1050Mhz
error: Bad command 'loop,1000
error: assertion 0 at hipCommander.cpp:885: bad command in command-stream

Aborted
root@debian:/opt/rocm/hip/samples/1_Utils/hipCommander# ./hipCommander -v -f classic.cmd 
Device:Fiji [Radeon R9 FURY / NANO Series] Mem=4.0GB #CUs=64 Freq=1050Mhz
error: Bad command 'nullkernel,d2h,streamsync
error: assertion 0 at hipCommander.cpp:885: bad command in co mmand-stream

Aborted
#
```

and in many other files.

Some of the command files in `perf/` subdirectory work.

`latency2.hcm` for example works.

I am guessing this is because the format of commands is different.

`loop,1000;` (not working) vs `loop(1000)` (working).

And similar for streams.

Also there appears to be missing semicolons in various .cmd / .hcm files.

nartmada:
@baryluk, apologies for the lack of response from AMD.  Do you still need assistance with this issue?  Thanks.

harkgill-amd:
Hi @baryluk, the hipCommander sample test is no longer in use. The latest sample tests can be found at [ROCm/hip-tests](https://github.com/ROCm/hip-tests). Thanks!

PR #1207, opened 2019-07-08
HIPCommander code cleanup

ansurya:
=>Removed dependency on HSA and HC. 
=>Corrected scripts to use latest hipCommander syntax.
=>updated code to pass args as per module API 

Issues Addressed:
https://github.com/ROCm-Developer-Tools/HIP/issues/808


samples/1_Utils/hipCommander/hipCommander.cpp
```diff
@@ -9,8 +9,6 @@
 #include <hip/hip_runtime.h>
 #ifdef __HIP_PLATFORM_HCC__
 #include <elf.h>
```
mangupta: I think this include is also not required. We should be able to remove it.

ansurya: Fixed.

gargrahul:
@mangupta Shall we keep AQLCommand test ? OR with hipLauncKernelGGL , there shouldn't be any difference ? I think earlier AQLCommand was useful since PFE was used.

mangupta:
@gargrahul The reason why we are removing aqlcommand test is because of hip-clang and vdi. hipCommander sample should purely be using only HIP APIs. If we need aql tests, we should keep them separate from the samples.

samples/1_Utils/hipCommander/hipCommander.cpp
```diff
@@ -8,9 +8,6 @@
 
 #include <hip/hip_runtime.h>
 #ifdef __HIP_PLATFORM_HCC__
```
mangupta: I don't see the point of keeping an empty ```#ifdef ... #endif```.

bensander:
the HSA/ROCR modes could be useful for latency comparisons.  We could make these an #ifdef so tool still compiles on systems where ROCR is not available.

PR #1211, opened 2019-07-08
Add amdgpu targets gfx1010 and gfx1012

yxsamliu:


mangupta:
We probably also need to enable them in CMakeLists,txt while building HIP runtime.

yxsamliu:
> We probably also need to enable them in CMakeLists,txt while building HIP runtime.

This change is for passing the option to hip-clang. HCC does not support them yet. CMakeLists.txt need to be updated after HCC supports these archs.

PR #1213, opened 2019-07-09
UChar and UShort textures as Normalized Float

ansurya:
JIRA ISSUE : SWDEV-151668
Issue when reading UChar and UShort textures as Normalized Float

signed/unsigned char,short range is remapped to 0.f->1.f on texture read


gargrahul:
@ansurya You will have to modify other texture read functions as well.

PR #1216, opened 2019-07-10
Add missing compound assignment operators.

mhbliao:


mhbliao:
the CI failed due to irrelevant reason as follows, could someone help me have a look what's wrong on CI machine

[rocm_head] Error response from daemon: conflict: unable to remove repository reference "hip/build-ubuntu-16.04:latest" (must force) - container b0bfbe069362 is using its referenced image ff4f1cc14265

PR #1217, opened 2019-07-11
Increased the number of implicit-kernarg bytes to 56

cdevadas:
Allocate 8 bytes more for the new implicit kernarg - multi grid synchronization. Any consumers of this argument should note that it will be appended to the end of existing implicit kernargs.


cdevadas:
Ping.

PR #1224, opened 2019-07-12
Fix hipMemcpy-size test running out of Host Mem

aaronenyeshi:
The hipMemcpy-size uses a maxElem calculated from the total GPU mem /8. Then it will allocate 4 times that amount of host memory. This tests begins failing when there is not enough host memory, such as on systems with 32GB GPU mem, and 16GB RAM. This fixes the test if not enough host memory is available on the system.

### DIFF
--- a/tests/src/runtimeApi/memory/hipMemcpy.cpp
+++ b/tests/src/runtimeApi/memory/hipMemcpy.cpp
@@ -35,6 +35,7 @@ THE SOFTWARE.
 #include "test_common.h"
 
 #ifdef _WIN32
+#define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 #else
 #include "sys/types.h"
--- a/tests/src/runtimeApi/memory/hipMemcpy.cpp
+++ b/tests/src/runtimeApi/memory/hipMemcpy.cpp
@@ -299,10 +299,10 @@ void memcpytest2_get_host_memory(size_t& free, size_t& total) {
 struct sysinfo memInfo;
 void memcpytest2_get_host_memory(size_t& free, size_t& total) {
     sysinfo(&memInfo);
-    long long freePhysMem=memInfo.freeram + memInfo.freeswap;
+    long long freePhysMem=memInfo.freeram;
     freePhysMem *= memInfo.mem_unit;
     free = freePhysMem;
-    long long totalPhysMem=memInfo.totalram + memInfo.totalswap;
+    long long totalPhysMem=memInfo.totalram;
     totalPhysMem *= memInfo.mem_unit;
     total = totalPhysMem;
 }
--- a/tests/src/runtimeApi/memory/hipMemcpy.cpp
+++ b/tests/src/runtimeApi/memory/hipMemcpy.cpp
@@ -33,9 +33,13 @@ THE SOFTWARE.
 #include "hip/hip_runtime.h"
 #include "hip/hip_runtime.h"
 #include "test_common.h"
+
+#ifdef _WIN32
+#include <windows.h>
+#else
 #include "sys/types.h"
 #include "sys/sysinfo.h"
-
+#endif
 
 void printSep() {
     printf(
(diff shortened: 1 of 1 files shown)


aaronenyeshi:
The Jenkins failure is unrelated to my changes:
```
[rocm_head] The following tests FAILED:
[rocm_head] 	 98 - directed_tests/runtimeApi/memory/hipMemoryAllocateCoherentDriver.tst 
```

tests/src/runtimeApi/memory/hipMemcpy.cpp
```diff
@@ -33,6 +33,8 @@ THE SOFTWARE.
 #include "hip/hip_runtime.h"
 #include "hip/hip_runtime.h"
 #include "test_common.h"
+#include "sys/types.h"
```
yxsamliu: this does not work on windows. you may check https://stackoverflow.com/questions/2513505/how-to-get-available-memory-c-g

aaronenyeshi: Which macro do you use to guard linux and windows?

yxsamliu: #if _WIN32

aaronenyeshi:
@yxsamliu I've added the Windows support. I have verified it works on Linux, can you help verify on Windows?

tests/src/runtimeApi/memory/hipMemcpy.cpp
```diff
@@ -34,6 +34,12 @@ THE SOFTWARE.
 #include "hip/hip_runtime.h"
 #include "test_common.h"
 
+#ifdef _WIN32
+#include <windows.h>
```
yxsamliu: could you please add 
#define WIN32_LEAN_AND_MEAN
before #include <windows.h>
this is to avoid linking extra libs which we do not need.

tests/src/runtimeApi/memory/hipMemcpy.cpp
```diff
@@ -280,6 +287,26 @@ void memcpytest2_for_type(size_t numElements) {
...
+    status.dwLength = sizeof(status);
+    GlobalMemoryStatusEx(&status);
+    free = status.ullAvailPhys;
+    total = status.ullTotalPhys;
+}
+#else
+struct sysinfo memInfo;
+void memcpytest2_get_host_memory(size_t& free, size_t& total) {
+    sysinfo(&memInfo);
+    long long freePhysMem=memInfo.freeram + memInfo.freeswap;
```
yxsamliu: why do you add freeswap? we only want to use physical memory right?

aaronenyeshi: I thought if we include swap, then machines with low RAM (4GB), but high GPU RAM (32GB) can still run if they have swap space (32GB). Or should we just use physical?

tests/src/runtimeApi/memory/hipMemcpy.cpp
```diff
@@ -291,12 +318,20 @@ void memcpytest2_sizes(size_t maxElem = 0) {
...
-    size_t free, total;
+    size_t free, total, freeCPU, totalCPU;
     HIPCHECK(hipMemGetInfo(&free, &total));
+    memcpytest2_get_host_memory(freeCPU, totalCPU);
 
     if (maxElem == 0) {
-        maxElem = free / sizeof(T) / 8;
+        // Use lesser maxElem if not enough host memory available
+        size_t maxElemGPU = free / sizeof(T) / 8;
+        size_t maxElemCPU = freeCPU / sizeof(T) / 8;
```
yxsamliu: Probably  use maxElemCPU = freeCPU / sizeof(T) / 4 here. Then total host memory used is freeCPU/2 since there are two host arrays allocated. Then it should still be safe.

yxsamliu:
> The hipMemcpy-size uses a maxElem calculated from the total GPU mem /8. Then it will allocate 4 times that amount of host memory. This tests begins failing when there is not enough host memory, such as on systems with 32GB GPU mem, and 16GB RAM. This fixes the test if not enough host memory is available on the system.

This test will try to pin the host memory in one of the arrays. My understanding is that we can only pin physical memory. So we do not want to count swap space. If we can make sure the total size of allocated host memory is less than half of maximum free physical host memory it should be safe.

tests/src/runtimeApi/memory/hipMemcpy.cpp
aaronenyeshi: Hi Sam, it seems that each call to HostMemory<T> will allocate 2 arrays. So in this test, we are allocating 4 host memory array, 2 pinned and 2 not. So if we give freecpu/4, won't it occupy most of the cpu mem?

yxsamliu: You are right. So let's use maxElemCPU = freeCPU / sizeof(T) / 8

PR #1225, opened 2019-07-13
Use hcc when building with hcc compiler

pfultz2:


pfultz2:
Ping?

CMakeLists.txt
```diff
@@ -43,6 +43,15 @@ add_to_config(_versionInfo HIP_VERSION_MAJOR)
 add_to_config(_versionInfo HIP_VERSION_MINOR)
 add_to_config(_versionInfo HIP_VERSION_PATCH)
 
+if(CMAKE_CXX_COMPILER MATCHES ".*hcc")
+    set(HIP_PLATFORM "hcc" CACHE STRING "HIP Platform")
+    set(HIP_RUNTIME "HCC" CACHE STRING "HIP Runtime")
```
mangupta: Since HIP_RUNTIME and HIP_COMPILER default to hcc, you don't need this. Even if you do, you probably want to set HIP_COMPILER if CXX is hcc.

pfultz2: I removed setting `HIP_RUNTIME`.

PR #1226, opened 2019-07-14
Add Max Texture 1D,2D,3D device properties

ansurya:
Added Max Texture 1D,2D,3D Device Properties
JIRA [SWDEV-194646] :
 [HIP] Provide hipDeviceProp_t.maxTexture2D()


tests/src/runtimeApi/device/hipGetDeviceAttribute.cpp
```diff
@@ -109,5 +109,12 @@ int main(int argc, char* argv[]) {
     CHECK(test_hipDeviceGetAttribute(deviceId, hipDeviceAttributeMaxSharedMemoryPerMultiprocessor,
                                      props.maxSharedMemoryPerMultiProcessor));
     CHECK(test_hipDeviceGetAttribute(deviceId, hipDeviceAttributeIntegrated, props.integrated));
+
+    CHECK(test_hipDeviceGetAttribute(deviceId, hipDeviceAttributeMaxTexture1D, props.maxTexture1D));
+    CHECK(test_hipDeviceGetAttribute(deviceId, hipDeviceAttributeMaxTexture2DDimX, props.maxTexture2D[0]));
```
emankov: Please rename `hipDeviceAttributeMaxTexture2DDimX `and all below attributes to the correct names, as declared in `enum hipDeviceAttribute_t`.

ansurya: Thanks , changed test case to use correct enum values.

include/hip/hip_runtime_api.h
```diff
@@ -116,7 +116,10 @@ typedef struct hipDeviceProp_t {
     int gcnArch;                              ///< AMD GCN Arch Value. Eg: 803, 701
     int integrated;            ///< APU vs dGPU
     int cooperativeLaunch;            ///< HIP device supports cooperative launch
-    int cooperativeMultiDeviceLaunch; ///< HIP device supports cooperative launch on multiple devices 
+    int cooperativeMultiDeviceLaunch; ///< HIP device supports cooperative launch on multiple devices
```
yxsamliu: This is not backward compatible.

A HIP app can define a variable of hipDeviceProp_t. A HIP app built before this change will have a hipDeviceProp_t variable without these fields and will corrupt memory when runs with HIP runtime built after this change. Actually we saw PyTorch tests failing when running PyTorch built before this change with HIP runtime built after this change.

include/hip/nvcc_detail/hip_runtime_api.h
```diff
@@ -772,6 +772,13 @@ inline static hipError_t hipGetDeviceProperties(hipDeviceProp_t* p_prop, int dev
     p_prop->maxSharedMemoryPerMultiProcessor = cdprop.sharedMemPerMultiprocessor;
     p_prop->isMultiGpuBoard = cdprop.isMultiGpuBoard;
 
+    p_prop->maxTexture1D    = cdprop.maxTexture1D;
```
yxsamliu: To maintain backward compatibility, I think you just need to condition these writes with HIP patch version.

yxsamliu: > To maintain backward compatibility, I think you just need to condition these writes with HIP patch version.

sorry this does not work.

PR #1238, opened 2019-07-18
Added query of hipDeviceAttributeHdpMemFlushCntl and hipDeviceAttribu…

wkwchau:
…teHdpRegFlushCntl


wkwchau:
Abandon the original PR and resubmit this one to avoid issues due to code merged.

tests/src/runtimeApi/device/hipGetDeviceAttribute.cpp
```diff
@@ -116,5 +132,7 @@ int main(int argc, char* argv[]) {
     CHECK(test_hipDeviceGetAttribute(deviceId, hipDeviceAttributeMaxTexture3DWidth, props.maxTexture3D[0]));
     CHECK(test_hipDeviceGetAttribute(deviceId, hipDeviceAttributeMaxTexture3DHeight, props.maxTexture3D[1]));
     CHECK(test_hipDeviceGetAttribute(deviceId, hipDeviceAttributeMaxTexture3DDepth, props.maxTexture3D[2]));
+    CHECK(test_hipDeviceGetHdpAddress(deviceId, hipDeviceAttributeHdpMemFlushCntl, props.hdpMemFlushCntl));
```
mangupta: These two lines should not be executed on nvidia devices. So you probably need something like ```#ifndef __HIP_PLATFORM_NVCC__```.

mangupta:
@yxsamliu Do you need these 2 new device prop fields in this PR to also be excluded from VDI path?

yxsamliu:
> @yxsamliu Do you need these 2 new device prop fields in this PR to also be excluded from VDI path?

Not for now. These two fields should be excluded from VDI for now, like the above 3.

PR #1240, opened 2019-07-19
Added support of hipOccupancyMaxActiveBlocksPerMultiprocessor & hipOc…

wkwchau:
…cupancyMaxActiveBlocksPerMultiprocessorWithFlags APIs

### DIFF
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -885,10 +885,8 @@ hipError_t ihipOccupancyMaxPotentialBlockSize(uint32_t* gridSize, uint32_t* bloc
     using namespace hip_impl;
 
     auto ctx = ihipGetTlsDefaultCtx();
-    hipError_t ret = hipSuccess;
-
     if (ctx == nullptr) {
-        ret = hipErrorInvalidDevice;
+        return hipErrorInvalidDevice;
     }
 
     hipDeviceProp_t prop{};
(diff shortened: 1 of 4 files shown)


src/hip_module.cpp
```diff
@@ -1021,3 +1018,76 @@ hipError_t hipOccupancyMaxPotentialBlockSize(uint32_t* gridSize, uint32_t* block
...
+
+    // Due to SPI and private memory limitations, the max of wavefronts per CU in 32
+    size_t wavefrontSize = prop.warpSize;
+    size_t maxWavefrontsPerCU = min(prop.maxThreadsPerMultiProcessor / wavefrontSize, 32);
+
+    const size_t simdPerCU = 4;
+    const size_t maxWavesPerSimd = maxWavefrontsPerCU / simdPerCU;
+
+    size_t availableVGPRs = (prop.regsPerBlock / wavefrontSize / simdPerCU);
+    size_t alu_occupancy = simdPerCU * std::min(maxWavesPerSimd, availableVGPRs / usedVGPRS);
```
jlgreathouse: I know we don't normally run into SGPR limits, but should this also take into consideration that we have have 800 SGPRs per SIMD on gfx8+ and 512 SGPRs per SIMD on gfx9? This could cause lower occupancy than this current calculation in some corner cases.

wkwchau: Updated the codes to take into account for SGPRs usage.

PR #1243, opened 2019-07-22
remove stream locks where it is safe to do so

jeffdaily:
Do not lock the HIP stream while waiting for an event.  Unlike previous versions of HCC, waiting on an event no longer causes HCC to reclaim stream resources, it only causes HCC to free HSA signals associated with the event.

Remove unused `_kernelCnt` member.


mangupta:
@jeffdaily Do you know when this HCC change was introduced? Ideally we should guard the changes against appropriate hcc_workweek values.

jeffdaily:
@mangupta Should be when https://github.com/RadeonOpenCompute/hcc/pull/1046 got merged.

PR #1247, opened 2019-07-23
Added CooperativeLaunch and CooperativeMultiDeviceLaunch flag and pro…

wkwchau:
…perty for hipDeviceGetAttribute() and hipGetDeviceProperties()

** Resubmit the change to avoid the build issue **


tests/src/runtimeApi/device/hipGetDeviceAttribute.cpp
```diff
@@ -116,5 +116,9 @@ int main(int argc, char* argv[]) {
     CHECK(test_hipDeviceGetAttribute(deviceId, hipDeviceAttributeMaxTexture3DWidth, props.maxTexture3D[0]));
     CHECK(test_hipDeviceGetAttribute(deviceId, hipDeviceAttributeMaxTexture3DHeight, props.maxTexture3D[1]));
     CHECK(test_hipDeviceGetAttribute(deviceId, hipDeviceAttributeMaxTexture3DDepth, props.maxTexture3D[2]));
+
+    CHECK(test_hipDeviceGetAttribute(deviceId, hipDeviceAttributeCooperativeLaunch, props.cooperativeLaunch));
```
mangupta: We need to add ```hipDeviceAttributeCooperativeLaunch``` and ```hipDeviceAttributeCooperativeMultiDeviceLaunch``` to include/hip/nvcc_detail/hip_runtime_api.h as well. Otherwise it breaks this test on nvcc path.

PR #1254, opened 2019-07-25
Fixing Compilation warnings

cjatin:
Fixing Few warnings which occur during compilation

- Initialize a bool variable
- Removed call to tmpnam, which is unsafe, replaced it with a log of incremental file name generation

### DIFF
--- a/src/hiprtc.cpp
+++ b/src/hiprtc.cpp
@@ -354,13 +354,17 @@ namespace
 {
     class Unique_temporary_path {
         // DATA
+        std::string defaultName{"/tmp/hipTmpFile_"};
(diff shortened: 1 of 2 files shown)


src/hiprtc.cpp
```diff
@@ -354,13 +354,17 @@ namespace
 {
     class Unique_temporary_path {
         // DATA
+        std::string defaultName{"/tmp/hipTmpFile_"};
```
emankov: Not a cross-platform solution.

mangupta: I agree. I feel continuing to use ```tmpnam``` is better.

PR #1264, opened 2019-07-29
Fix cross-reference in porting guide

dalg24:


gargrahul:
@dalg24 Thanks for submitting your change, however are you using HIP texture reference APIs. In HIP we no longer need and additional argument while using texture reference as being used @
https://github.com/ROCm-Developer-Tools/HIP/blob/master/tests/src/texture/hipTextureRef2D.cpp#L17

gargrahul:
@dalg24 I submitted #1265 , which overwrites your change.

dalg24:
> @dalg24 Thanks for submitting your change, however are you using HIP texture reference APIs ?

No I am not.

Thanks for updating the doc.

PR #1267, opened 2019-07-30
Testcase to validate signed/unsigned char,short as normalized float

ansurya:
Added new testcase to validate char,uchar,short,ushort texture as normalized float.

### DIFF
--- a/tests/src/texture/hipNormalizedFloatValueTex.cpp
+++ b/tests/src/texture/hipNormalizedFloatValueTex.cpp
@@ -21,7 +21,7 @@ THE SOFTWARE.
 */
 
 /* HIT_START
- * BUILD: %t %s ../../test_common.cpp EXCLUDE_HIP_PLATFORM nvcc
+ * BUILD: %t %s ../test_common.cpp EXCLUDE_HIP_PLATFORM nvcc
  * TEST: %t
  * HIT_END
  */
--- a/tests/src/texture/hipNormalizedFloatValueTex.cpp
+++ b/tests/src/texture/hipNormalizedFloatValueTex.cpp
@@ -1,0 +1,122 @@
+/*
+Copyright (c) 2019-present Advanced Micro Devices, Inc. All rights reserved.
+ 
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+ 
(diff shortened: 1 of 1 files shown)


tests/src/texture/hipNormalizedFloatValueTex.cpp
```diff
@@ -0,0 +1,122 @@
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+/* HIT_START
+ * BUILD: %t %s ../test_common.cpp EXCLUDE_HIP_PLATFORM nvcc
```
mangupta: Is there any specific reason why we are excluding this on nvidia? Ideally this test should work there as well.

tests/src/texture/hipNormalizedFloatValueTex.cpp
```diff
@@ -0,0 +1,122 @@
...
+ * TEST: %t
+ * HIT_END
+ */
+
+#define CHECK(cmd) \
```
gargrahul: There is HIP_CHECK already defined in test_common.h , which is included below.

tests/src/texture/hipNormalizedFloatValueTex.cpp
```diff
@@ -0,0 +1,122 @@
...
+    desc.Height = 1;
+    desc.Format = texFormat;
+    desc.NumChannels = 1;
+    CHECK(hipTexRefSetAddress2D(texRef, &desc, (hipDeviceptr_t)dData, sizeof(T)*SIZE));
+    
+    bool testResult = true;
+    float *dOutputData = NULL;
+    CHECK(hipMalloc((void **) &dOutputData, sizeof(float)*SIZE));
+ 
+    hipLaunchKernelGGL(HIP_KERNEL_NAME(normalizedValTextureTest), dim3(1,1,1), dim3(SIZE,1,1), 0, 0, SIZE, dOutputData);
```
gargrahul: Do we really need HIP_KERNEL_NAME here?

gargrahul:
@mangupta @ansurya  There is another bigger issue with this test. hipTexRefXXXX APIs are equivalent to CUDA driver APIs whereas this sample has used these as runtime APIs. This is why I think we should try to run tests on NVCC path as well.

PR #1271, opened 2019-07-31
[HIP] hip_throw fails to output runtime_error string (github_issue#1008)

SarbojitAMD:
hip_hcc.cpp, hip_util.h : Created new hip_throw macro and removed old implementation
 macro_based_grid_launch.hpp : rethrow should not use exception object hence removed from catch block


gargrahul:
@SarbojitAMD Is there a way to test these changes?

SarbojitAMD:
@gargrahul , Offline I shared test files that I used for testing the change, please let me know if anything specific you are looking for.

AlexVlx:
No we should absolutely not use macros, and especially function like macros, they are ghastly. I am conceptually against this. Typecasting is fine at this point, you already threw and caught exception, things are already expensive. There are no other error types, these are the roots for the stdexcept hierarchy; if you do not inherit from these you need to rewrite your impl anyway.

mangupta:
closed in favor of PR #1285.

PR #1286, opened 2019-08-05
Partial revert of 'Add HSA_PATH to hip_Includes in cmake and hipconfig (#1260)'

mangupta:
Fixes:
SWDEV-198421
SWDEV-198753

### DIFF
--- a/hip-config-clang.cmake.in
+++ b/hip-config-clang.cmake.in
@@ -1,16 +1,4 @@
 @PACKAGE_INIT@
-if(NOT DEFINED HSA_PATH)
-    if(NOT DEFINED ENV{HSA_PATH})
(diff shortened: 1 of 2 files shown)


emankov:
As long as `HSA_PATH` is already used, in order to be compatible with `CMAKE_PREFIX_PATH`, `HSA_PATH` might be included in `find_path` instead of reverting PR #1260.

mangupta:
> As long as `HSA_PATH` is already used, in order to be compatible with `CMAKE_PREFIX_PATH`, `HSA_PATH` might be included in `find_path` instead of reverting PR #1260.
The revert is because this breaks the build of some components. It needs to be reworked as well as accompanied possibly by changes in those components. Till then reverting the "offending" part of the commit.

PR #1287, opened 2019-08-05
[HIP] Fix for bug : hipMemcpyPeer accepts same device as Peer(#140614)

SarbojitAMD:


src/hip_peer.cpp
```diff
@@ -152,6 +152,8 @@ hipError_t ihipEnablePeerAccess(hipCtx_t peerCtx, unsigned int flags) {
 //---
 hipError_t hipMemcpyPeer(void* dst, hipCtx_t dstCtx, const void* src, hipCtx_t srcCtx,
                          size_t sizeBytes) {
+    if(dstCtx == srcCtx) return hipErrorInvalidValue;
```
mangupta: This check should be after the HIP_INIT_API to ensure HIP tracing & lazy init works correctly. Also return should be wrapped in ihipLogStatus() so that logging works correctly.

src/hip_peer.cpp
```diff
@@ -163,6 +165,8 @@ hipError_t hipMemcpyPeer(void* dst, hipCtx_t dstCtx, const void* src, hipCtx_t s
 //---
 hipError_t hipMemcpyPeerAsync(void* dst, hipCtx_t dstDevice, const void* src, hipCtx_t srcDevice,
                               size_t sizeBytes, hipStream_t stream) {
+    if(dstDevice == srcDevice) return hipErrorInvalidValue;
```
mangupta: See above.

gargrahul:
@SarbojitAMD I think this change is NOT needed. Did you check the behavior of equivalent API?

SarbojitAMD:
Closing the PR as we have decided not to make change.

PR #1293, opened 2019-08-06
[dtests] Fix build issues with hipLaunchParm.cpp on windows

amd-lthakur:
"Removed unwanted #include sys/time.h , gettimeofday() and timeval variables and this also helps avoid compilation error in windows due to gettimeofday() call equivalent of which is not available in windows"

### DIFF
--- a/tests/src/kernel/hipLaunchParm.cpp
+++ b/tests/src/kernel/hipLaunchParm.cpp
@@ -577,7 +577,7 @@ __global__ void vAdd(float* a) {}
 // Some wrapper macro for testing:
 #define WRAP(...) __VA_ARGS__
 
-#define GPU_PRINT_TIME(cmd, elapsed, quiet)                         \
+#define MY_LAUNCH_MACRO(cmd, elapsed, quiet)                         \
     do {                                                            \
         hipDeviceSynchronize();                                     \
         cmd;                                                        \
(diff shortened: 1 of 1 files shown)


gargrahul:
@amd-lthakur Thanks for changes however instead of removing non-windows specific pieces can't you guard those using #if to avoid them while testing on Windows.

amd-lthakur:
> @amd-lthakur Thanks for changes however instead of removing non-windows specific pieces can't you guard those using #if to avoid them while testing on Windows.

Hi Rahul,
The changes provided were agreed upon by Maneesh. Basically the include part was for gettimeofday() which was actually not being used at all in the macro. Hence removed them.

tests/src/kernel/hipLaunchParm.cpp
```diff
@@ -577,16 +577,11 @@ __global__ void vAdd(float* a) {}
 // Some wrapper macro for testing:
 #define WRAP(...) __VA_ARGS__
 
-#include <sys/time.h>
 #define GPU_PRINT_TIME(cmd, elapsed, quiet)                         \
```
mangupta: Just a cosmetic change. But can we rename the MACRO to something else? Say MY_LAUNCH_MACRO or something? GPU_PRINT_TIME seems odd especially since we don't have any GPU timing code or prints in this macro.

PR #1296, opened 2019-08-06
Add hipMemcpyParam2DAsync

gargrahul:
Equivalent to cuMemcpy2DAsync.

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -1694,10 +1694,6 @@ hipError_t hipMemcpy2D(void* dst, size_t dpitch, const void* src, size_t spitch,
     return ihipLogStatus(e);
 }
 
-hipError_t hipMemcpy2DAsync(void* dst, size_t dpitch, const void* src, size_t spitch, size_t width,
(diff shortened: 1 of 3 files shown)


src/hip_memory.cpp
```diff
@@ -1750,6 +1757,17 @@ hipError_t hipMemcpyParam2D(const hip_Memcpy2D* pCopy) {
     return ihipLogStatus(e);
 }
 
+hipError_t hipMemcpyParam2DAsync(const hip_Memcpy2D* pCopy, hipStream_t stream) {
+    HIP_INIT_SPECIAL_API(hipMemcpyParam2DAsync, (TRACE_MCMD), pCopy, stream);
+    hipError_t e = hipSuccess;
+    if (pCopy == nullptr) {
+        e = hipErrorInvalidValue;
+    }
+    e = ihipMemcpy2DAsync(pCopy->dstArray->data, pCopy->WidthInBytes, pCopy->srcHost, pCopy->srcPitch,
```
mangupta: This should probably be under the else case...

PR #1281, opened 2019-08-02
fixed getopt requirement for windows, fixed setenv() system calls req…

amd-lthakur:
fixed getopt requirement for windows, fixed setenv() system calls requirements for windows platform, fixed popen, pclose system call requirements for windows platform.

PR #1299, opened 2019-08-07
[dtests[ Fix build issues with hipEnvVar*.cpp on windows

ansurya:
Platform specific implementation to use setenv(), popen, pclose system calls for linux and  _putenv_s, _popen, _pclose for windows. platform is detected with predefined macros.
Replaced getopt with clara based implementation which is platform independent.

### DIFF
--- a/tests/src/hipEnvVar.cpp
+++ b/tests/src/hipEnvVar.cpp
@@ -33,16 +33,11 @@ THE SOFTWARE.
 #include "hip/hip_runtime.h"
 
 #ifdef _WIN64
-    #include <tchar.h>
+#include <tchar.h>
 #endif
 
-
 #ifdef _WIN64
-    #define SETENV(x,y,z) _putenv_s(x,y)
-#elif __unix__
-    #define SETENV(x,y,z) setenv(x,y,z)
-#else
-    #error "Could not detect OS platform."
+#define setenv(x,y,z) _putenv_s(x,y)
 #endif
 
 using namespace std;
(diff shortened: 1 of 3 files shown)


ansurya:
setenv(), popen, pclose system call requirements for windows platform is taken as is from PR #1281. Replaced getopt with clara based implementation.

emankov:
LGTM. The only concern is about BSD 2-Clause license. Does it allow such manipulations with header files?

tests/src/hipEnvVar.cpp
```diff
@@ -27,57 +27,61 @@ THE SOFTWARE.
```
gargrahul: Is this test being run on CI?

gargrahul:
These changes would need verification on Windows too.

tests/src/hipEnvVar.cpp
ansurya: This test is used by hipEnvVarDriver.cpp which is run on CI.

ansurya:
These changes are verified by QA team on windows.

mangupta:
> LGTM. The only concern is about BSD 2-Clause license. Does it allow such manipulations with header files?

The clara.hpp is already being used in HIP. See https://github.com/ROCm-Developer-Tools/HIP/blob/master/lpl_ca/clara/clara.hpp.

emankov:
> The clara.hpp is already being used in HIP. See https://github.com/ROCm-Developer-Tools/HIP/blob/master/lpl_ca/clara/clara.hpp.

I see, but the question is in such a mixed usage of header files under the BSD 2-Clause license. Is it allowed, or the original header files should be used?

mangupta:
> > The clara.hpp is already being used in HIP. See https://github.com/ROCm-Developer-Tools/HIP/blob/master/lpl_ca/clara/clara.hpp.
> 
> I see, but the question is in such a mixed usage of header files under the BSD 2-Clause license. Is it allowed, or the original header files should be used?

I am not sure which header are we modifying. We are using the single_include version of clara.hpp. We are also probably using an older version of the file. Most likely this one: https://github.com/catchorg/Clara/blob/6462131410c4c579ea60fea2db34845de88f468c/single_include/clara.hpp.

emankov:
> I am not sure which header are we modifying. We are using the single_include version of clara.hpp. We are also probably using an older version of the file. Most likely this one: https://github.com/catchorg/Clara/blob/6462131410c4c579ea60fea2db34845de88f468c/single_include/clara.hpp.

I've been worrying about the following:
```
// ----------- end of #include from clara_textflow.hpp -----------
// ........... back in clara.hpp
```
As long as it is so in the original (single_include) version of Clara, FMPOV all is ok.

ansurya:
> LGTM. Please write a couple of sentences about what this PR is done for.

Done.

PR #1306, opened 2019-08-08
[dtests] Fix build issue with hipMemcpy_simple.cpp on windows

amd-lthakur:
Compilation error being observed on windows due to aligned_alloc() call. Mapped the call to _aligned_malloc() for windows.


gargrahul:
@mangupta wouldn't it be useful to put this fix in test_common.h so that multiple individual tests were not required changes ?

PR #1309, opened 2019-08-09
[dtests] Fix build issues with hipMemoryAllocateCoherentDriver.cpp on windows

amd-lthakur:
Compilation error being observed due to popen(), pclose() and setenv() linux calls on windows. Replaced with appropriate calls on windows.


gargrahul:
@mangupta  IMO it is better to make these windows specific fixes part of test_common.h so that multiple tests can make use of it as needed.

PR #1320, opened 2019-08-12
Add hipMemcpy3DAsync

gargrahul:


src/hip_memory.cpp
```diff
@@ -1471,6 +1476,22 @@ hipError_t hipMemcpy3D(const struct hipMemcpy3DParms* p) {
     } else {
         e = hipErrorInvalidValue;
     }
+    return e;
+}
+
+hipError_t hipMemcpy3D(const struct hipMemcpy3DParms* p) {
+    HIP_INIT_SPECIAL_API(hipMemcpy3D, (TRACE_MCMD), p);
+    hipError_t e = hipSuccess;
+    hipStream_t stream = ihipSyncAndResolveStream(hipStreamNull);
```
mangupta: We shouldn't do this here. What if p is NULL? We simply end up sync'ing the stream.

src/hip_memory.cpp
```diff
@@ -1471,6 +1476,22 @@ hipError_t hipMemcpy3D(const struct hipMemcpy3DParms* p) {
...
+    e = ihipMemcpy3D(p, stream, false);
+    return ihipLogStatus(e);
+}
+
+hipError_t hipMemcpy3DAsync(const struct hipMemcpy3DParms* p, hipStream_t inStream) {
+    HIP_INIT_SPECIAL_API(hipMemcpy3DAsync, (TRACE_MCMD), p, inStream);
+    hipError_t e = hipSuccess;
+    hipStream_t stream = ihipSyncAndResolveStream(inStream);
```
mangupta: Same as above.

src/hip_memory.cpp
```diff
@@ -1471,6 +1476,22 @@ hipError_t hipMemcpy3D(const struct hipMemcpy3DParms* p) {
...
         e = hipErrorInvalidValue;
     }
+    return e;
+}
+
+hipError_t hipMemcpy3D(const struct hipMemcpy3DParms* p) {
+    HIP_INIT_SPECIAL_API(hipMemcpy3D, (TRACE_MCMD), p);
+    hipError_t e = hipSuccess;
+    hipStream_t stream = ihipSyncAndResolveStream(hipStreamNull);
+    e = ihipMemcpy3D(p, stream, false);
```
mangupta: Probably better to pass hipStreamNull here and move the resolve back to ihipMemcpy3D

src/hip_memory.cpp
```diff
@@ -1448,11 +1447,14 @@ hipError_t hipMemcpy3D(const struct hipMemcpy3DParms* p) {
             ySize = p->srcPtr.ysize;
             dstPitch = p->dstPtr.pitch;
         }
-        hipStream_t stream = ihipSyncAndResolveStream(hipStreamNull);
```
mangupta: Think it is better to keep it here.

PR #1321, opened 2019-08-13
Fix memcpy with IPC slowness

gargrahul:


src/hip_memory.cpp
```diff
@@ -2137,18 +2137,35 @@ hipError_t hipIpcOpenMemHandle(void** devPtr, hipIpcMemHandle_t handle, unsigned
...
-        auto ctx = ihipGetTlsDefaultCtx();
-        {
-            LockedAccessor_CtxCrit_t crit(ctx->criticalData());
-            // the peerCnt always stores self so make sure the trace actually
-            hsa_status_t hsa_status = hsa_amd_ipc_memory_attach(
-                (hsa_amd_ipc_memory_t*)&(iHandle->ipc_handle), iHandle->psize, crit->peerCnt(),
-                crit->peerAgents(), devPtr);
-            if (hsa_status != HSA_STATUS_SUCCESS) hipStatus = hipErrorMapBufferObjectFailed;
+        if (!agent) {
+            hipStatus = hipErrorInvalidResourceHandle;
```
srinivamd: Nit: How about add the return statement (line 2174) here after 2141, and decrease indent of the entire else block? It would make reading the code easier ( and let the compiler do its magic!)

src/hip_memory.cpp
```diff
@@ -2137,18 +2137,35 @@ hipError_t hipIpcOpenMemHandle(void** devPtr, hipIpcMemHandle_t handle, unsigned
...
+        } else {
+            ihipIpcMemHandle_t* iHandle = (ihipIpcMemHandle_t*)&handle;
+            // Attach ipc memory
+            auto ctx = ihipGetTlsDefaultCtx();
+            {
```
srinivamd: Nit: Why is this block {..} needed when there is nothing after it?

src/hip_memory.cpp
```diff
@@ -2164,8 +2181,12 @@ hipError_t hipIpcCloseMemHandle(void* devPtr) {
         hipStatus = hipErrorInvalidValue;
     } else {
 #if USE_IPC
-        hsa_status_t hsa_status = hsa_amd_ipc_memory_detach(devPtr);
-        if (hsa_status != HSA_STATUS_SUCCESS) return hipErrorInvalidResourceHandle;
+        am_status_t am_status = hc::am_memtracker_remove(devPtr);
```
srinivamd: Should this also be protected by LockedAccessor_CtxCrit as the attach/am_memtracker_add are - see line 2147 above?

src/hip_memory.cpp
```diff
@@ -2164,8 +2181,12 @@ hipError_t hipIpcCloseMemHandle(void* devPtr) {
...
+        if (am_status == AM_SUCCESS) {
+            hsa_status_t hsa_status = hsa_amd_ipc_memory_detach(devPtr);
+            if (hsa_status != HSA_STATUS_SUCCESS) 
+                hipStatus = hipErrorInvalidResourceHandle;
+        } else hipStatus = hipErrorInvalidValue;
```
srinivamd: How about logging an error on am_memtracker_remove error return, assuming that it should not really happen but if it does it would be good to have the log?

src/hip_memory.cpp
```diff
@@ -2137,18 +2137,35 @@ hipError_t hipIpcOpenMemHandle(void** devPtr, hipIpcMemHandle_t handle, unsigned
...
+                // the peerCnt always stores self so make sure the trace actually
+                hsa_status_t hsa_status = hsa_amd_ipc_memory_attach(
+                    (hsa_amd_ipc_memory_t*)&(iHandle->ipc_handle), iHandle->psize, crit->peerCnt(),
+                    crit->peerAgents(), devPtr);
+                if (hsa_status == HSA_STATUS_SUCCESS) {
+                    hc::AmPointerInfo ampi(NULL, *devPtr, *devPtr, sizeof(*devPtr), acc, true, true);
+                    am_status_t am_status = hc::am_memtracker_add(*devPtr,ampi);
+                    if (am_status == AM_SUCCESS) {
+#if USE_APP_PTR_FOR_CTX
+                        hc::am_memtracker_update(*devPtr, device->_deviceId, 0, ctx);
```
srinivamd: Return error code from am_memtracker_update() not checked - should there be a check?

src/hip_memory.cpp
```diff
@@ -2137,18 +2137,35 @@ hipError_t hipIpcOpenMemHandle(void** devPtr, hipIpcMemHandle_t handle, unsigned
...
+#else
+                        hc::am_memtracker_update(*devPtr, device->_deviceId, 0);
```
srinivamd: Same as above: should return code from am_memtracker_update() be checked?

src/hip_memory.cpp
gargrahul: Made changes to make an early return when there is an error.

src/hip_memory.cpp
gargrahul: This is to understand scope of LockedAccessor_CtxCrit_t.

src/hip_memory.cpp
gargrahul: No

src/hip_memory.cpp
gargrahul: Added these checks.

src/hip_memory.cpp
gargrahul: added the check.

src/hip_memory.cpp
```diff
@@ -2131,24 +2131,43 @@ hipError_t hipIpcOpenMemHandle(void** devPtr, hipIpcMemHandle_t handle, unsigned
...
-        {
-            LockedAccessor_CtxCrit_t crit(ctx->criticalData());
-            // the peerCnt always stores self so make sure the trace actually
-            hsa_status_t hsa_status = hsa_amd_ipc_memory_attach(
-                (hsa_amd_ipc_memory_t*)&(iHandle->ipc_handle), iHandle->psize, crit->peerCnt(),
-                crit->peerAgents(), devPtr);
-            if (hsa_status != HSA_STATUS_SUCCESS) hipStatus = hipErrorMapBufferObjectFailed;
+        if (!agent) {
+            return ihipLogStatus(hipErrorInvalidResourceHandle);
+        } else {
```
emankov: excessive `else`

src/hip_memory.cpp
```diff
@@ -2131,24 +2131,43 @@ hipError_t hipIpcOpenMemHandle(void** devPtr, hipIpcMemHandle_t handle, unsigned
...
+            // Attach ipc memory
+            auto ctx = ihipGetTlsDefaultCtx();
+            {
+                LockedAccessor_CtxCrit_t crit(ctx->criticalData());
+                auto device = ctx->getWriteableDevice();
+                // the peerCnt always stores self so make sure the trace actually
+                hsa_status_t hsa_status = hsa_amd_ipc_memory_attach(
+                    (hsa_amd_ipc_memory_t*)&(iHandle->ipc_handle), iHandle->psize, crit->peerCnt(),
+                    crit->peerAgents(), devPtr);
+                if (hsa_status == HSA_STATUS_SUCCESS) {
```
emankov: excessive `hsa_status_t hsa_status`

src/hip_memory.cpp
```diff
@@ -2131,24 +2131,43 @@ hipError_t hipIpcOpenMemHandle(void** devPtr, hipIpcMemHandle_t handle, unsigned
...
+                        am_status = hc::am_memtracker_update(*devPtr, device->_deviceId, 0);
+#endif
+                        if(am_status != AM_SUCCESS)
+                            return ihipLogStatus(hipErrorMapBufferObjectFailed);
+                    } else {
+                        hipStatus = hipErrorInvalidValue;
+                    }
+                } else {
+                    hipStatus = hipErrorMapBufferObjectFailed;
+                }
```
emankov: excessive braces for both `else`'s

src/hip_memory.cpp
```diff
@@ -2130,69 +2130,68 @@ hipError_t hipIpcGetMemHandle(hipIpcMemHandle_t* handle, void* devPtr) {
...
 #if USE_IPC
-        am_status_t am_status = hc::am_memtracker_remove(devPtr);
-        if (am_status == AM_SUCCESS) {
-            hsa_status_t hsa_status = hsa_amd_ipc_memory_detach(devPtr);
-            if (hsa_status != HSA_STATUS_SUCCESS) 
-                hipStatus = hipErrorInvalidResourceHandle;
-        } else hipStatus = hipErrorInvalidValue;
+    am_status_t am_status = hc::am_memtracker_remove(devPtr);
+    if (am_status != AM_SUCCESS)
+        return ihipLogStatus(hipErrorInvalidValue);
```
emankov: `am_status_t am_status` is excessive.

src/hip_memory.cpp
```diff
@@ -2130,69 +2130,68 @@ hipError_t hipIpcGetMemHandle(hipIpcMemHandle_t* handle, void* devPtr) {
...
+
+    hsa_status_t hsa_status = hsa_amd_ipc_memory_detach(devPtr);
+    if (hsa_status != HSA_STATUS_SUCCESS)
+        return ihipLogStatus(hipErrorInvalidResourceHandle);
```
emankov: `hsa_status_t hsa_status` is excessive.

src/hip_memory.cpp
```diff
@@ -2130,69 +2130,68 @@ hipError_t hipIpcGetMemHandle(hipIpcMemHandle_t* handle, void* devPtr) {
...
+    {
+        LockedAccessor_CtxCrit_t crit(ctx->criticalData());
+        auto device = ctx->getWriteableDevice();
+        // the peerCnt always stores self so make sure the trace actually
+        hsa_status_t hsa_status = hsa_amd_ipc_memory_attach(
+            (hsa_amd_ipc_memory_t*)&(iHandle->ipc_handle), iHandle->psize, crit->peerCnt(),
+             crit->peerAgents(), devPtr);
+
+        if (hsa_status != HSA_STATUS_SUCCESS)
+            return ihipLogStatus(hipErrorRuntimeOther);
```
emankov: `hsa_status_t hsa_status` is excessive.

src/hip_memory.cpp
gargrahul: Not sure what do you mean. am_status is being used later too.

emankov: > Not sure what do you mean. am_status is being used later too.

Not in this function.

PR #1322, opened 2019-08-13
[HIP] Fix for hipArray_t failure on nvcc path

SarbojitAMD:
Fixes SWDEV#148407

### DIFF
--- a/include/hip/nvcc_detail/hip_runtime_api.h
+++ b/include/hip/nvcc_detail/hip_runtime_api.h
@@ -160,6 +160,7 @@ typedef CUmodule hipModule_t;
 typedef CUfunction hipFunction_t;
 typedef CUdeviceptr hipDeviceptr_t;
 typedef struct cudaArray hipArray;
+typedef struct cudaArray* hipArray_t;
 typedef struct cudaArray* hipArray_const_t;
 typedef cudaFuncAttributes hipFuncAttributes;
 typedef CUfunction_attribute hipFunction_attribute;


gargrahul:
We already have **typedef struct cudaArray* hipArray_const_t**. I am not sure why we added *const*.

emankov:
> We already have typedef struct cudaArray hipArray_const_t*. I am not sure why we added const.

I'm not sure why we didn't add `const`:
```cpp
typedef struct cudaArray* hipArray_const_t;
```
like in CUDA's driver_types.h:
```cpp
typedef const struct cudaArray *cudaArray_const_t;
```
Just a typo?

gargrahul:
So do we need both hipArray_const_t (already there) and hipArray_t (purpose of this PR)?

emankov:
> So do we need both hipArray_const_t (already there) and hipArray_t (purpose of this PR)?
I think, yes, we need both, but we should aware, that hipArray_const_t is already used in the stack (somehow).

Sarbojit2019:
I could see hipArray_const_t & hipArray_t used in multiple places so if we are realizing that we don't need both then we have to do cleanup where we should replace the one that we feel not required. This PR is just to make CUDA runtime and HIP runtime in sync.

In current codebase under hip/hcc_details/driver_types.h  has "typedef struct hipArray* hipArray_t;" corresponding change is missing for Cuda which this PR is addressing.

Feel free to share your thoughts

Issue #1196, opened 2019-07-01
HIP clang vs. deprecated HCC


tdd11235813:
Can you shed some light on the new compiler workflow on HIP(AMD)?
We currently use the deprecated HCC for our HIP codes on the AMD platform (ROCm 2.5).
By reading hipcc script, it looks like, the new workflow will involve the `/opt/rocm/llvm` as default for HIP clang path, which is currently not present on our ROCm 2.5 installation (ubuntu), but it probably will come with new ROCm? Or is it part of the llvm AMDGPU toolchain?

tdd11235813:
update, WIP: working on compiling hip-clang via [INSTALL.md](https://github.com/ROCm-Developer-Tools/HIP/blob/master/INSTALL.md#hip-clang) (please fix _AMD-clang_ link to _HIP-clang_ in TOC).
Here is the current install script. It compiles, but could not compile hip code with hip-clang yet. It still uses HCC stuff, but it seems it is intended. I'll update when I know more...

__manual install hip-clang__
```bash
#!/bin/bash
# https://github.com/ROCm-Developer-Tools/HIP/blob/master/INSTALL.md#hip-clang

# requirements:
# libelf-dev and rocm libs with hcc (atm still required for compiling HIP)

#BUILD_TYPE="RelWithDebInfo"
BUILD_TYPE="Release"

SRC_DIR=$HOME/sources/hipclang
LLVM_INSTALL_DIR=/opt/rocm/llvm # make sure you have the privileges
HIP_INSTALL_DIR=${LLVM_INSTALL_DIR}/hip # within llvm dir to avoid clash with regular hip installation

mkdir -p $SRC_DIR && cd $SRC_DIR

# Users need to build LLVM, clang, lld, ROCm device library, and HIP from source
# llvm
[ -d "llvm/" ] || git clone --single-branch --recursive -b amd-common https://github.com/RadeonOpenCompute/llvm.git llvm

cd ${SRC_DIR}/llvm/tools
[ -d "clang/" ] || git clone --single-branch --recursive -b amd-common https://github.com/RadeonOpenCompute/clang
[ -d "lld/" ] || git clone --single-branch --recursive -b amd-common https://github.com/RadeonOpenCompute/lld

# build llvm/clang/lld
cd ${SRC_DIR}/llvm
if [ ! -d "build/" ]; then
    mkdir build && cd build
    cmake .. \
          -DLLVM_BUILD_TESTS=OFF \
[...]

PR #1331, opened 2019-08-14
Fix build issues seen with hip-clang path

gargrahul:


gargrahul:
> 
> 
> LGTM, but could you provide issues observed? It's not obvious, you know, from the diff.

@emankov Please refer https://github.com/ROCm-Developer-Tools/HIP/issues/1196#issuecomment-520953315

davidrohr:
I can confirm this fixes the issue of #1196

PR #1338, opened 2019-08-16
Fix missing decl for hip-clang

yxsamliu:
Add back decl for hipHccModuleLaunchKernel and hipExtModuleLaunchKernel for HIP/VDI only

### DIFF
--- a/include/hip/hcc_detail/hip_runtime.h
+++ b/include/hip/hcc_detail/hip_runtime.h
@@ -462,6 +462,25 @@ hc_get_workitem_absolute_id(int dim)
 #undef __CUDA__
 #pragma pop_macro("__CUDA__")
 
+#if __HIP_VDI__
+hipError_t hipExtModuleLaunchKernel(hipFunction_t f, uint32_t globalWorkSizeX,
+                                    uint32_t globalWorkSizeY, uint32_t globalWorkSizeZ,
+                                    uint32_t localWorkSizeX, uint32_t localWorkSizeY,
+                                    uint32_t localWorkSizeZ, size_t sharedMemBytes,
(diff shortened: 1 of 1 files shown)


aaronenyeshi:
https://github.com/ROCm-Developer-Tools/HIP/pull/1330 removed this, but HIP-Clang needs this back.

gargrahul:
Why can't declaration in https://github.com/ROCm-Developer-Tools/HIP/blob/master/include/hip/hip_hcc.h#L94-#L112 be used by hip-clang?

aaronenyeshi:
It can be used if we update: https://github.com/ROCm-Developer-Tools/HIP/blob/master/include/hip/hip_hcc.h#L26
But do we want more hcc named files? (isn't it deprecating?)

yxsamliu:
> Why can't declaration in https://github.com/ROCm-Developer-Tools/HIP/blob/master/include/hip/hip_hcc.h#L94-#L112 be used by hip-clang?

This file is for HCC only by its name. It is not proper to include it for HIP/VDI.

mhbliao:
Before PR#1341 is finalized, shall we merge this first? builds of two major math libs are blocked days and through the weekend, there are concerns more issues may be hidden without successful builds.

yxsamliu:
> Before PR#1341 is finalized, shall we merge this first? builds of two major math libs are blocked days and through the weekend, there are concerns more issues may be hidden without successful builds.

@gargrahul @mangupta Yes we need this urgently. teamcity miopen and rocblas build has been broken for long time

gargrahul:
@yxsamliu Sure. We can merge this for now and can reverse this in #1341 when ready.

aaronenyeshi:
HIP-Clang CI builds require this fix. Thank you

PR #1339, opened 2019-08-16
[hip] add initial implementation for hipLaunchCooperativeKernel API

asalmanp:


src/hip_module.cpp
```diff
@@ -386,6 +386,118 @@ hipError_t hipExtLaunchMultiKernelMultiDevice(hipLaunchParams* launchParamsList,
...
+    void *gwsKernelParams[2];
+    uint nwm1, rid;
+    rid = 0;
+
+    const uint32_t wavefrontSize = 1 << kd->_header->wavefront_size;
+    uint32_t alignedBlockDim = ((blockDimX.x * blockDimX.y * blockDimX.z) + wavefrontSize - 1) &
+                                ~(wavefrontSize - 1);
+
+    // calculate total number of wavefronts minus 1 for the main kernel
+    nwm1 = ((gridDim.x * gridDim.y * gridDim.z) *
```
b-sumner: This barrier should be initialized with the number of work groups minus 1 since that is what the sync is expecting.

asalmanp: changed the initialization of the barrier to the number of work groups.

src/hip_module.cpp
```diff
@@ -386,6 +386,113 @@ hipError_t hipExtLaunchMultiKernelMultiDevice(hipLaunchParams* launchParamsList,
     return ihipLogStatus(result);
 }
 
+namespace {
+// kernel for initializing GWS
+// nwm1 is the total number of workgroups minus 1 and rid is the GWS resource id
+__global__ void init_gws(uint nwm1, uint rid) {
+    __ockl_gws_init(nwm1, rid);
```
b-sumner: This kernel needs to initialize resource 0.   Seems to be a waste of cycles on both host and device to set up two arguments and make this handle an arbitrary resource id.

asalmanp: Removed the second argument for init_gws kernel.

src/hip_module.cpp
```diff
@@ -386,6 +386,107 @@ hipError_t hipExtLaunchMultiKernelMultiDevice(hipLaunchParams* launchParamsList,
...
+
+    HIP_INIT_API(hipLaunchCooperativeKernel, f, gridDim, blockDimX, kernelParams, sharedMemBytes, stream);
+    hipError_t result;
+
+
+    if ((f == nullptr) || (stream == nullptr) || (kernelParams == nullptr)) {
+        return ihipLogStatus(hipErrorNotInitialized);
+    }
+
+    auto ctx = ihipGetTlsDefaultCtx();
```
mangupta: I think lines 409-418 is wrong. We should be using the device associated with the stream for performing the checks and not the default device on this current thread.

src/hip_module.cpp
```diff
@@ -386,6 +386,107 @@ hipError_t hipExtLaunchMultiKernelMultiDevice(hipLaunchParams* launchParamsList,
...
+        stream->criticalData().unlock();
+#if (__hcc_workweek__ >= 19213)
+        stream->criticalData()._av.release_locked_hsa_queue();
+#endif
+
+        return ihipLogStatus(hipErrorLaunchFailure);
+    }
+
+    // launch the main kernel
+    result = ihipModuleLaunchKernel(tls, kd,
```
mangupta: Question: Do we need to do anything special in case launching the init_gws kernel passes but launching the actual kernel fails?

src/hip_module.cpp
asalmanp: Modified the code to use the device associated with the stream for checking the device properties

src/hip_module.cpp
asalmanp: No we don't need to do anything special in this case.

Issue #1186, opened 2019-06-24
Use of uninitialized value $HIP_RUNTIME when compiling with hipcc


Alkeryn:
Whenever i compile using hipcc it throw me these errors:
![image](https://user-images.githubusercontent.com/11599075/60055874-52a61400-96df-11e9-902e-c0b5e228d7e0.png)
Tho it successfully compile and the executable successfully runs

SarbojitAMD:
Thanks for pointing it out raised PR https://github.com/ROCm-Developer-Tools/HIP/pull/1343

cederikhoefs:
(Sorry to open this old thread up again, but I get a
`Use of uninitialized value $HIP_COMPILER in string ne at /opt/rocm/hip/bin//hipcc line 657.`
at each invocation of `hipcc`. `nvcc` is being used as the back end.

PR #1343, opened 2019-08-20
Fix for github issue #1186

SarbojitAMD:
Issue : Warning was thrown while compiling hip code on Nvidia platform as "Use of uninitialized value"   
Root cause : Added a check to make sure value exist before using the value.

Note: Please ignore other commits shown here. Those are showing up as my first few changes I raised under my forked main.


SarbojitAMD:
I have raised fresh hence closing this PR

PR #1346, opened 2019-08-20
Using HSA API for hipMemsetAsync

cjatin:
Using HSA api for better performance

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -1557,6 +1557,13 @@ hipError_t ihipMemPtrGetInfo(void* ptr, size_t* size) {
 
 template <typename T>
 void ihipMemsetKernel(hipStream_t stream, T* ptr, T val, size_t count) {
+    // Just Use count, instead of dividing by 4, the calling API already does it
+    if (sizeof(T) == sizeof(uint32_t) && (count % sizeof(uint32_t) == 0) &&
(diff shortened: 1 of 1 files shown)


src/hip_memory.cpp
```diff
@@ -1557,6 +1557,15 @@ hipError_t ihipMemPtrGetInfo(void* ptr, size_t* size) {
...
 void ihipMemsetKernel(hipStream_t stream, T* ptr, T val, size_t count) {
+    if (sizeof(T) == sizeof(uint32_t) && (count % sizeof(uint32_t) == 0)) {
+        // Just Use count, instead of dividing by 4
+        // The calling API already does it
+        auto err = hsa_amd_memory_fill(ptr, reinterpret_cast<const std::uint32_t&>(val), count);
+        if (!err)
+            return;
+        // Only return if the execution completes without error
+        // if error occured, try the normal version
+    }
```
emankov: LGTM, **thanks**! I'd have just put the comments before code lines with adding _"if aligned"_ in the first one and shrink the code a bit.
```c++
// Just use count if aligned instead of ...
if (sizeof(T) == sizeof(uint32_t) && (count % sizeof(uint32_t) == 0) &&
  !hsa_amd_memory_fill(ptr, reinterpret_cast<const std::uint32_t&>(val), count))
  // Only return if ...
  return;
```

cjatin: Done

PR #1347, opened 2019-08-20
[HIP] Make hipMemcpy*Symbol* APIs C compliant

SarbojitAMD:
- Fixes SWDEV#199247
- Following APIs have been addressed: hipMemcpyToSymbol , hipMemcpyToSymbolAsync, hipMemcpyFromSymbol and hipMemcpyFromSymbolAsync.

### DIFF
--- a/include/hip/hcc_detail/hip_runtime_api.h
+++ b/include/hip/hcc_detail/hip_runtime_api.h
@@ -1514,13 +1514,6 @@ hipError_t hipGetSymbolSize(size_t* size, const void* symbolName) {
 #endif
 
 #ifdef __cplusplus
-namespace hip_impl {
-hipError_t hipMemcpyToSymbol(void*, const void*, size_t, size_t, hipMemcpyKind,
-                             const char*);
-} // Namespace hip_impl.
-#endif
-
-#if defined(__cplusplus)
 extern "C" {
 #endif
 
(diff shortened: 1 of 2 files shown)


SarbojitAMD:
I have raised the same change again as previous one had some extra commits. I have closed that PR

SarbojitAMD:
@scchan, would be great if you can take a look.  Thanks

include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -1587,18 +1614,51 @@ extern "C" {
...
+inline
+__attribute__((visibility("hidden")))
 hipError_t hipMemcpyFromSymbol(void* dst, const void* symbolName,
                                size_t sizeBytes, size_t offset __dparm(0),
-                               hipMemcpyKind kind __dparm(hipMemcpyDeviceToHost));
+                               hipMemcpyKind kind __dparm(hipMemcpyDeviceToHost)) {
+    if (!symbolName) return hipErrorInvalidSymbol;
 
+    hipDeviceptr_t src = NULL;
+    hipGetSymbolAddress(&src, symbolName);
```
gargrahul: What if it fails?

include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -1542,9 +1549,19 @@ extern "C" {
...
+inline
+__attribute__((visibility("hidden")))
 hipError_t hipMemcpyToSymbol(const void* symbolName, const void* src,
                              size_t sizeBytes, size_t offset __dparm(0),
-                             hipMemcpyKind kind __dparm(hipMemcpyHostToDevice));
+                             hipMemcpyKind kind __dparm(hipMemcpyHostToDevice)) {
+    if (!symbolName) return hipErrorInvalidSymbol;
+
+    hipDeviceptr_t dst = NULL;
+    hipGetSymbolAddress(&dst, (const char*)symbolName);
```
gargrahul: What if it fails?

include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -1587,18 +1614,51 @@ extern "C" {
...
                                   size_t sizeBytes, size_t offset,
-                                  hipMemcpyKind kind, hipStream_t stream __dparm(0));
+                                  hipMemcpyKind kind, hipStream_t stream __dparm(0)){
 
+    if (!symbolName) return hipErrorInvalidSymbol;
+
+    hipDeviceptr_t dst = NULL;
+    hipGetSymbolAddress(&dst, symbolName);
+
+    return hip_impl::hipMemcpyToSymbolAsync(dst, src, sizeBytes, offset, kind,
```
gargrahul: Is there a way to avoid nested HIP API calls?

src/hip_memory.cpp
```diff
@@ -1532,6 +1532,57 @@ __global__ void hip_copy2d_n(T* dst, const T* src, size_t width, size_t height,
 }
 }  // namespace
 
+hipError_t hipMemcpyToSymbol(const void* symbolName, const void* src,
```
scchan: If we move the C++ implementation of these functions from hip_runtime_api.h into this .cpp file than then we won't need to duplicate the logic/code like below

Sarbojit2019: Thanks for review,
It seems hipMemcpy*Symbol* was brought to header file in order to fix dlopen issue. More details can be found in PR# https://github.com/ROCm-Developer-Tools/HIP/pull/929. I was not sure if bringing it back to .CPP may break dlopen scenario. Hence for CPP I did not change the behavior and for gcc I thought of using older implementation of these APIs.

Please let me know your thoughts

mangupta:

PR #1350, opened 2019-08-20
Add hiprtc as a dependency of hip_hcc should hiprtc be configured.

whchung:
This allows following sequence work without issue:

```
cmake ..
make -j$(nproc) package
```

Prior to this PR, it has to go like this:

```
cmake ..
make -j$(nproc)
make hiprtc
make package
```


gargrahul:
@whchung is it not getting addressed in #1348  @ https://github.com/ROCm-Developer-Tools/HIP/pull/1348/files#diff-af3b638bc2a3e6c650974192a53c7291R440 ?

whchung:
@gargrahul thanks. not knowing the existence of such PR. close it hereby.

PR #1349, opened 2019-08-20
occupancy.cpp with Makefile

satyanveshd:
Added a new sample occupancy.cpp, where it first measures the performance of a kernel launch with the default block size and next with the optimal block size given by the API hipOccupancyMaxPotentialBlockSize.  After this, the occupancy in terms of max number of active blocks (or warps) per multiprocessor is calculated using the API hipOccupancyMaxActiveBlocksPerMultiprocessor.

PR #1352, opened 2019-08-21
[sample] add new cookbook sample - occupancy

satyanveshd:
Updated the  sample occupancy.cpp and Makefile according to the review comments.

### DIFF
--- a/samples/2_Cookbook/13_occupancy/occupancy.cpp
+++ b/samples/2_Cookbook/13_occupancy/occupancy.cpp
@@ -44,7 +44,7 @@ void multiplyCPU(float* C, float* A, float* B, int N){
     }
 }
 
-void launchKernel(float* C, float* A, float* B, int N, bool manual){
+void launchKernel(float* C, float* A, float* B, bool manual){
      
      hipDeviceProp_t devProp;
      HIP_CHECK(hipGetDeviceProperties(&devProp, 0));
@@ -56,18 +56,20 @@ void launchKernel(float* C, float* A, float* B, int N, bool manual){
      const unsigned threadsperblock = 32;
      const unsigned blocks = (NUM/threadsperblock)+1;
 
+     uint32_t mingridSize = 0;
      uint32_t gridSize = 0;
      uint32_t blockSize = 0;
-      
+     
      if (manual){
 	blockSize = threadsperblock; 
 	gridSize  = blocks;
 	std::cout << std::endl << "Manual Configuration with block size " << blockSize << std::endl;
      }
      else{
-	HIP_CHECK(hipOccupancyMaxPotentialBlockSize(&gridSize, &blockSize, multiply, 0, 0));
+	HIP_CHECK(hipOccupancyMaxPotentialBlockSize(&mingridSize, &blockSize, multiply, 0, 0));
 	std::cout << std::endl << "Automatic Configuation based on hipOccupancyMaxPotentialBlockSize " << std::endl;
-	std::cout << "Suggested blocksize is " << blockSize << ", Minimum gridsize is " << gridSize << std::endl; 
+	std::cout << "Suggested blocksize is " << blockSize << ", Minimum gridsize is " << mingridSize << std::endl;
+	gridSize = (NUM/blockSize)+1; 
      }
 
      // Record the start event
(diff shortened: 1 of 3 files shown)


gargrahul:
Is this the duplicate of #1349? Should #1349 be closed then?

samples/2_Cookbook/13_occupancy/occupancy.cpp
```diff
@@ -0,0 +1,167 @@
...
+     
+     if(devProp.warpSize){
+	activeWarps = numBlock* blockSize/devProp.warpSize;
+	maxWarps = devProp.maxThreadsPerMultiProcessor/devProp.warpSize;
+	std::cout << "Theoretical Occupancy is " << (double)activeWarps/maxWarps * 100 << "%" << std::endl;
+     }	
+    else{
+	 std::cout << "Invalid warpsize" << std::endl;
+	 std::cout << "No. of Active Blocks per multiprocessor returned by hipOccupancyMaxActiveBlocksPerMultiprocessor is " << numBlock << std::endl;
+         exit(0);
```
gargrahul: Better to return an error and exit only after cleaning up.

samples/2_Cookbook/13_occupancy/occupancy.cpp
```diff
@@ -0,0 +1,167 @@
...
+     float eventMs = 1.0f;
+     int activeWarps, maxWarps;
+      
+     if (manual){
+	blockSize = threadsperblock; 
+     	gridSize  = blocks;
+        std::cout << std::endl << "Manual Configuration with block size " << blockSize << std::endl;
+     }
+     else{
+     	hipOccupancyMaxPotentialBlockSize(&gridSize, &blockSize, multiply, 0, 0);
```
gargrahul: Missing HIP_CHECK

samples/2_Cookbook/13_occupancy/occupancy.cpp
```diff
@@ -0,0 +1,167 @@
...
+     // Record the stop event
+     HIP_CHECK(hipEventRecord(stop, NULL));
+     HIP_CHECK(hipEventSynchronize(stop));
+         
+     HIP_CHECK(hipEventElapsedTime(&eventMs, start, stop));
+     printf("kernel Execution time = %6.3fms\n", eventMs);
+
+     //Calculate Occupancy
+     uint32_t numBlock = 0;
+     hipOccupancyMaxActiveBlocksPerMultiprocessor(&numBlock, multiply, blockSize, 0);
```
gargrahul: Missing HIP_CHECK

samples/2_Cookbook/13_occupancy/occupancy.cpp
```diff
@@ -0,0 +1,167 @@
...
+     
+     if(devProp.warpSize){
```
mangupta: warpSize does not impact the final theoretical occupancy calculation since it cancels out. You could simplify this if you wanted to just do (numBlock * blockSize / devProp.maxThreadsPerMultiProcessor). You can check for maxThreadsPerMultiProcessor being 0.

samples/2_Cookbook/13_occupancy/occupancy.cpp
mangupta: Not being able to compute the theoretical occupancy should not be a blocking failure. I suggest that we just drop the else part of the code and continue.

samples/2_Cookbook/13_occupancy/occupancy.cpp
```diff
@@ -0,0 +1,167 @@
...
+*/
+
+#include "hip/hip_runtime.h"
+#include <iostream>
+#define NUM 1000000
+
+const unsigned threadsperblock = 32;
+const unsigned blocks = (NUM/threadsperblock)+1;
+
+uint32_t gridSize = 0;
```
mangupta: We don't need these 2 variables to be global. It should be fine to move the two inside the launchKernel function.

samples/2_Cookbook/13_occupancy/occupancy.cpp
```diff
@@ -0,0 +1,157 @@
...
+     // CPU computation
+     multiplyCPU(cpuC, A, B, NUM);
+
+     //verify the results
+     errors = 0;
+     double eps = 1.0E-6;
+     
+     for (i = 0; i < NUM; i++) {
+	  if (std::abs(C[i] - cpuC[i]) > eps) {
+		  errors++;
```
mangupta: We are effectively only checking if there are errors in the output of the 2nd kernel execution. Maybe we should either move the verification into the launchKernel function or get the kernel output into different buffers for both the kernel runs such as C0d and C1d.

PR #1353, opened 2019-08-21
[HIP] Reclaiming hipLaunchKernel API

SarbojitAMD:
Reclaiming the old deprecated hipLaunchKernel symbol to implement hipLaunchKernel that matches cudaLaunchKernel signature.

Note: Incorporated all review comments

### DIFF
--- a/include/hip/hcc_detail/hip_runtime_api.h
+++ b/include/hip/hcc_detail/hip_runtime_api.h
@@ -1520,6 +1520,35 @@ hipError_t hipMemcpyToSymbol(void*, const void*, size_t, size_t, hipMemcpyKind,
 } // Namespace hip_impl.
 #endif
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief C compliant kernel launch API
+ *
+ * @param [in] function_address - kernel function pointer.
+ * @param [in] numBlocks - number of blocks
+ * @param [in] dimBlocks - dimension of a block
+ * @param [in] args - kernel arguments
+ * @param [in] sharedMemBytes - Amount of dynamic shared memory to allocate for this kernel.  The
+ *  Kernel can access this with HIP_DYNAMIC_SHARED.
+ * @param [in] stream - Stream where the kernel should be dispatched.  May be 0, in which case th
+ *  default stream is used with associated synchronization rules.
+ *
+ * @returns #hipSuccess, #hipErrorInvalidValue, hipInvalidDevice
+ *
+ * @warning should be used from gcc compiled host code only
+ */
+
+hipError_t  hipLaunchKernel(const void* function_address,
+			    dim3 numBlocks, dim3 dimBlocks, void** args,
+				size_t sharedMemBytes, hipStream_t stream);
+
+#ifdef __cplusplus
+}
+#endif
+
 #if defined(__cplusplus)
 extern "C" {
 #endif
--- a/include/hip/hcc_detail/hip_runtime_api.h
+++ b/include/hip/hcc_detail/hip_runtime_api.h
@@ -1538,7 +1538,6 @@ extern "C" {
  *
  * @returns #hipSuccess, #hipErrorInvalidValue, hipInvalidDevice
  *
- * @warning should be used from gcc compiled host code only
  */
 
 hipError_t  hipLaunchKernel(const void* function_address,
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -1144,3 +1144,32 @@ hipError_t hipOccupancyMaxActiveBlocksPerMultiprocessorWithFlags(
     return ihipLogStatus(ihipOccupancyMaxActiveBlocksPerMultiprocessor(
         tls, numBlocks, f, blockSize, dynSharedMemPerBlk));
 }
+
+hipError_t hipLaunchKernel(
+    const void* func_addr, dim3 numBlocks, dim3 dimBlocks, void** args,
+    size_t sharedMemBytes, hipStream_t stream)
+{
+   HIP_INIT_API(hipLaunchKernel,func_addr,numBlocks,dimBlocks,args,sharedMemBytes,stream);
+
+   hipFunction_t kd = hip_impl::get_program_state().kernel_descriptor((std::uintptr_t)func_addr,
+                                                           hip_impl::target_agent(stream));
+   size_t szKernArg = 0;
+
+   if(kd != nullptr && kd->_header != nullptr)
+       szKernArg = kd->_header->kernarg_segment_byte_size;
+
+   void* config[]{
+        HIP_LAUNCH_PARAM_BUFFER_POINTER,
+        args,
+        HIP_LAUNCH_PARAM_BUFFER_SIZE,
+	    &szKernArg,
+        HIP_LAUNCH_PARAM_END};
+
+        if(args == NULL)
+        {
+           config[1] = NULL;
+        }
+
+   return ihipLogStatus(ihipModuleLaunchKernel(tls, kd, numBlocks.x * dimBlocks.x, numBlocks.y * dimBlocks.y, numBlocks.z * dimBlocks.z,
+                          dimBlocks.x, dimBlocks.y, dimBlocks.z, sharedMemBytes, stream, nullptr, (void**)&config, nullptr, nullptr, 0));
+}
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -1153,10 +1153,14 @@ hipError_t hipLaunchKernel(
 
    hipFunction_t kd = hip_impl::get_program_state().kernel_descriptor((std::uintptr_t)func_addr,
                                                            hip_impl::target_agent(stream));
-   size_t szKernArg = 0;
 
-   if(kd != nullptr && kd->_header != nullptr)
-       szKernArg = kd->_header->kernarg_segment_byte_size;
+   if(kd == nullptr || kd->_header == nullptr)
+       return ihipLogStatus(hipErrorInvalidValue);
+
+   size_t szKernArg = kd->_header->kernarg_segment_byte_size;
+
+   if(args == NULL && szKernArg != 0)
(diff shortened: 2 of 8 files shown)


src/hip_module.cpp
```diff
@@ -1144,3 +1144,32 @@ hipError_t hipOccupancyMaxActiveBlocksPerMultiprocessorWithFlags(
...
+    const void* func_addr, dim3 numBlocks, dim3 dimBlocks, void** args,
+    size_t sharedMemBytes, hipStream_t stream)
+{
+   HIP_INIT_API(hipLaunchKernel,func_addr,numBlocks,dimBlocks,args,sharedMemBytes,stream);
+
+   hipFunction_t kd = hip_impl::get_program_state().kernel_descriptor((std::uintptr_t)func_addr,
+                                                           hip_impl::target_agent(stream));
+   size_t szKernArg = 0;
+
+   if(kd != nullptr && kd->_header != nullptr)
```
gargrahul: Add else case to return error.

tests/src/gcc/LaunchKernel.c
```diff
@@ -0,0 +1,180 @@
...
+  int A = 0;
+  int *A_d = NULL;
+  dim3 blocks       = {1,1,1};
+  dim3 threads      = {1,1,1};
+
+  printf(" -----: Launch One Argument Kernel :------ \n");
+
+  // Allocate Device memory
+  status = hipMalloc((void**)&A_d, sizeof(int));
+  printf("hipMalloc status %s\n", hipGetErrorString(status));
```
gargrahul: No need to print status of every HIP call. you can use HIPCHECK as used in other dtests.

tests/src/gcc/LaunchKernel.c
```diff
@@ -0,0 +1,180 @@
...
+
+  void* Args[]={A_d};
+  hipLaunchKernel(kernel1, blocks, threads, Args,0,0);
```
gargrahul: Check return error using HIPCHECK.

tests/src/gcc/LaunchKernel.c
```diff
@@ -0,0 +1,180 @@
...
+  printf("hipLaunchKernel status %s\n",hipGetErrorString(status));
+
+  // Get the result back to host memory
+  status = hipMemcpy(&A, A_d, sizeof(int), hipMemcpyDeviceToHost);
```
gargrahul: Check return error using HIPCHECK.

tests/src/gcc/LaunchKernel.c
```diff
@@ -0,0 +1,180 @@
...
+  printf("hipMalloc status %s\n", hipGetErrorString(status));
+
+  if(A == 333)
+    printf("Test Passed!!\n\n");
+  else
+    printf("Test Failed :Value of A is %d\n\n",A);
+
+  hipFree(A_d);
```
gargrahul: Check return error using HIPCHECK.

tests/src/gcc/LaunchKernel.c
```diff
@@ -0,0 +1,180 @@
...
+  int *A_d = NULL;
+  int *B_d = NULL;
+
+  dim3 blocks       = {1,1,1};
+  dim3 threads      = {1,1,1};
+
+  printf(" -----: Launch Two Argument Kernel :------ \n");
+
+  // Allocate Device memory
+  status = hipMalloc((void**)&A_d, sizeof(int));
```
gargrahul: Check return error using HIPCHECK.

tests/src/gcc/LaunchKernel.c
```diff
@@ -0,0 +1,180 @@
...
+  printf("hipMalloc status %s\n", hipGetErrorString(status));
+
+  status = hipMalloc((void**)&B_d, sizeof(int));
```
gargrahul: Check return error using HIPCHECK.

tests/src/gcc/LaunchKernel.c
```diff
@@ -0,0 +1,180 @@
...
+  printf("hipMalloc status %s\n", hipGetErrorString(status));
+
+  // Copy data from host memory to device memory
+  status = hipMemcpy(B_d,&B, sizeof(int), hipMemcpyHostToDevice);
```
gargrahul: Check return error using HIPCHECK.

tests/src/gcc/LaunchKernel.c
```diff
@@ -0,0 +1,180 @@
...
+  printf("hipMalloc status %s\n", hipGetErrorString(status));
+
+
+  void* Args[]={A_d,B_d};
+  hipLaunchKernel(kernel2, blocks, threads, Args,0,0);
```
gargrahul: Check return error using HIPCHECK.

tests/src/gcc/LaunchKernel.c
```diff
@@ -0,0 +1,180 @@
...
+
+  printf(" -----: Launch One Argument Kernel :------ \n");
+
+  // Allocate Device memory
+  status = hipMalloc((void**)&A_d, sizeof(int));
+  printf("hipMalloc status %s\n", hipGetErrorString(status));
+
+  void* Args[]={A_d};
+  hipLaunchKernel(kernel1, blocks, threads, Args,0,0);
+  printf("hipLaunchKernel status %s\n",hipGetErrorString(status));
```
gargrahul: Are you really checking the status of hipLaunchKernel ?

tests/src/gcc/LaunchKernel.c
```diff
@@ -0,0 +1,180 @@
...
+  printf("hipMalloc status %s\n", hipGetErrorString(status));
+
+  // Copy data from host memory to device memory
+  status = hipMemcpy(B_d,&B, sizeof(int), hipMemcpyHostToDevice);
+  printf("hipMalloc status %s\n", hipGetErrorString(status));
+
+
+  void* Args[]={A_d,B_d};
+  hipLaunchKernel(kernel2, blocks, threads, Args,0,0);
+  printf("hipLaunchKernel status %s\n", hipGetErrorString(status));
```
gargrahul: Status being checked is not from hipLaunchKernel.

tests/src/gcc/LaunchKernel.c
```diff
@@ -0,0 +1,180 @@
...
+
+  // Get the result back to host memory
+  status = hipMemcpy(&A, A_d, sizeof(int), hipMemcpyDeviceToHost);
```
gargrahul: Check return error using HIPCHECK

tests/src/gcc/LaunchKernel.c
```diff
@@ -0,0 +1,180 @@
...
+
+
+int main()
+{
+  LaunchKernelArg();
+  LaunchKernelArg1();
+  LaunchKernelArg2();
+  LaunchKernelArg3();
+
+  printf("PASSED!\n");
```
gargrahul: Will this test always PASS?

tests/src/gcc/LaunchKernel.c
```diff
@@ -0,0 +1,180 @@
...
+  status = hipMemcpy(A_d,&A, sizeof(int), hipMemcpyHostToDevice);
+  printf("hipMalloc status %s\n", hipGetErrorString(status));
+
+  status = hipMemcpy(B_d,&B, sizeof(int), hipMemcpyHostToDevice);
+  printf("hipMalloc status %s\n", hipGetErrorString(status));
+
+
+  void* Args[]={A_d,B_d,C_d};
+  hipLaunchKernel(kernel3, blocks, threads, Args,0,0);
+  printf("hipLaunchKernel status %s\n", hipGetErrorString(status));
```
gargrahul: Is status being checked from hipLaunchKernel ?

src/hip_module.cpp
```diff
@@ -1144,3 +1144,32 @@ hipError_t hipOccupancyMaxActiveBlocksPerMultiprocessorWithFlags(
...
+   void* config[]{
+        HIP_LAUNCH_PARAM_BUFFER_POINTER,
+        args,
+        HIP_LAUNCH_PARAM_BUFFER_SIZE,
+	    &szKernArg,
+        HIP_LAUNCH_PARAM_END};
+
+        if(args == NULL)
+        {
+           config[1] = NULL;
```
mangupta: Is this required? If args is NULL, then config[1] should already be set to NULL as per line 1163. Also it might be useful to check if args is NULL but szKernArg is not 0 and return an error appriopriately.

tests/src/gcc/LaunchKernel.c
mangupta: The right way to do it is that each of the LaunchKernelArg* functions returns a bool based on pass or fail. You declare a bool in main whose initial value is true. And you AND the return value from each of these LaunchKernelArg functions. Then you check the final return value to determine PASS or FAIL.

include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -1520,6 +1520,35 @@ hipError_t hipMemcpyToSymbol(void*, const void*, size_t, size_t, hipMemcpyKind,
...
+ * @param [in] dimBlocks - dimension of a block
+ * @param [in] args - kernel arguments
+ * @param [in] sharedMemBytes - Amount of dynamic shared memory to allocate for this kernel.  The
+ *  Kernel can access this with HIP_DYNAMIC_SHARED.
+ * @param [in] stream - Stream where the kernel should be dispatched.  May be 0, in which case th
+ *  default stream is used with associated synchronization rules.
+ *
+ * @returns #hipSuccess, #hipErrorInvalidValue, hipInvalidDevice
+ *
+ * @warning should be used from gcc compiled host code only
```
mangupta: I am not sure why is the API restricted to being used from gcc only? It should be fine to use this API from g++/hcc/hip-clang/nvcc compiled code as well.

SarbojitAMD:
@gargrahul 
In current change Host code will be compiled using gcc but test_common.h is not C complaint hence I could not use it.

PR #1356, opened 2019-08-21
[samples] Change -g to -O3 and use CXXFLAGS in HIP cookbook samples

gargrahul:


emankov:
What about common Makefile?

gargrahul:
> 
> 
> What about common Makefile?

Which common Makefile?

mangupta:
> > What about common Makefile?
> 
> Which common Makefile?

I think what @emankov is suggesting is to create a common makefile that can be used by all the samples.

samples/2_Cookbook/0_MatrixTranspose/Makefile
```diff
@@ -17,12 +17,11 @@ EXECUTABLE=./MatrixTranspose
...
 all: $(EXECUTABLE) test
 
-CXXFLAGS =-g
-CXX=$(HIPCC)
+CXXFLAGS =-O3
 
 
 $(EXECUTABLE): $(OBJECTS)
-	$(HIPCC) $(OBJECTS) -o $@
+	$(HIPCC) $(CXXFLAGS)  $(OBJECTS) -o $@
```
mangupta: Never noticed this before since the samples were building fine. But we don't seem to have a rule for building the OBJECTS. So either we add that or drop the OBJECTS completely from the makefile.

gargrahul:
> 
> 
> > > What about common Makefile?
> > 
> > 
> > Which common Makefile?
> 
> I think what @emankov is suggesting is to create a common makefile that can be used by all the samples.

Ok, we can do that however I think individual makefile was used so that each sample is independent.

emankov:
> > > > What about common Makefile?
> > > 
> > > 
> > > Which common Makefile?
> > 
> > 
> > I think what @emankov is suggesting is to create a common makefile that can be used by all the samples.
> 
> Ok, we can do that however I think individual makefile was used so that each sample is independent.

Individual makefiles, containing only test specific things (if any) + single up one level common Makefile, which is included in all individual makefiles. Every test might either be built separately or together.

mangupta:

PR #1358, opened 2019-08-22
Set device before processing each one

chrispaquot:
If we don't call hipSetDevice(i), the total/free memory will be wrong.

### DIFF
--- a/samples/1_Utils/hipInfo/hipInfo.cpp
+++ b/samples/1_Utils/hipInfo/hipInfo.cpp
@@ -192,6 +192,7 @@ int main(int argc, char* argv[]) {
     HIPCHECK(hipGetDeviceCount(&deviceCnt));
 
     for (int i = 0; i < deviceCnt; i++) {
+        hipSetDevice(i);
         printDeviceProp(i);
     }
 


gargrahul:
@chrispaquot is this needed for HIP/VDI path? For HIP/HCC we don't need to set device in this sample.

chrispaquot:
> @chrispaquot is this needed for HIP/VDI path? For HIP/HCC we don't need to set device in this sample.
> 
> https://github.com/ROCm-Developer-Tools/HIP/blob/4e5ba72be8ad6c787bfab0785d31faeeb34f64eb/samples/1_Utils/hipInfo/hipInfo.cpp#L81
> 
> already takes the deviceID as input.

Sure but hipMemGetInfo uses the current device to get the info from. And right now hipSetDevice isn't called at all so it always returns the data from the device 0.

gargrahul:
@mangupta Would you like to move setdevice call to just before  https://github.com/ROCm-Developer-Tools/HIP/blob/4e5ba72be8ad6c787bfab0785d31faeeb34f64eb/samples/1_Utils/hipInfo/hipInfo.cpp#L175

or is it fine to set as set in this PR ?

PR #1368, opened 2019-08-26
[hip] add initial support for hipLaunchCooperativeKernelMultiDevice API

asalmanp:


src/hip_module.cpp
```diff
@@ -480,6 +480,116 @@ hipError_t hipLaunchCooperativeKernel(const void* f, dim3 gridDim,
...
+
+        result = ihipModuleLaunchKernel(tls, gwsKds[i], 1, 1, 1, 1, 1, 1,
+                0, lp.stream, gwsKernelParam, nullptr, nullptr, nullptr, 0, true);
+
+        if (result != hipSuccess) {
+            for (int j = 0; j < numDevices; ++j) {
+                launchParamsList[j].stream->criticalData().unlock();
+#if (__hcc_workweek__ >= 19213)
+                launchParamsList[j].stream->criticalData()._av.release_locked_hsa_queue();
+#endif
```
emankov: Will it work correctly without releasing for `__hcc_workweek__ < 19213`?

src/hip_module.cpp
```diff
@@ -480,6 +480,116 @@ hipError_t hipLaunchCooperativeKernel(const void* f, dim3 gridDim,
     return ihipLogStatus(result);
 }
 
+hipError_t hipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsList,
+        int  numDevices, unsigned int  flags) {
+
+    HIP_INIT_API(hipLaunchCooperativeKernelMultiDevice, launchParamsList, numDevices, flags);
+    hipError_t result;
+
+    if ((numDevices > g_deviceCnt) || (launchParamsList == nullptr)) {
```
emankov: Formatting: 2 couples of excessive brackets in if condition.

src/hip_module.cpp
```diff
@@ -480,6 +480,116 @@ hipError_t hipLaunchCooperativeKernel(const void* f, dim3 gridDim,
...
+
+    for (int i = 0; i < numDevices; ++i) {
+        if (!launchParamsList[i].stream->getDevice()->_props.cooperativeMultiDeviceLaunch) {
+            return ihipLogStatus(hipErrorInvalidConfiguration);
+        }
+    }
+
+    hipFunction_t* gwsKds = reinterpret_cast<hipFunction_t*>(malloc(sizeof(hipFunction_t) * numDevices));
+    hipFunction_t* kds    = reinterpret_cast<hipFunction_t*>(malloc(sizeof(hipFunction_t) * numDevices));
+    if ((kds == nullptr) || (gwsKds == nullptr)) {
```
emankov: Formatting: 2 couples of excessive brackets in if condition.

src/hip_module.cpp
```diff
@@ -480,6 +480,116 @@ hipError_t hipLaunchCooperativeKernel(const void* f, dim3 gridDim,
...
+                lp.gridDim.y * lp.blockDim.y,
+                lp.gridDim.z * lp.blockDim.z,
+                lp.blockDim.x, lp.blockDim.y,
+                lp.blockDim.z, lp.sharedMem,
+                lp.stream, lp.args, nullptr, nullptr, nullptr, 0,
+                true);
+    }
+
+    // unlock all streams
+    for (int i = 0; i < numDevices; ++i) {
```
emankov: What about locking / launching / unlocking in a single loop?

asalmanp:
Ping!

src/hip_module.cpp
```diff
@@ -480,6 +480,116 @@ hipError_t hipLaunchCooperativeKernel(const void* f, dim3 gridDim,
...
+            for (int j = 0; j < numDevices; ++j) {
+                launchParamsList[j].stream->criticalData().unlock();
+#if (__hcc_workweek__ >= 19213)
+                launchParamsList[j].stream->criticalData()._av.release_locked_hsa_queue();
+#endif
+            }
+            return ihipLogStatus(hipErrorLaunchFailure);
+        }
+
+        result = ihipModuleLaunchKernel(tls, kds[i],
```
mangupta: What happens if gwsKds is successful on device0, device1,.. but fails on deviceN? We would end up enqueuing the actual kernels on some devices before returning LaunchFailure. Is that fine?

asalmanp: I think this scenario is less likely since the checks for gwsKds is done before the launch and even if it happens for whatever reason, not much can be done for already enqueued main kernels and only an error is returned to the caller notifying the launch failure for the API.
We could also have similar issue if the gwsKds is successful on all devices but main kernel is successful only on device0, device1, .. but fails on deviceN. in this case a launch failure error is similarly reported to the caller and the API call still fails..

PR #1373, opened 2019-08-29
Added hipLaunchKernel API into hipify

SarbojitAMD:


SarbojitAMD:
@emankov, thanks for pointing it out. I think in that case we don't need any change. Will close this PR.

PR #1379, opened 2019-08-30
remove dependencies on /opt/rocm path

srinivamd:
Made changes to the hipcc script to remove default settings to use derived paths based on hipcc location instead of hardcode "/opt/rocm".
Tested it to compile a simple hip code from a non-standard rocm installation.

### DIFF
--- a/bin/hipcc
+++ b/bin/hipcc
@@ -22,8 +22,10 @@ use Cwd 'abs_path';
 # Other environment variable controls:
 # HIP_PATH  : Path to HIP directory, default is one dir level above location of this script
 # CUDA_PATH : Path to CUDA SDK (default /usr/local/cuda). Used on NVIDIA platforms only.
-# HCC_HOME  : Path to HCC SDK (default /opt/rocm/hcc).  Used on AMD platforms only.
-# HSA_PATH  : Path to HSA dir (default /opt/rocm/hsa).  Used on AMD platforms only.
+# HCC_HOME  : Path to HCC SDK (defaults to ../../hcc relative to this
+#             script's abs_path). Used on AMD platforms only.
+# HSA_PATH  : Path to HSA dir (defaults to ../../hsa relative to abs_path
+#             of this script). Used on AMD platforms only.
 # HIP_VDI_HOME : Path to HIP/VDI directory. Used on AMD platforms only.
 
 if(scalar @ARGV == 0){
@@ -58,7 +60,18 @@ $isWindows = $^O eq 'MSWin32';
 $HIPCC_COMPILE_FLAGS_APPEND=$ENV{'HIPCC_COMPILE_FLAGS_APPEND'};
 $HIPCC_LINK_FLAGS_APPEND=$ENV{'HIPCC_LINK_FLAGS_APPEND'};
 
-$HIP_PATH=$ENV{'HIP_PATH'} // dirname (dirname $0); # use parent directory of hipcc
+#
+# TODO: Fix rpath LDFLAGS settings
+#
+# Since this hipcc script gets installed at two uneven hierarchical levels,
+# linked by symlink, the absolute path of this script should be used to
+# derive HIP_PATH, as dirname $0 could be /opt/rocm/bin or /opt/rocm/hip/bin
+# depending on how it gets invoked.
+# HIP_PATH points to <rocm_install_dir>/hip. 
+# Defining ROCM_ROOTDIR to be used instead of hardcoded /opt/rocm defaults
+#
+$HIP_PATH=$ENV{'HIP_PATH'} // dirname(Cwd::abs_path("$0/../")); # use parent directory of hipcc
+$ROCM_ROOTDIR=dirname("$HIP_PATH"); # use parent directory of HIP_PATH
 $HIP_VDI_HOME=$ENV{'HIP_VDI_HOME'};
 $HIP_LIB_PATH=$ENV{'HIP_LIB_PATH'};
 $HIP_CLANG_PATH=$ENV{'HIP_CLANG_PATH'};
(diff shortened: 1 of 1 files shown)


srinivamd:
@mangupta Your comment appears to be about the existing code. The code fix submitted addresses the ugliness of hipcc installed in uneven hierarchical position. Please give it a test run.
Lines 66 + comments added describes the situation.
So the question of whether this script gets a symlink in different places is moot as long as the abs_path of this is under <rocm>/hip/bin.
Maybe I misreading your comment? Please copy the script to a standard installation or non-standard one, and give it a test run by invoking <rocm>/hip/bin/hipcc and symlink from <rocm>/bin/hipcc. I tested that and it worked fine but maybe I missed some case.

srinivamd:
@mangupta Please let me know what changes you think are necessary.

bin/hipcc
```diff
@@ -58,7 +60,18 @@ $isWindows = $^O eq 'MSWin32';
...
+# TODO: Fix rpath LDFLAGS settings
+#
+# Since this hipcc script gets installed at two uneven hierarchical levels,
+# linked by symlink, the absolute path of this script should be used to
+# derive HIP_PATH, as dirname $0 could be /opt/rocm/bin or /opt/rocm/hip/bin
+# depending on how it gets invoked.
+# HIP_PATH points to <rocm_install_dir>/hip. 
+# Defining ROCM_ROOTDIR to be used instead of hardcoded /opt/rocm defaults
+#
+$HIP_PATH=$ENV{'HIP_PATH'} // dirname(Cwd::abs_path("$0/../")); # use parent directory of hipcc
```
yxsamliu: There is a soft link /opt/rocm/bin/hipcc -> /opt/rocm/hip/bin/hipcc. In many cases hipcc is called through /opt/rocm/bin/hipcc instead of /opt/rocm/hip/bin/hipcc. You need to be able to handle this situation. Probably follow the soft link if hipcc is a soft link?

scchan: I abs_path would resolve soft link and return the canonical path

srinivamd: Yes, @scchan answered your question @yxsamliu 
I have tested it from both places. Please NOTE: the real file MUST exist under /opt/roc/hip/bin/hipcc. You can have symlinks from anywhere to that.

bin/hipcc
```diff
@@ -215,7 +228,7 @@ if ($HIP_PLATFORM eq "clang") {
     }
 
     if ($HIP_RUNTIME eq "HCC" ) {
-      $HSA_PATH=$ENV{'HSA_PATH'} // "/opt/rocm/hsa";
+      $HSA_PATH=$ENV{'HSA_PATH'} // "$HIP_PATH/../hsa";
```
mangupta: Probably want ```$ROCM_ROOTDIR/hsa``` here.

bin/hipcc
```diff
@@ -224,17 +237,17 @@ if ($HIP_PLATFORM eq "clang") {
     if (! defined $HIP_LIB_PATH) {
         $HIP_LIB_PATH = "$HIP_PATH/lib";
     }
-    $HSA_PATH=$ENV{'HSA_PATH'} // "/opt/rocm/hsa";
+    $HSA_PATH=$ENV{'HSA_PATH'} // "$HIP_PATH/../hsa";
```
mangupta: ```$ROCM_ROOTDIR/hsa```

bin/hipcc
```diff
@@ -224,17 +237,17 @@ if ($HIP_PLATFORM eq "clang") {
...
 
-    $HCC_HOME=$ENV{'HCC_HOME'} // $hipConfig{'HCC_HOME'} // "/opt/rocm/hcc";
+    $HCC_HOME=$ENV{'HCC_HOME'} // $hipConfig{'HCC_HOME'} // "$HIP_PATH/../hcc";
```
mangupta: Probably want ```$ROCM_ROOTDIR/hcc``` here.

srinivamd: Not needed.
Please integrate this pull request if there are no technical issues.
This is a MUST to support multiple ROCm releases installed side-by-side. Delaying this to use favorite variables names should be avoided IMHO.

bin/hipcc
srinivamd: No need to use ROCM_ROOTDIR as it is an invented variable. HIP_PATH is better, mentioned in HIP readme/docs. Let's stick to the point of removing hardcoded "/opt/rocm" from code.

bin/hipcc
srinivamd: Why? What's wrong with $HIP_PATH/../hsa? You don't a separate variable for every directory.
The point of this fix is to not hardcode "/opt/rocm". This is not about inventing variable names for every directory or item used.

gargrahul: In my understanding , as per current code, $HIP_PATH/../hsa need not be valid always. I can clone and build and install HIP - standalone- at any path today and set that as HIP_PATH.

srinivamd: @gargrahul Unfortunately, "your understanding" is not addressed by this PR. Your scenario will get screwed whether $ROCM_ROOTDIR (which is derived from $HIP_PATH) is used or not.
Let's keep it simple.

gargrahul: "My understanding" is captured here - https://github.com/ROCm-Developer-Tools/HIP/blob/master/INSTALL.md#building-hip-from-source- That's how HIP is being used for now.

I am NOT against keeping it simple though this needs to be thought are we going away where can we build and install HIP?

srinivamd: The changes made are consistent with the web link as the default locations of hip and hsa are preserved. Please read the default paths and environment variables section under AMD-hcc.
Regarding,  "going away where can we build and install HIP" , I have no idea what you are talking about as that is orthogonal to this change. Once again, I ask we keep it simple.

srinivamd: Rather than conjuring up what-if scenarious, please try the code changes and report problems that the original code addressed better.

PR #1392, opened 2019-09-05
Disable name mangling for hipRTC symbols

saleelk:


saleelk:
This isnt needed anymore.

PR #1397, opened 2019-09-06
SWDEV-202043 hipMemcpyDtoD() issue

satyanveshd:
Added Null check in hipMemcpyDtoD(), hipMemcpyHtoD(), hipMemcpyDtoH(), hipMemcpyHtoH() to avoid seg fault when src or dest is passed as NULL.

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -1166,6 +1166,11 @@ hipError_t hipMemcpyHtoD(hipDeviceptr_t dst, void* src, size_t sizeBytes) {
 
     hipError_t e = hipSuccess;
 
+    if(dst==NULL || src==NULL)
+	{
+	e=hipErrorInvalidValue;
+	return ihipLogStatus(e);
+	}
     try {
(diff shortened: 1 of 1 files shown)


src/hip_memory.cpp
```diff
@@ -1166,6 +1166,11 @@ hipError_t hipMemcpyHtoD(hipDeviceptr_t dst, void* src, size_t sizeBytes) {
 
     hipError_t e = hipSuccess;
 
+    if(dst==NULL || src==NULL)
+	{
+	e=hipErrorInvalidValue;
+	return ihipLogStatus(e);
+	}
```
emankov: Here and below: 1. formatting: left parenthesis should be placed on the same line as 'if' statement; 2. Excessive assignment.

satyanveshd: I didn't get the 2nd point. can u please let me know.

satyanveshd:
Added the check if sizeBytes 0, return hipSuccess,  and updated the PR.

PR #1424, opened 2019-09-17
Fix PCI Domain ID query

gargrahul:
Requires not yet publicly released ROCr agent attribute.


gargrahul:
@wenkaidu Could you approve this PR ? Since it is working as tested internally , however it will continue to fail for ROCm2.7 in CI.

src/hip_hcc.cpp
```diff
@@ -782,10 +782,12 @@ hipError_t ihipDevice_t::initProperties(hipDeviceProp_t* prop) {
     DeviceErrorCheck(err);
 
     // BDFID is 16bit uint: [8bit - BusID | 5bit - Device ID | 3bit - Function/DomainID]
```
emankov: Are the last 3bits of BDFID still Function/Domain? Aren't BusID and DeviceID inquired by a separate HSA API call too?

PR #1425, opened 2019-09-17
[HIP] Fixed hipStreamAddCallback [SWDEV#165185]

SarbojitAMD:
1. Fixed hipStreamAddCallback() as requested in SWDEV#165185
2. Added unit test to test the behavior

### DIFF
--- a/src/hip_hcc.cpp
+++ b/src/hip_hcc.cpp
@@ -1501,6 +1501,9 @@ bool ihipStreamCallbackHandler(hsa_signal_value_t value, void* cbArgs) {
 
     ihipStreamCallback_t* cb = static_cast<ihipStreamCallback_t*> (cbArgs);
 
+    if(cb->comFuture.valid())
+        cb->comFuture.wait();
+
     // Call registered callback function
     cb->_callback(cb->_stream, e, cb->_userData);
 
--- a/src/hip_hcc.cpp
+++ b/src/hip_hcc.cpp
@@ -1496,18 +1496,18 @@ hipError_t ihipStreamSynchronize(TlsData *tls, hipStream_t stream) {
     return e;
 }
 
-void ihipStreamCallbackHandler(ihipStreamCallback_t* cb) {
+bool ihipStreamCallbackHandler(hsa_signal_value_t value, void* cbArgs) {
     hipError_t e = hipSuccess;
 
-    // Synchronize stream
-    tprintf(DB_SYNC, "ihipStreamCallbackHandler wait on stream %s\n",
-            ToString(cb->_stream).c_str());
-    GET_TLS();
-    e = ihipStreamSynchronize(tls, cb->_stream);
+    ihipStreamCallback_t* cb = static_cast<ihipStreamCallback_t*> (cbArgs);
 
     // Call registered callback function
     cb->_callback(cb->_stream, e, cb->_userData);
+
+    hsa_signal_store_screlease(cb->_signal,0);
+
     delete cb;
+    return false;
 }
 
 //---
--- a/src/hip_stream.cpp
+++ b/src/hip_stream.cpp
@@ -315,8 +315,11 @@ hipError_t hipStreamAddCallback(hipStream_t stream, hipStreamCallback_t callback
         stream = device->_defaultStream;
     }
 
+    // Lock the stream
+    LockedAccessor_StreamCrit_t crit(stream->criticalData(), true /*unlock at destruction*/);
+
     // 1. Lock the queue
-    hsa_queue_t* lockedQ = static_cast<hsa_queue_t*> (stream->criticalData()._av.acquire_locked_hsa_queue());
+    hsa_queue_t* lockedQ = static_cast<hsa_queue_t*> (crit->_av.acquire_locked_hsa_queue());
 
     if(lockedQ == nullptr)
     {
@@ -330,6 +333,7 @@ hipError_t hipStreamAddCallback(hipStream_t stream, hipStreamCallback_t callback
 
     if(status != HSA_STATUS_SUCCESS)
     {
+        crit->_av.release_locked_hsa_queue();
         return ihipLogStatus(hipErrorInvalidValue);
     }
 
@@ -338,13 +342,15 @@ hipError_t hipStreamAddCallback(hipStream_t stream, hipStreamCallback_t callback
 
     if(status != HSA_STATUS_SUCCESS)
     {
+        crit->_av.release_locked_hsa_queue();
         return ihipLogStatus(hipErrorInvalidValue);
     }
 
-    // 3. Store callback details
+    // 3. Store callback details, will destroy allocation in callback handler
     ihipStreamCallback_t* cb = new ihipStreamCallback_t(stream, callback, userData);
     if(cb == nullptr)
     {
+        crit->_av.release_locked_hsa_queue();
         return ihipLogStatus(hipErrorMemoryAllocation);
     }
     cb->_signal = depSignal;
@@ -375,7 +381,7 @@ hipError_t hipStreamAddCallback(hipStream_t stream, hipStreamCallback_t callback
     hsa_signal_store_relaxed(lockedQ->doorbell_signal, index+1);
 
     // 7. Release queue
-    stream->criticalData()._av.release_locked_hsa_queue();
+    crit->_av.release_locked_hsa_queue();
 
     // 8. Register signal callback
     hsa_amd_signal_async_handler(signal, HSA_SIGNAL_CONDITION_EQ, 0, ihipStreamCallbackHandler, cb);
--- a/src/hip_stream.cpp
+++ b/src/hip_stream.cpp
@@ -309,8 +309,13 @@ hipError_t hipStreamAddCallback(hipStream_t stream, hipStreamCallback_t callback
         stream = device->_defaultStream;
     }
 
+    stream = ihipSyncAndResolveStream(stream);
+
     // Lock the stream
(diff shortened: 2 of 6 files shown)


src/hip_stream.cpp
```diff
@@ -254,14 +254,101 @@ hipError_t hipStreamGetPriority(hipStream_t stream, int* priority) {
...
+    hsa_queue_store_write_index_screlease(lockedQ, nextIndex);
+    hsa_signal_store_relaxed(lockedQ->doorbell_signal, index+1);
+
+    // 7. Release queue
+    stream->criticalData()._av.release_locked_hsa_queue();
+
+    // 8. Register signal callback
+    hsa_amd_signal_async_handler(signal, HSA_SIGNAL_CONDITION_EQ, 0, ihipStreamCallbackHandler, cb);
 
     return ihipLogStatus(e);
```
emankov: Will always return `hipSuccess`.

src/hip_stream.cpp
```diff
@@ -254,14 +254,101 @@ hipError_t hipStreamGetPriority(hipStream_t stream, int* priority) {
...
-    // Create a thread in detached mode to handle callback
+    // 1. Lock the queue
+    hsa_queue_t* lockedQ = static_cast<hsa_queue_t*> (stream->criticalData()._av.acquire_locked_hsa_queue());
+
+    // 2. Allocate a singals
+    hsa_signal_t signal;
+    hsa_status_t status = hsa_signal_create(1, 0, NULL, &signal);
+
+    hsa_signal_t depSignal;
+    status = hsa_signal_create(1, 0, NULL, &depSignal);
```
emankov: Here and below: no status checks. Please provide them and their translation to HIP's ones with corresponding returns.

src/hip_hcc.cpp
```diff
@@ -1496,18 +1496,18 @@ hipError_t ihipStreamSynchronize(TlsData *tls, hipStream_t stream) {
...
-    e = ihipStreamSynchronize(tls, cb->_stream);
+    ihipStreamCallback_t* cb = static_cast<ihipStreamCallback_t*> (cbArgs);
 
     // Call registered callback function
     cb->_callback(cb->_stream, e, cb->_userData);
+
+    hsa_signal_store_screlease(cb->_signal,0);
+
     delete cb;
+    return false;
```
emankov: Where `hipError_t e` is handled? Should the function unconditionally return `false`?

src/hip_stream.cpp
```diff
@@ -254,14 +254,101 @@ hipError_t hipStreamGetPriority(hipStream_t stream, int* priority) {
...
 
 //---
+hipError_t setCallbackPacket(hsa_queue_t* queue,
+                             uint64_t& index, uint64_t& nextIndex,
+                             hsa_barrier_and_packet_t** barrier1,
+                             hsa_barrier_and_packet_t** barrier2){
+
+    uint64_t tempIndex = 0;
+    uint32_t mask = queue->size - 1;
+    hsa_barrier_and_packet_t* tempBarrier;
```
emankov: nullptr initialization.

tests/src/runtimeApi/stream/StreamAddCallback.cpp
```diff
@@ -0,0 +1,117 @@
...
+ * HIT_END
+ */
+
+struct UserData
+{
+    size_t size;
+    int* ptr;
+};
+
+void myCallback(hipStream_t stream, hipError_t status, void* user_data)
```
jlgreathouse: Is this actually automatically testing anything? I appreciate that this is likely visually showing that the two printfs in the callback should happen before anything after the streamsync, but the expected behavior is not described anywhere in the test code, and the test itself does not ensure that this happens. If the callback fails to actually block the streamsync, this test will pass and no one will notice until they go back and specifically look at the order of prints (assuming they know this is what they're looking for).

I think this test could be augmented by having the functions atomically increment some global variables, and ensuring that the resulting values do not indicate that the functions ran out of order.

For instance:
a global atomic_int callback_val.
At the top of test(), atomic_set(&callback_val, 0)

Before the first printf and after the second printf in the callback, fetch_add(&callback_val, 1)

After the hipDeviceSynchronize in test(), do an atomic_load(&callback) and verify that the value is 2. If it's <2, then that means that you have passed the streamsync without having completed the entirety of the callback.

This would allow you to return false from the test if the callback doesn't function correctly, not just if the copies fail for some reason.

src/hip_stream.cpp
```diff
@@ -254,14 +254,101 @@ hipError_t hipStreamGetPriority(hipStream_t stream, int* priority) {
 
 
 //---
+hipError_t setCallbackPacket(hsa_queue_t* queue,
+                             uint64_t& index, uint64_t& nextIndex,
+                             hsa_barrier_and_packet_t** barrier1,
+                             hsa_barrier_and_packet_t** barrier2){
+
+    uint64_t tempIndex = 0;
+    uint32_t mask = queue->size - 1;
```
jlgreathouse: I'm not 100% up to date on the HIP runtime's coding style, but should you verify that queue != null before dereferencing it?

src/hip_stream.cpp
```diff
@@ -254,14 +254,101 @@ hipError_t hipStreamGetPriority(hipStream_t stream, int* priority) {
...
+    hsa_barrier_and_packet_t* tempBarrier;
+
+    // Check for empty barrier packets
+    do{
+
```
jlgreathouse: Blank space?

src/hip_hcc.cpp
SarbojitAMD: Yes, function needs to return false in order to indicate signal completion.

src/hip_stream.cpp
SarbojitAMD: Do we really need to initialize?

tests/src/runtimeApi/stream/StreamAddCallback.cpp
SarbojitAMD: Agree

src/hip_stream.cpp
```diff
@@ -254,14 +254,131 @@ hipError_t hipStreamGetPriority(hipStream_t stream, int* priority) {
...
+    depBarrier->header = header;
+    barrier->header = header;
+
+    // 6. Trigger the doorbell
+    nextIndex = nextIndex + 1;
+    hsa_queue_store_write_index_screlease(lockedQ, nextIndex);
+    hsa_signal_store_relaxed(lockedQ->doorbell_signal, index+1);
+
+    // 7. Release queue
+    stream->criticalData()._av.release_locked_hsa_queue();
```
asalmanp: Based on my previous request to lock the stream first in the above code then then the stream needs to be unlocked here as well.

SarbojitAMD: Thanks for pointing out added the change

src/hip_stream.cpp
```diff
@@ -254,14 +254,137 @@ hipError_t hipStreamGetPriority(hipStream_t stream, int* priority) {
...
 
-    // Create a thread in detached mode to handle callback
+    if(stream == hipStreamNull)
+    {
+        ihipCtx_t* device = ihipGetTlsDefaultCtx();
+        stream = device->_defaultStream;
+    }
+
+    // Lock the stream
+    LockedAccessor_StreamCrit_t crit(stream->criticalData(), true /*unlock at destruction*/);
```
asalmanp: the default value of autoUnlock argument of LockedAccessor constructor is true, so if your intention is to use the default value then it's not necessary to pass true here when you lock the stream.

Sarbojit2019: I have purposefully kept 'true' here as passing it will not have any extra cost but add better readability.

src/hip_stream.cpp
```diff
@@ -254,14 +254,137 @@ hipError_t hipStreamGetPriority(hipStream_t stream, int* priority) {
 
 
 //---
+hipError_t setCallbackPacket(hsa_queue_t* queue,
+                             uint64_t& index, uint64_t& nextIndex,
+                             hsa_barrier_and_packet_t** barrier1,
+                             hsa_barrier_and_packet_t** barrier2){
+
+    if(queue == nullptr ||
```
mangupta: None of these checks are needed since setCallbackPacket() is an internal function and only called from within hipStreamAddCallback() where all the necessary checks have been performed.

src/hip_stream.cpp
```diff
@@ -254,14 +254,137 @@ hipError_t hipStreamGetPriority(hipStream_t stream, int* priority) {
 
 
 //---
+hipError_t setCallbackPacket(hsa_queue_t* queue,
```
mangupta: If you look at the usage of this internal function it would never return anything other than hipSuccess. Moreover we don't check the return value of this function anywhere. So it should be fine to change return type to void.

tests/src/runtimeApi/stream/StreamAddCallback.cpp
```diff
@@ -0,0 +1,145 @@
...
+    return result;
+}
+
+int main()
+{
+    // Test involves multithreading hence running multiple times
+    // to make sure consitency in the behavior
+    bool status = true;
+
+    for(int i=0; i < 10; i++){
```
mangupta: In my opinion it is better to run different stream callback scenarios rather than running the same test 10 times. Instead we could add a subtest that enqueues 2 callbacks to the same stream. We could also try a multithread scenario  where one thread issues a stream callback while another thread is trying to enqueue work to the same queue.

SarbojitAMD: Agree, will create a new task for adding more scenarios.

PR #1429, opened 2019-09-18
Separated C macros from CPP header file

SarbojitAMD:
test_common.h is CPP compliant hence can't be included for gcc enabled tests. Separated C compliant Macros in common.h file so that it can be used in C/CPP tests.


Sarbojit2019:
Ideally I wanted to rename test_common.h as test_common.hpp but I was afraid of the impact it will have hence choose the safest path :). 
I can do as you suggested but only problem I see we need to have clear section called out in the file for C++ & C code else in future it will become like hip_runtime.h where we have C/C++ code scared all over.

mangupta:
> 
> 
> Ideally I wanted to rename test_common.h as test_common.hpp but I was afraid of the impact it will have hence choose the safest path :).
> I can do as you suggested but only problem I see we need to have clear section called out in the file for C++ & C code else in future it will become like hip_runtime.h where we have C/C++ code scared all over.

This should be fine since we don't modify test_common.h often and it is not a large file.

Issue #1433, opened 2019-09-18
tccDriver is missing in hipDeviceProp_t


Kaveh01:
tccDriver is missing in hipDeviceProp_t. It is obviously important only on Nvidia machines and I can use cudaDeviceProp instead with conditional compilation but it would be better if we had that in hipDeviceProp_t.

```
#include <hip/hip_runtime.h>
#include<stdio.h>
int main()
{
	hipDeviceProp_t deviceProp;
	hipGetDeviceProperties(&deviceProp, 0);
	int a = deviceProp.tccDriver;
	printf ("%d\n",a);
return 0;
}
```
The error: 
```
$ hipcc test.c
test.c(7): error: class "hipDeviceProp_t" has no member "tccDriver"

1 error detected in the compilation of "/tmp/tmpxft_00007587_00000000-8_test.cpp1.ii".
```

SarbojitAMD:
Raised PR https://github.com/ROCm-Developer-Tools/HIP/pull/1434/files

PR #1434, opened 2019-09-18
[HIP] Add tccDriver info in hipDeviceProp [github#1433]

SarbojitAMD:
It is to address ask on #1433

### DIFF
--- a/include/hip/hip_runtime_api.h
+++ b/include/hip/hip_runtime_api.h
@@ -126,6 +126,10 @@ typedef struct hipDeviceProp_t {
     size_t textureAlignment;         ///<Alignment requirement for textures
     int kernelExecTimeoutEnabled;    ///<Run time limit for kernels executed on the device
     int ECCEnabled;                  ///<Device has ECC support enabled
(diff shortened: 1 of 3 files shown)


include/hip/hip_runtime_api.h
```diff
@@ -126,6 +126,10 @@ typedef struct hipDeviceProp_t {
     size_t textureAlignment;         ///<Alignment requirement for textures
     int kernelExecTimeoutEnabled;    ///<Run time limit for kernels executed on the device
     int ECCEnabled;                  ///<Device has ECC support enabled
+#ifdef __HIP_PLATFORM_NVCC__
+    int tccDriver;                   ///< If device is Tesla device using TCC driver
+#endif
+
```
emankov: We may leave this field in HIP stack as well for compatibility but report 0.

mangupta: This should not be guarded. Otherwise code will not compile on both platforms. We should set this to field to 0 on AMD path.

PR #1215, opened 2019-07-10
Added gfx906 kernel binary to the HIPFB file for hipModule test.

kpyzhov:

PR #1444, opened 2019-09-19
[hip] Added gfx803,gfx1010,gfx1012 kernel binary hipModule test.

AlexXAmd:
This is to address hipModule test seg fault in GFX1010.


gargrahul:
@AlexXAmd Can't this be auto-generated now as discussed before in #1215 ?

AlexXAmd:
Yes. This can be auto generated. But such mechanism does not exist yet.

mangupta:
> 
> Yes. This can be auto generated. This default file can still cause problem...

@AlexXAmd So can we switch the build scripts to auto generate this file and modify the PR to delete this file instead?

AlexXAmd:
Yes. I was thinking about deleting the file. I am busy with some other stuff now. So I cannot do it.

PR #1446, opened 2019-09-20
Memset refactor with HSA api

cjatin:


gargrahul:
> 
> 
> HSA provides a memory fill API which will be faster than any of the custom kernels here - should we use that instead?

Hi Ben,

Yes we are trying to make use of HSA API here.  Please check https://github.com/ROCm-Developer-Tools/HIP/pull/1446/files#diff-97332d585992cdcb801525a79b121837R1638.

Regards,
Rahul

mangupta:
Closing this in favor of work being done in #1495. Refactoring changes (if any) will be taken up  after the other PR is merged.

PR #1447, opened 2019-09-20
Reimplement hipMemGetInfo

satyanveshd:
Addresses SWDEV-136570. hipMemGetInfo changed to compute free memory based on information from kfd instead of relying on hc::am_tracker.

### DIFF
--- a/src/hip_hcc_internal.h
+++ b/src/hip_hcc_internal.h
@@ -804,7 +804,7 @@ class ihipDevice_t {
     int _isLargeBar;
    
     // Node id reported by kfd for this device
-    uint32_t internal_node_id;
+    uint32_t _driver_node_id;
 
     ihipCtx_t* _primaryCtx;
 
--- a/src/hip_hcc_internal.h
+++ b/src/hip_hcc_internal.h
@@ -802,6 +802,9 @@ class ihipDevice_t {
 
     // TODO - report this through device properties, base on HCC API call.
     int _isLargeBar;
+   
+    // Node id reported by kfd for this device
+    uint32_t internal_node_id;
 
     ihipCtx_t* _primaryCtx;
 
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -1976,17 +1976,24 @@ hipError_t hipMemGetInfo(size_t* free, size_t* total) {
             } else {
 		uint32_t internal_node_id;
 		hsa_status_t err = hsa_agent_get_info(*agent, (hsa_agent_info_t) HSA_AMD_AGENT_INFO_DRIVER_NODE_ID, &internal_node_id);
-		if(err != HSA_STATUS_SUCCESS) return hipErrorInvalidDevice;
+		if (err != HSA_STATUS_SUCCESS) return ihipLogStatus(hipErrorInvalidDevice);
 		
-		size_t deviceMemSize;
 		std::string fileName = std::string("/sys/class/kfd/kfd/topology/nodes/") + std::to_string(internal_node_id) + std::string("/mem_banks/0/used_memory");  
                 
 		std::ifstream file;
 		file.open(fileName);
-		if(!file) return ihipLogStatus(hipErrorFileNotFound);
-		
-		file >> deviceMemSize;    
-		*free = device->_props.totalGlobalMem - deviceMemSize;
+		if (!file) return ihipLogStatus(hipErrorFileNotFound);
+	
+                std::string deviceSize;	
+		size_t deviceMemSize;
+
+		file >> deviceSize;
+                if ((deviceMemSize=strtol(deviceSize.c_str(),NULL,10))){
+		    *free = device->_props.totalGlobalMem - deviceMemSize;
+		} else {
+                    file.close();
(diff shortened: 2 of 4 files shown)


src/hip_memory.cpp
```diff
@@ -1969,14 +1970,27 @@ hipError_t hipMemGetInfo(size_t* free, size_t* total) {
...
+		
+		size_t deviceMemSize;
+		std::string fileName = std::string("/sys/class/kfd/kfd/topology/nodes/") + std::to_string(internal_node_id) + std::string("/mem_banks/0/used_memory");  
+                
+		std::ifstream file;
+		file.open(fileName);
+		if(!file) return ihipLogStatus(hipErrorFileNotFound);
+		
+		file >> deviceMemSize;    
+		*free = device->_props.totalGlobalMem - deviceMemSize;
```
SarbojitAMD: I feel it is good to have a check on read data i.e. deviceMemSize. There is some existing function e.g. isdigit() can be used to make sure read value is indeed a number.

src/hip_memory.cpp
```diff
@@ -1969,14 +1970,27 @@ hipError_t hipMemGetInfo(size_t* free, size_t* total) {
...
         if (free) {
-            // TODO - replace with kernel-level for reporting free memory:
-            size_t deviceMemSize, hostMemSize, userMemSize;
-            hc::am_memtracker_sizeinfo(device->_acc, &deviceMemSize, &hostMemSize, &userMemSize);
-
-            *free = device->_props.totalGlobalMem - deviceMemSize;
-
-            // Deduct the amount of memory from the free memory reported from the system
-            if (HIP_HIDDEN_FREE_MEM) *free -= (size_t)HIP_HIDDEN_FREE_MEM * 1024 * 1024;
+            hsa_agent_t* agent = static_cast<hsa_agent_t*>(device->_acc.get_hsa_agent());
```
SarbojitAMD: Looks like implementation assumes that 'device' will be not null if 'context' exists, can you please confirm that?

src/hip_memory.cpp
satyanveshd: used strtol() as the usedmemory can be large,  to check the value read is a number. Updated the PR.

src/hip_memory.cpp
satyanveshd: Yes, as we cannot create a context without having an associated device.

src/hip_memory.cpp
```diff
@@ -1969,14 +1970,34 @@ hipError_t hipMemGetInfo(size_t* free, size_t* total) {
...
+            if (!agent){
+		return ihipLogStatus(hipErrorInvalidResourceHandle);
+            } else {
+		uint32_t internal_node_id;
+		hsa_status_t err = hsa_agent_get_info(*agent, (hsa_agent_info_t) HSA_AMD_AGENT_INFO_DRIVER_NODE_ID, &internal_node_id);
```
mangupta: We don't need to query for nodeID everytime. So a better idea would be to store this information in ihipDevice_t and initialize the value in https://github.com/ROCm-Developer-Tools/HIP/blob/master/src/hip_hcc.cpp#L504.

src/hip_memory.cpp
```diff
@@ -1969,14 +1970,34 @@ hipError_t hipMemGetInfo(size_t* free, size_t* total) {
...
+		std::string fileName = std::string("/sys/class/kfd/kfd/topology/nodes/") + std::to_string(internal_node_id) + std::string("/mem_banks/0/used_memory");  
+                
+		std::ifstream file;
+		file.open(fileName);
+		if (!file) return ihipLogStatus(hipErrorFileNotFound);
+	
+                std::string deviceSize;	
+		size_t deviceMemSize;
+
+		file >> deviceSize;
```
mangupta: close file immediately after this instead having to deal with it in both the if() as well as else() cases.

src/hip_memory.cpp
```diff
@@ -1969,14 +1970,34 @@ hipError_t hipMemGetInfo(size_t* free, size_t* total) {
...
+                if ((deviceMemSize=strtol(deviceSize.c_str(),NULL,10))){
+		    *free = device->_props.totalGlobalMem - deviceMemSize;
+		} else {
+                    file.close();
+ 		    return ihipLogStatus(hipErrorInvalidValue);
+		}
+		file.close();                 
+            	// Deduct the amount of memory from the free memory reported from the system
+		if (HIP_HIDDEN_FREE_MEM) *free -= (size_t)HIP_HIDDEN_FREE_MEM * 1024 * 1024;
```
mangupta: This can be moved into the if() case above.

src/hip_hcc_internal.h
```diff
@@ -802,6 +802,9 @@ class ihipDevice_t {
 
     // TODO - report this through device properties, base on HCC API call.
     int _isLargeBar;
+   
+    // Node id reported by kfd for this device
+    uint32_t internal_node_id;
```
mangupta: Rename to _node_id or _driver_node_id

satyanveshd: renamed to _driver_node_id. updated the PR.

PR #1449, opened 2019-09-21
[HIP] Added new hipFreeAll API [Feature]

SarbojitAMD:
hipFreeAll is a new proposed hip runtime API that takes variable arguments to free all of them. 
e.g. Say at any point in time user wants to free memory allocated in 10 different type of pointers in his program then he/she needs to write something similar mentioned below 
hipFree(varPtr1);
hipFree(varPtr2);
hipFree(varPtr3);
........
hipFree(varPtr10);    
With hipFreeAll API it can be achieved as hipFreeAll(varPtr1,varPtr2,varPtr3........varPtr10);

### DIFF
--- a/include/hip/hcc_detail/functional_grid_launch.hpp
+++ b/include/hip/hcc_detail/functional_grid_launch.hpp
@@ -177,4 +177,19 @@ void hipLaunchKernelGGL(F kernel, const dim3& numBlocks, const dim3& dimBlocks,
                                      stream, &config[0]);
 }
 
+template<class T>
+hipError_t hipFreeAll(T ptr)
+{
+    return hipFree((void *) ptr);            
+}
+
+template<class T, class... Args>
+hipError_t hipFreeAll(T first, Args... arg)
+{
+    hipError_t status = hipFree(first);
+    if(status == hipSuccess)
+       status = hipFreeAll(arg...);
+    return status;                            
+}
+
 #pragma GCC visibility pop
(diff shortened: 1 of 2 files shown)


gargrahul:
Why do we need this ? I am not sure I understand the value of this new API.

SarbojitAMD:
@gargrahul, Below are the list of use cases I can see. You can take a look at test that I have modified to show the use cases.

1. **Less code**: Today, HIP programmer has to add 10 hipFree calls in case of 10 hipMalloc, proposed API can take all pointers in one go hence user has to type less code.
2. **Generic interface**:  I want to have a generic API so that it can free any type of allocation done via hip runtime e.g. today user has to call hipFree / hipHostFree/..... 
Note:  In the current proposal I have made the change only for hipFree but in future, it should be able to handle all types of allocations.

mangupta:
I think free'ing multiple pointers together in the same API is a bad programming practice. What if there is memory corruption as a result the API fails. How is the user to figure out which buffer was the culprit?
Also if the user requires a quick way, they can use hipDeviceReset over each device to automatically free all memory allocated on the device.
So closing this PR for now.

Sarbojit2019:
Anyway, I don't think I understood the reasons very well.
1. First of all proposal was not to replace existing Free APIs so users will always have a choice to choose their convenience.
2. Even today crashing one of the free call will not tell user anything unless one goes and debug it. If you are referring about the traces that anyway will be there in current proposal as underlying interfaces remains are same.
3. hipDeviceReset and hipFreeAll has completely different use cases hence calling hipDeviceReset can be replacement of hipFreeAll is not correct. Say, I have 100 allocations and I want to free only 20 at any point in time which can't be done using hipDeviceReset.

PR #1456, opened 2019-09-23
[hip] add support for implicit kernel argument for multi-grid sync

asalmanp:
background: for the cooperative groups feature, runtime needs to setup a new implicit argument for multi-grid synchronisation as follows;

- set to 1 if the kernel is launched with hipLaunchCooperativeKernel
- set to a pointer to a properly initialized struct mg_info if the kernel is launched with hipLaunchCooperativeKernelMultiDevice

### DIFF
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -471,7 +471,7 @@ hipError_t hipLaunchCooperativeKernel(const void* f, dim3 gridDim,
         return ihipLogStatus(hipErrorLaunchFailure);
     }
 
-    uint impCoopArg = 1;
+    size_t impCoopArg = 1;
     void* impCoopParams[1];
     impCoopParams[0] = &impCoopArg;
 
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -583,9 +583,7 @@ hipError_t hipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsLi
 #endif
     }
 
-    void* impCoopParams[1];
-    ulong prev_sum = 0;
-    // launch the init_gws kernel to initialize the GWS followed by launching the main kernels for each device
+    // launch the init_gws kernel to initialize the GWS for each device
     for (int i = 0; i < numDevices; ++i) {
         const hipLaunchParams& lp = launchParamsList[i];
 
@@ -610,6 +608,14 @@ hipError_t hipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsLi
 
             return ihipLogStatus(hipErrorLaunchFailure);
         }
+    }
+
+    void* impCoopParams[1];
+    ulong prev_sum = 0;
+    // launch the main kernels for each device
+    for (int i = 0; i < numDevices; ++i) {
+        const hipLaunchParams& lp = launchParamsList[i];
+
         //initialize and setup the implicit kernel argument for multi-grid sync
         mg_info_ptr[i]->mgs       = mg_sync_ptr;
         mg_info_ptr[i]->grid_id   = i;
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -580,6 +580,7 @@ hipError_t hipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsLi
     }
 
     void* impCoopParams[1];
+    ulong prev_sum = 0;
     // launch the init_gws kernel to initialize the GWS followed by launching the main kernels for each device
     for (int i = 0; i < numDevices; ++i) {
         const hipLaunchParams& lp = launchParamsList[i];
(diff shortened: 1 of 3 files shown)


src/hip_module.cpp
```diff
@@ -565,6 +600,19 @@ hipError_t hipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsLi
             }
             return ihipLogStatus(hipErrorLaunchFailure);
         }
+        //initialize and setup the implicit kernel argument for multi-grid sync
+        mg_info_ptr[i]->mgs       = mg_sync_ptr;
+        mg_info_ptr[i]->grid_id   = i;
+        mg_info_ptr[i]->num_grids = numDevices;
+        mg_info_ptr[i]->all_sum   = all_sum;
+        mg_info_ptr[i]->prev_sum  = 0;
+        for (int j = 0; j < i; ++j) {
```
b-sumner: Wouldn't it be easier and faster to keep a running sum in the i loop instead of recalculating prev_sum on each iteration?

src/hip_module.cpp
```diff
@@ -463,14 +471,18 @@ hipError_t hipLaunchCooperativeKernel(const void* f, dim3 gridDim,
         return ihipLogStatus(hipErrorLaunchFailure);
     }
 
+    uint impCoopArg = 1;
```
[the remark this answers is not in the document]
asalmanp: The content of where impCoopParams[0] is pointing is actually inserted in the sixth position for the implicit kernel argument of the multi-grid sync. In the single grid, 1 is actually inserted there. I've tested this code by calling the builtin function (size_t *)__builtin_amdgcn_implicitarg_ptr())[6] and confirmed that this function returns 1 for the single grid case.

src/hip_module.cpp
asalmanp: Agree. Modified code as requested.

src/hip_module.cpp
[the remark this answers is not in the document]
b-sumner: But the type of impCoopArg is uint which is smaller than a pointer.  Copying could access out of bounds.  The type should be size_t or at least the same size as a pointer.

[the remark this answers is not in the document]
asalmanp: changed the type of the impCoopArg to size_t.

src/hip_module.cpp
```diff
@@ -537,6 +549,28 @@ hipError_t hipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsLi
...
+        return ihipLogStatus(hipErrorInvalidValue);
+    }
+    mg_sync_ptr->w0 = 0;
+    mg_sync_ptr->w1 = 0;
+
+    uint all_sum = 0;
+    for (int i = 0; i < numDevices; ++i) {
+        result = hip_internal::ihipHostMalloc(tls, (void **)&mg_info_ptr[i], sizeof(mg_info), hipHostMallocDefault);
+        if (result != hipSuccess) {
+            return ihipLogStatus(hipErrorInvalidValue);
```
b-sumner: Is there a concern about this leaking memory when there is a failure?

src/hip_module.cpp
```diff
@@ -565,6 +601,17 @@ hipError_t hipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsLi
             }
             return ihipLogStatus(hipErrorLaunchFailure);
         }
+        //initialize and setup the implicit kernel argument for multi-grid sync
```
b-sumner: If the init launch above fails, should the mg_infos be freed too?

src/hip_module.cpp
```diff
@@ -573,7 +620,7 @@ hipError_t hipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsLi
                 lp.blockDim.x, lp.blockDim.y,
                 lp.blockDim.z, lp.sharedMem,
                 lp.stream, lp.args, nullptr, nullptr, nullptr, 0,
-                true);
+                true, impCoopParams);
```
b-sumner: What if this fails?  Should there be some cleanup?

asalmanp: added the memory clean up

src/hip_module.cpp
```diff
@@ -587,6 +659,11 @@ hipError_t hipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsLi
     free(gwsKds);
     free(kds);
 
+    hip_internal::ihipHostFree(tls, mg_sync_ptr);
+    for (int j = 0; j < numDevices; ++j) {
+        hip_internal::ihipHostFree(tls, mg_info_ptr[j]);
+    }
```
b-sumner: Can these be freed here?  Don't we have to wait until all the kernels are finished before we can free these?

asalmanp: there is a wait inside ihipHostFree to make sure that the kernels are done before freeing memories.

src/hip_module.cpp
```diff
@@ -563,8 +603,24 @@ hipError_t hipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsLi
                 launchParamsList[j].stream->criticalData()._av.release_locked_hsa_queue();
 #endif
             }
+            hip_internal::ihipHostFree(tls, mg_sync_ptr);
+            for (int j = 0; j < numDevices; ++j) {
+                hip_internal::ihipHostFree(tls, mg_info_ptr[j]);
```
b-sumner: Suppose a few launch successfully, start running, and hit a multigrid barrier.  Then there is a failed launch.  Won't the internal wait in the free wait "forever"?   Should the successful launches be "killed", e.g. by resetting their queues?

asalmanp: There is no easy way to reset a queue on the fly without causing further problems. So, a deadlock may occur in your described scenario above. 

To reduce the chance of a deadlock for this particular memory cleanup, I have separated the launch of init_ gws and main kernels into two loops. This way, the GWS gets initialized on all devices before trying to launch the main kernels and that would eliminate the chance of the deadlock here if one of the init_gws launches fails.

The hazard would still remain for the memory clean up later in the code if one of the main kernels fail, but I would say the chance of this failure is low because when we reach to this point to launch the main kernels, all the sanity checks have been done (e.g., valid stream, valid kernel descriptors, successful launch of gws_kernel etc) so the launch hopefully should go in successfully. 

when we start testing this feature, we can design a negative test to particularly test such a scenario.

b-sumner:
Umm...actually I'm not sure I have the right to approve, but I endorse this change.

gargrahul:
> 
> 
> Umm...actually I'm not sure I have the right to approve, but I endorse this change.

Thanks Brian, you have full right to approve and we needed your approval here  👍

PR #1463, opened 2019-09-25
Add Texture reference APIs headers

kjayapra-amd:
Adding Texture API declarations to hip_runtime_api.h 
Added APIs-
- hipTexRefGetArray
- hipTexRefGetAddressMode
- hipTexRefGetAddress


emankov:
What's the point in adding declarations without definitions?

kjayapra-amd:
Rahul is working on definitions for HCC, VDI already has the definition implemented.

emankov:
This repo doesn't contain definitions. They should go together in a single commit, or this commit should go last.

gargrahul:
@emankov It appears that these functions are already implemented in HIP/VDI and they are waiting for header changes.
Personally I am NOT in favor of header ONLY changes.

gargrahul:
Closing it in favor of #1471

PR #1464, opened 2019-09-25
hipModuleUnload should remove global variables from memtracker

jeffdaily:


src/hip_module.cpp
```diff
@@ -821,6 +807,18 @@ inline hsa_status_t copy_agent_global_variables(hsa_executable_t, hsa_agent_t ag
     return HSA_STATUS_SUCCESS;
 }
 
+inline hsa_status_t remove_agent_global_variables(hsa_executable_t, hsa_agent_t agent,
+                                                  hsa_executable_symbol_t x, void* unused) {
+    hsa_symbol_kind_t t = {};
+    hsa_executable_symbol_get_info(x, HSA_EXECUTABLE_SYMBOL_INFO_TYPE, &t);
```
jerryyin: I don't know enough about the difference between this and `HSA_EXECUTABLE_SYMBOL_INFO_VARIABLE_ALLOCATION`, just FYI

jeffdaily: If you look at `copy_agent_global_variables`, just above this new function, I am performing the same loops as `copy_agent_global_variables`, but instead of tracking the memory, I un-track it.

jerryyin: That's right! I think we need a HIP reviewer to tell us:

1. Why are we seeing the issue only in global symbols
2. Is there any other cases that can cause a `g_amPointerTracker` collision, and how do they prevent that.

PR #1457, opened 2019-09-23
Add new kernel launching API for hip-clang

yxsamliu:

PR #1476, opened 2019-09-27
Add new kernel launching API for hip-clang

yxsamliu:
this is the same as https://github.com/ROCm-Developer-Tools/HIP/pull/1457
I created this one because CI seems to have issue with that one


yxsamliu:
the old one has passed. so close this one

PR #1381, opened 2019-08-31
Revert "Using HSA API for hipMemsetAsync"

gargrahul:
Reverts ROCm-Developer-Tools/HIP#1346

This implementation requires review and possible rework.

PR #1495, opened 2019-10-01
General hipMemset improvements

jeffdaily:
This is the next attempt to fix #1346 and #1381.


jeffdaily:
@emankov @mangupta please review.

src/hip_memory.cpp
```diff
@@ -1577,7 +1577,21 @@ hipError_t ihipMemPtrGetInfo(void* ptr, size_t* size) {
 }
 
 template <typename T>
-void ihipMemsetKernel(hipStream_t stream, T* ptr, T val, size_t count) {
+void ihipMemsetKernel(hipStream_t stream, T* ptr, T val, size_t count, bool isAsync) {
+    // Just Use count, instead of dividing by 4, the calling API already does it
+    if (sizeof(T) == sizeof(uint32_t) && (count % sizeof(uint32_t) == 0) && !isAsync) {
```
emankov: `!isAsync` check at the beginning looks more preferable, otherwise LGTM.

jeffdaily: Fixed by https://github.com/ROCm-Developer-Tools/HIP/pull/1495/commits/26d166127649674295668c3fe9c9b455ece6f1ee .

jeffdaily:
Had to rebase PR due to conflicts caused by 0b52c1d9d839864a36ee019f7ce3345c405e1fe6.

jeffdaily:
@emankov requesting a new review.  After some discussion with @AlexVlx , the PR was refined to cover additional synchronous cases.

Also, I split the ihipMemset into two distinct functions, ihipMemsetSync/Async.  The two code paths are distinct and clear.  All call sites for ihipMemsetSync/Async have been cleaned up

- removed redundant stream syncs after ihipMemsetSync is called
- removed redundant stream == null checks

src/hip_memory.cpp
```diff
@@ -1861,182 +1932,78 @@ hipError_t hipMemcpyParam2DAsync(const hip_Memcpy2D* pCopy, hipStream_t stream)
...
-    hipStream_t stream = hipStreamNull;
-    stream = ihipSyncAndResolveStream(stream);
-    if (stream) {
-        e = ihipMemset(dst, value, count, stream, ihipMemsetDataTypeChar);
-        stream->locked_wait();
-    } else {
-        e = hipErrorInvalidValue;
-    }
-    return ihipLogStatus(e);
+    hipStream_t stream = ihipSyncAndResolveStream(hipStreamNull);
```
gargrahul: Move stream resolution inside ihipMemset, there is no point in syncing if ihipMemset returns early due to count =0 or !dst.

src/hip_memory.cpp
```diff
@@ -1861,182 +1932,78 @@ hipError_t hipMemcpyParam2DAsync(const hip_Memcpy2D* pCopy, hipStream_t stream)
...
-    hipStream_t stream = hipStreamNull;
-    stream =  ihipSyncAndResolveStream(stream);
-    if (stream) {
-        e = ihipMemset(dst, value, sizeBytes, stream, ihipMemsetDataTypeChar);
-        stream->locked_wait();
-    } else {
-        e = hipErrorInvalidValue;
-    }
-    return ihipLogStatus(e);
+    hipStream_t stream = ihipSyncAndResolveStream(hipStreamNull);
```
gargrahul: Move stream resolution inside ihipMemset, there is no point in syncing if ihipMemset returns early due to count =0 or !dst.

AlexVlx:
@gargrahul @mangupta you gentlemen reckon we could try to get this in so that it makes it into the next release?

iotamudelta:
To add color to @AlexVlx 's comment above - this is an enhancement for some important PyTorch workloads and this PR must make it in for 3.0.

gargrahul:
> 
> 
> To add color to @AlexVlx 's comment above - this is an enhancement for some important PyTorch workloads and this PR must make it in for 3.0.

@iotamudelta I am not sure how does this help PyTorch - hsa API is being used in memsetSync only. Though this should get merged after suggested minor changes and be part of 3.0.

AlexVlx:
This is false. If you look carefully you will observe that the kernel that is invoked for the async case has also been rewritten. As surprising as it may appear, this makes it faster. As I mentioned already, I think that the extra stream resolution is spurious, since it precedes a kernel call that would be ordered by its very nature. We need stream resolution when making an out-of-band call to the HSA RT, in the synchronous case, but that's a different kettle of fish.

iotamudelta:
@gargrahul you are commenting on changes in the async part, yes? Also please see commit history.

gargrahul:
> 
> 
> This is false. If you look carefully you will observe that the kernel that is invoked for the async case has also been rewritten. As surprising as it may appear, this makes it faster. As I mentioned already, I think that the extra stream resolution is spurious, since it precedes a kernel call that would be ordered by its very nature. We need stream resolution when making an out-of-band call to the HSA RT, in the synchronous case, but that's a different kettle of fish.

Agree with you about stream resolution part being taken care by ihipModuleLaunchKernel. We should still be consistent though. In few memsetAsync APIs we are resolving stream whereas in few we are skipping.

AlexVlx:
@gargrahul that is an excellent idea. We should be consistent in not doing something spurious, so the one implementation that was still syncing and resolving the stream has been corrected, thank you ever so much for identifying it. You say "few" - did I miss any spots?

AlexVlx:
@gargrahul I appreciate your willingness to do a thing, but perhaps in the future we can discuss your intention, and figure out if it's a good idea or not. Drive-by commits are problematic without discussing things with the authors of a particular piece of code. To wit, as I've outlined in the comments, the most recent addition isn't ideal.

gargrahul:
@AlexVlx 

> 
> 
> @gargrahul I appreciate your willingness to do a thing, but perhaps in the future we can discuss your intention, and figure out if it's a good idea or not. Drive-by commits are problematic without discussing things with the authors of a particular piece of code. To wit, as I've outlined in the comments, the most recent addition isn't ideal.

@AlexVlx I am in agreement with you on discussing before pushing commits this way. My sole intention was to avoid another iteration, we are getting multiple reminders to get this PR merged.

Am I missing something? All sync APIs were passing nullptr as stream and was this still needed https://github.com/ROCm-Developer-Tools/HIP/pull/1495/commits/e0fed1bedc4af07900026d0cdded065bd3f97787#diff-97332d585992cdcb801525a79b121837R1753

AlexVlx:
@gargrahul the intention is noble, but there are a couple of issues with how it was being expressed:
- reformulating it as such means that if we ever want to do a synchronous memset per stream, we'd have to either add back the old interface, or add yet another function - a synchronous per-stream memset looks potentially helpful to me, and we may want to explore it a bit, even though if it doesn't exist in CUDA per se;
- I've added two comments which explained why I believed that the way in which the implementation morphed was not ideal - whilst the fact that we'd be doing stream resolution outside of the try block is benign, if quaint, the extra wait was all pain, no gain.
TL;DR, unless we have technical concerns (which would be super useful to hear, since important client apps have already successfully crunched through this, but that's not the the fount of all knowledge), IMHO this is perfectly adequate for merging as is, and I'd go so far as to say that it meets the high standards of the existing HIP codebase.

PR #1509, opened 2019-10-07
[HIP] Introduce library_types.h as a common header for libs

emankov:
[Reason]
Currently, hipFFT, hipBLAS and other HIP libs use their own data types, prefixed with HIPFFT or HIPBLAS, whereas in CUDA those types are common and declared in library_types.h

[TODO]
Switch hipFFT, hipBLAS and other HIP libs to use common library_types.h.

### DIFF
--- a/include/hip/hcc_detail/hip_runtime_api.h
+++ b/include/hip/hcc_detail/hip_runtime_api.h
@@ -41,7 +41,6 @@ THE SOFTWARE.
 
 #include <hip/hcc_detail/host_defines.h>
 #include <hip/hcc_detail/driver_types.h>
-#include <hip/hcc_detail/library_types.h>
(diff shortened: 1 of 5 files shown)


emankov:
ping

emankov:
ping

include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -41,6 +41,7 @@ THE SOFTWARE.
 
 #include <hip/hcc_detail/host_defines.h>
 #include <hip/hcc_detail/driver_types.h>
+#include <hip/hcc_detail/library_types.h>
```
mangupta: I don't see why it should be included here. The libraries that require it should include library_types.h similar to how it is in CUDA.

PR #1511, opened 2019-10-08
Fix for directed tests failure

ansurya:
SWDEV-203971:
directed_tests/runtimeApi/module/hipLaunchCooperativeKernel.tst - Disabling test temporarily until driver support is available.
directed_tests/runtimeApi/memory/hipArray.tst


ansurya:
Disabled hipArray.tst as behavior is random on different GPU's. It needs clean up as it uses driver and runtime API's. Disabling test temporarily.

PR #1522, opened 2019-10-10
[hip] Enclose library_types with namespace

mhbliao:
- to avoid conflicting as these are quite common identifiers.


mangupta:
Closing this since #1527 has been merged which is a better fix.

PR #1528, opened 2019-10-11
Guard against division by zero for no VGPR usage (e.g., in an empty k…

arghdos:
…ernel)

### DIFF
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -1326,15 +1326,15 @@ hipError_t ihipOccupancyMaxActiveBlocksPerMultiprocessor(
     size_t numWavefronts = (blockSize + wavefrontSize - 1) / wavefrontSize;
 
     size_t availableVGPRs = (prop.regsPerBlock / wavefrontSize / simdPerCU);
-    size_t vgprs_alu_occupancy = simdPerCU * (usedVGPRS == 0) ? maxWavesPerSimd
-        : std::min(maxWavesPerSimd, availableVGPRs / usedVGPRS);
+    size_t vgprs_alu_occupancy = simdPerCU * (usedVGPRS == 0 ? maxWavesPerSimd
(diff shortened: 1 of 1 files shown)


src/hip_module.cpp
```diff
@@ -1326,7 +1326,8 @@ hipError_t ihipOccupancyMaxActiveBlocksPerMultiprocessor(
     size_t numWavefronts = (blockSize + wavefrontSize - 1) / wavefrontSize;
 
     size_t availableVGPRs = (prop.regsPerBlock / wavefrontSize / simdPerCU);
-    size_t vgprs_alu_occupancy = simdPerCU * std::min(maxWavesPerSimd, availableVGPRs / usedVGPRS);
+    size_t vgprs_alu_occupancy = simdPerCU * ((usedVGPRS == 0) ? maxWavesPerSimd
```
emankov: LGTM, otherwise excessive brackets.

arghdos: I actually used the bracket format from the SGPR check a few lines down :p
Should I change both?

arghdos: Actually, after removing I get gcc errors:

>/home/research/code/HIP/src/hip_module.cpp:1329:63: warning: operator '?:' has lower precedence than '*'; '*' will be evaluated first [-Wparentheses]
>    size_t vgprs_alu_occupancy = simdPerCU * (usedVGPRS == 0) ? maxWavesPerSimd

looks like we need to keep them

emankov: There were other (inner) brackets in mind, actually, smth. like:
```c++
size_t vgprs_alu_occupancy = simdPerCU * (usedVGPRS ? std::min(maxWavesPerSimd, availableVGPRs / usedVGPRS) : maxWavesPerSimd);
```

arghdos: gotcha

PR #1523, opened 2019-10-10
Add hipExtLaunchKernelGGL

gargrahul:
Provides goodness of hipExtModuleLaunchKernel as a C++ API.

PR #1530, opened 2019-10-12
Add support for extended launch syntax.

AlexVlx:
This depends on https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/183. It supersedes https://github.com/ROCm-Developer-Tools/HIP/pull/1523.

### DIFF
--- a/tests/src/kernel/hipExtLaunchKernelGGL.cpp
+++ b/tests/src/kernel/hipExtLaunchKernelGGL.cpp
@@ -25,6 +25,7 @@ THE SOFTWARE.
  */
 
 #include "hip/hip_runtime.h"
+#include "hip/hip_ext.h"
 #include "test_common.h"
 
 void test(size_t N) {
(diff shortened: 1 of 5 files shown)


tests/src/kernel/hipExtLaunchKernelGGL.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+    HIPCHECK(hipDeviceSynchronize());
+
+    HipTest::checkVectorADD(A_h, B_h, C_h, N);
+}
+
+int main(int argc, char* argv[]) {
+    HipTest::parseStandardArguments(argc, argv, true);
+
+    test_gl2(N);
```
mangupta: Typo: should be ```test```.

AlexVlx:
@mangupta this won't work until the Clang support is in place (that is the reason for the build failure). Thanks for spotting the typo.

PR #1531, opened 2019-10-12
Improve scalar access into vector types.

AlexVlx:
The improvement is based on the ideas here: https://t0rakka.silvrback.com/simd-scalar-accessor. It yields significantly better ISA when the base's .xyzw members are used.


AlexVlx:
> Can't this scalar accessor improvement be applied to vec3 scenario at lines 310-317? LGTM otherwise.

Not quite as those are already "OK", they don't alias with a native vector type, so there's no overhead associated with in. If the 3-wide vectors ever revert to using native vectors it would be necessary to do this there as well, but a choice was made to optimise those for storage size, and not for performance.

PR #1539, opened 2019-10-14
Improve performance of v2 arg handling

searlmc1:


AlexVlx:
@searlmc1 this is missing a change: on https://github.com/ROCm-Developer-Tools/HIP/blob/41b29c699fa65a53abba01d0cd151b6b0cfa46b0/src/program_state.inl#L152 the vector holds `std::string`, not `std::vector<char>`. This is why the CI fails. Otherwise LGTM.

PR #1540, opened 2019-10-14
Temporary add -D_OPENMP to clang options in hipcc to allow using CMak…

kpyzhov:
…e OpenMP detection with hip-clang (until updated CMake version is available).

### DIFF
--- a/bin/hipcc
+++ b/bin/hipcc
@@ -526,6 +526,13 @@ foreach $arg (@ARGV)
         $optArg = $arg;
     }
 
+    ## This is a temporary workaround for CMake detection of OpenMP support.
+    ## It should be removed when the OpenMP detection c++ test in CMake is updated
+    ## and corrected CMake version is available.
+    if($arg eq '-fopenmp') {
+        $HIPCXXFLAGS .= " -D_OPENMP "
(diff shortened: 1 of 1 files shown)


kpyzhov:
The CMake OpenMP detection mechanism "find_package(OpenMP)" does not work with hipcc and hip-clang. The small test that CMake uses to detect OpenMP support in the compiler fails to compile by hipcc because _OPENMP macro is not defined for the DEVICE compilation pass.

I submitted a corrected CMake test to the CMake Gitlab repo, and the change has been approved. However, it will take considerable time to be able to use updated CMake version. So, let's make this temporary change to the hipcc script to add "-D_OPENMP" to the clang options and keep this change until fixed CMake version is available.

yxsamliu:
sorry I thought it was a rocblas change.

I do have some concerns about this: -D_OPENMP indicates -fopenmp is passed to the compiler.

we should limit this to hip-clang only (HIP_COMPILER == "clang") and only when -fopenmp is passed to the compiler.

kpyzhov:
The check for "-fopenmp" is already there, I will add check for clang compiler Thanks!

PR #1543, opened 2019-10-15
[HIPRTC] Fix hard-coded path to hipcc

shoshijak:
If the environment variable `HIP_PATH` does not exist, HIPRTC searches
for the hipcc executable at `/opt/rocm/bin`, when in fact, hipcc is
located at `/opt/rocm/hip/bin`.
This commit corrects the hard-coded path to hipcc.

---

**Note**
a workaround for this issue is to set `HIP_PATH` in the environment: `export HIP_PATH=/opt/rocm/hip`

### DIFF
--- a/src/hiprtc.cpp
+++ b/src/hiprtc.cpp
@@ -481,7 +481,7 @@ hiprtcResult hiprtcCompileProgram(hiprtcProgram p, int n, const char** o)
 
     static const string hipcc{
         getenv("HIP_PATH") ? (getenv("HIP_PATH") + string{"/bin/hipcc"})
-                           : "/opt/rocm/bin/hipcc"};
+                           : "/opt/rocm/hip/bin/hipcc"};
 
     if (!experimental::filesystem::exists(hipcc)) {
         return HIPRTC_ERROR_INTERNAL_ERROR;


gargrahul:
@shoshijak were you seeing any issue ? There should be /opt/rocm/bin/hipcc (symlinked to /opt/rocm/hip/bin/hipcc).

shoshijak:
@gargrahul Yes, I see an issue when I run ROCm's [jitted saxpy example](https://github.com/ROCm-Developer-Tools/HIP/tree/master/tests/src/hiprtc). Running this example on AMD fails with 

```
runtime fail: hiprtcCompileProgram fails with error HIPRTC_ERROR_INTERNAL_ERROR
```

mangupta:
> 
> @gargrahul Yes, I see an issue when I run ROCm's [jitted saxpy example](https://github.com/ROCm-Developer-Tools/HIP/tree/master/tests/src/hiprtc). Running this example on AMD fails with
> 
> ```
> runtime fail: hiprtcCompileProgram fails with error HIPRTC_ERROR_INTERNAL_ERROR
> ```

@shoshijak I am guessing that you are not using HIP deb/rpm packages and are instead building and installing HIP from source. The deb/rpm packages contain scripts that create the necessary symlinks.

mangupta:
With PR #1379 we are trying to move to a state where we can have multiple ROCm installations in the system. So we need to determine HIP_PATH in the relative path. Maybe something along the lines of https://github.com/ROCm-Developer-Tools/HIP/pull/1097/files/f0304546944f3ef496c2586222d4c8c87d00a29a#r284146990?

shoshijak:
> @shoshijak I am guessing that you are not using HIP deb/rpm packages and are instead building and installing HIP from source.

Correct.

> The deb/rpm packages contain scripts that create the necessary symlinks.

I see. If these symlinks are actually necessary (as they appear to be in this case, for just-in-time compilation), shouldn't the build procedure create these symlinks as well?

> Maybe something along the lines of https://github.com/ROCm-Developer-Tools/HIP/pull/1097/files/f0304546944f3ef496c2586222d4c8c87d00a29a#r284146990?

I just read the discussion you pointed to and found @AlexVlx 's argument convincing: hiprtc shouldn't search for the `hipcc` executable relatively to `libhiprtc.so`. 
But here's a different idea:

> With PR #1379 we are trying to move to a state where we can have multiple ROCm installations in the system. So we need to determine HIP_PATH in the relative path.

Right. In that case, since `HIP_PATH` is a variable set in cmake, why not have cmake pass it to `hiprtc.cpp` as a preprocessor definition? This would allow hiprtc to find `hipcc` 
a) even in the absence of a `HIP_PATH` environment variable
b) even with "multiple ROCm installations in the system"

If you think this solution makes sense, I can update this current PR accordingly.

mangupta:

PR #1399, opened 2019-09-09
Added new Memory API's

ansurya:
Added new memory API's hipMemAllocPitch, hipMemAllocHost, hipMemsetD16, hipMemsetD16Async, hipMemsetD8Async
Modified to support all scenarios hipMemcpyParam2DAsync, hipMemcpyParam2D.

PR #1544, opened 2019-10-15
QoL changes to the hipMemset family

vsytch:


tests/src/runtimeApi/memory/hipMemset2D.cpp
```diff
@@ -113,8 +114,8 @@ int main(int argc, char *argv[])
     hipCtxCreate(&context, 0, p_gpuDevice);
 
     bool testResult = false;
```
mangupta: testResult should be set to true by default for the '&' to work correctly in the following lines.

tests/src/runtimeApi/memory/hipMemset3D.cpp
```diff
@@ -82,9 +133,10 @@ bool testhipMemset3D(int memsetval,int p_gpuDevice)
 int main(int argc, char *argv[])
 {
     HipTest::parseStandardArguments(argc, argv, true);
-    bool testResult = false;
     HIPCHECK(hipSetDevice(p_gpuDevice));
-    testResult = testhipMemset3D(memsetval, p_gpuDevice);
+    bool testResult = false;
```
mangupta: testResult should be set to true by default for the '&' to work correctly in the following lines.

gargrahul:
Also current hipMemset2D and hipMemset3D tests don't need hipCtxCreate and destroy calls. This is not related to this PR changes though. These were added in #1399.

PR #1549, opened 2019-10-16
[HIPIFY][cmake] Add LLVM_DIR and default path to find LLVM

aaronenyeshi:
We need this change for the CI to look inside LLVM_DIR and also for regular usage (with llvm-amdgpu) to find at /opt/rocm/llvm.

### DIFF
--- a/hipify-clang/CMakeLists.txt
+++ b/hipify-clang/CMakeLists.txt
@@ -7,7 +7,7 @@ if (MSVC AND MSVC_VERSION VERSION_LESS "1900")
     return()
 endif()
 
-find_package(LLVM REQUIRED)
+find_package(LLVM REQUIRED PATHS ${LLVM_DIR} "/opt/rocm/llvm" NO_DEFAULT_PATH)
 message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}:")
 message(STATUS "   - CMake module path: ${LLVM_CMAKE_DIR}")
 message(STATUS "   - Include path     : ${LLVM_INCLUDE_DIRS}")


hipify-clang/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ if (MSVC AND MSVC_VERSION VERSION_LESS "1900")
     return()
 endif()
 
-find_package(LLVM REQUIRED)
+find_package(LLVM REQUIRED PATHS ${LLVM_DIR} "/opt/rocm/llvm" NO_DEFAULT_PATH)
```
emankov: We are not building hipify-clang against Rocm LLVM; only against official releases or trunk LLVM. 

https://github.com/ROCm-Developer-Tools/HIP/tree/master/hipify-clang#-dependencies:

In most cases, you can get a suitable version of LLVM+CLANG with your package manager.

Failing that or having multiple versions of LLVM, you can [download a release archive](http://releases.llvm.org/), build or install it, and set
[CMAKE_PREFIX_PATH](https://cmake.org/cmake/help/v3.5/variable/CMAKE_PREFIX_PATH.html) so `cmake` can find it; for instance: `-DCMAKE_PREFIX_PATH=f:\LLVM\9.0.0\dist`

aaronenyeshi: Okay perhaps I can use CMAKE_PREFIX_PATH instead of LLVM_DIR. But we need to discuss if we can support ROCm LLVM

emankov: We can try. Additionally, all the needed features and fixes should be checked: firstly, unit tests should pass (against CUDA 10.1).

aaronenyeshi: @aakanksha555 - have you ran hipify-clang with ROCm LLVM?

emankov: Never

aaronenyeshi: I've just tested it. hipify-clang successfully builds and installs with ROCm LLVM/Clang. I had to make a small change to turn the cmake version to 3.5.1.

aaronenyeshi:
Closing this ticket because using CMAKE_PREFIX_PATH to LLVM's dist works.

PR #1552, opened 2019-10-17
fix code object v3 switch for hcc

scchan:
- enable code object v3 if hcc version >= 2.10
- pass -mcode-object-v3 to hcc when compiling
- -Wl,-hcc-cov3 doesn't seem to have any effect, replace it with -mcode-object-v3 (seems to work at link step as well)


mangupta:
Closing this in favor of #1836.

PR #1553, opened 2019-10-17
fixed if the kernel arguments are passed as NULL

satyanveshd:
Handled the case where a NULL is passed as one of the kernel arguments (instead of nullptr), as part of SWDEV-193444.

### DIFF
--- a/include/hip/hcc_detail/functional_grid_launch.hpp
+++ b/include/hip/hcc_detail/functional_grid_launch.hpp
@@ -43,6 +43,31 @@ inline T round_up_to_next_multiple_nonnegative(T x, T y) {
     return tmp - tmp % y;
 }
 
+//Index & Make Index Sequence Impl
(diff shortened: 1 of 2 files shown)


mangupta:
@satyanveshd Can we also add a test that validates this?

satyanveshd:
Added a test to validate when NULL is passed as a kernel argument.

AlexVlx:
@mangupta can we hold off on merging this? It'd cross streams with a similar change that also mildly improves performance.

AlexVlx:
@mangupta #1686 is the similar change I mentioned, which addresses the same issue in a slightly simpler way, amongst other things., (I nicked the unit test from here since it was better than the one I initially wrote). Perhaps you can determine the sequencing of these two?

mangupta:
replaced by #1686.

PR #1573, opened 2019-10-22
Make CAS loops use the TTAS idiom.

AlexVlx:
For certain atomic ops which lack hardware support we use CAS loops as the implementation. At the moment these loops are TAS, which makes them even more awful than they need to be. This takes the trivial step of using TTAS, which is notably cheaper for us since the first T boils down to a trivial read.


iotamudelta:
Can confirm that this improves end-to-end performance for a relevant workload.

PR #1580, opened 2019-10-24
Fix memcpyParam2D regression

gargrahul:


gargrahul:
@mangupta if we merge #1510, then this PR may not be needed.

mangupta:
Closing this since PR #1510 is being merged which makes this change redundant.

PR #1587, opened 2019-10-24
Fix gcc build on NVCC path

gargrahul:


gargrahul:
Closing in favor of #1661

PR #1591, opened 2019-10-27
Use native support for atomic FADD when address is in LDS

AlexVlx:


mangupta:
@AlexVlx Can we also add a test case that validates this?

AlexVlx:
@mangupta sort of (I'll add something). In order to actually verify this beyond "atomic FPADD works with `__shared__`" we'd need infrastructure that's not available with HIP tests i.e. looking at IR/ISA and checking that we're generating `ds_add`. Since this is a transparent optimisation, just ensuring that atomics to `__shared__` work should suffice though.

AlexVlx:
@mangupta test added.

mangupta:
This test fails on cuda path probably because shared atomics is not supported.

AlexVlx:
They're supported on CUDA just fine. I don't think it's worth the time to investigate the failure though, do we have a macro that identifies NVCC / CUDA compilation?

gargrahul:
> 
> 
> They're supported on CUDA just fine. I don't think it's worth the time to investigate the failure though, do we have a macro that identifies NVCC / CUDA compilation?

#if defined(__HIP_PLATFORM_NVCC__)

AlexVlx:
@gargrahul arigato!

Issue #1571, opened 2019-10-22
[HIPIFY] hipify-clang goes to install_root/ instead of  install_root/bin


Kaveh01:
After commit f19e7c29dfcef3d0c941c977ce6d1448ad4e288a hipify-clang goes to the install_root/
and not to the install_root/bin:

`rm -rf ~/software/hip && rm -rf * && cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_HIPIFY_CLANG=ON -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DHIP_COMPILER=clang   -DCMAKE_INSTALL_PREFIX=/home/kaveh/software/hip ../ && make -j8`

```
$ ls ~/software/hip 
bin  cmake  hipify-clang  include  src
$ ls ~/software/hip/bin/ 
findcode.sh                hipconfig                  hipexamine-perl.sh
finduncodep.sh             hipconvertinplace-perl.sh  hipexamine.sh
hipcc                      hipconvertinplace.sh       hipify-cmakefile
hipcc_cmake_linker_helper  hipdemangleatp             hipify-perl
```

emankov:
Please verify.

Kaveh01:
When I am not specifying "-DBIN_INSTALL_DIR" , it still installs hipify-clang in the install_root:
 
```
$ rm -rf ~/software/hip && rm -rf * && cmake -DCMAKE_BUILD_TYPE=Release
 -DBUILD_HIPIFY_CLANG=ON -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DHIP_COMPILER=clang  
 -DCMAKE_INSTALL_PREFIX=/home/kaveh/software/hip ../ && make -j8 install
...
-- Installing: /home/kaveh/software/hip/include/avxintrin.h
-- Installing: /home/kaveh/software/hip/include/vadefs.h
-- Installing: /home/kaveh/software/hip/include/float.h
-- Installing: /home/kaveh/software/hip/include/avx512vp2intersectintrin.h

$ ls ~/software/hip
bin  cmake  hipify-clang  include  src
$ ls ~/software/hip/bin
findcode.sh                hipconfig                  hipexamine-perl.sh
finduncodep.sh             hipconvertinplace-perl.sh  hipexamine.sh
hipcc                      hipconvertinplace.sh       hipify-cmakefile
[...]

PR #1593, opened 2019-10-28
[HIP][cmake] Move all *_INSTALL_DIR variables up before first add_subdirectory()

emankov:
[REASON]
Those vars (may) used by cmake in subdirectories (#1571)


emankov:
Ping

emankov:
Merging this without review, could not wait any longer as it is a fix for a couple of issues and a show stopper for hipify-clang packaging. All checks have passed.

PR #1607, opened 2019-10-30
[HIP] Correct headers and add missing function templates for hip-clang.

mhbliao:
- Add missing function templates of them in hip-clang. So far, that function template is only available on HCC path.


mhbliao:
could someone merge it?

PR #1610, opened 2019-10-30
Add templates for defining the static members in struct Coordinates

scchan:
Static members x,y,z in Coordinates have only been declared by not
defined.  That means that there isn't an object owning the
"storage" of these static member (even though their size is zero).
This works fine when compiling code with optimization since
references and symbols to these static memebers would get
cleaned up by compiler optimizers.  However, this causes
a problem at no-opt since references and symbols to these
static members were not cleaned up and the final code
object would end up having an extern symbol to these
members.

### DIFF
--- a/include/hip/hcc_detail/hip_runtime.h
+++ b/include/hip/hcc_detail/hip_runtime.h
@@ -244,6 +244,12 @@ struct Coordinates {
   static constexpr Y y{};
   static constexpr Z z{};
 };
+template<typename F>
+__device__ constexpr typename Coordinates<F>::X Coordinates<F>::x __attribute__((__visibility__("hidden")));
+template<typename F>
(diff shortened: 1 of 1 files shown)


AlexVlx:
Don't do this. There are written as they are on purpose, and giving them actual embodiment will almost certainly break ODR. I know what you are trying to fix, and the actual solution there is to modify the RT to ignore symbols that are undefined both on the host and in the code object, as opposed to erroring out. I have pointed this out in the past as something that needs fixing for reasons broader than this.

b-sumner:
Hi @AlexVlx, "when" should this ignoring be done?  Surely at load time we need to complain?

mangupta:
@scchan @AlexVlx is this PR still needed? Or are there any plans to address the concerns.

mangupta:
Closing this since there does not seem to be any plans to merge this right now.

PR #1614, opened 2019-10-31
Updated hipEnvVarDriver to work with Windows

jiabaxie:
hipEnvVarDriver would not run with windows because of the following:

1. Windows didn't recognize unsetenv(). Windows use _putenv.
2. Path to hipEnvVar used forward slashes (/) while windows only recognizes back slashes (\)
3. _popen on windows does not actually return NULL on invalid command, therefore after checking for directed_test/hipEnvVar, it doesn't go to check hipEnvVar at the same level.

Changes have been made to fix these.

### DIFF
--- a/tests/src/hipEnvVarDriver.cpp
+++ b/tests/src/hipEnvVarDriver.cpp
@@ -112,8 +112,6 @@ void getDevicePCIBusNum(int deviceID, char* pciBusID) {
 }
 
 int main() {
-    //extern const char* HIP_VISIBLE_DEVICES_STR;
-    //extern const char* CUDA_VISIBLE_DEVICES_STR;
     unsetenv(HIP_VISIBLE_DEVICES_STR);
     unsetenv(CUDA_VISIBLE_DEVICES_STR);
     
--- a/tests/src/hipEnvVarDriver.cpp
+++ b/tests/src/hipEnvVarDriver.cpp
@@ -20,13 +20,6 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
  * TEST: %t
  * HIT_END
  */
-#ifdef _WIN32
-    #define PLATFORM_NAME "windows"
-#elif __linux__
-    #define PLATFORM_NAME "linux"
-#else
-    #define PLATFORM_NAME "NULL"
-#endif
 
 #include <iostream>
 #include <vector>
@@ -43,13 +36,12 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
 using namespace std;
 
 int getDeviceNumber() {
-    FILE* directed_in;
     FILE* in;
-    string directed_dir;
-    string dir;
     char buff[512];
-
-    if (PLATFORM_NAME == "windows"){
+    const char* directed_dir;
+    const char* dir;
+    
+    if (PLATFORM_NAME == "WINDOWS"){
         directed_dir = "directed_tests\\hipEnvVar -c";
         dir = "hipEnvVar -c";
     } else {
@@ -58,33 +50,31 @@ int getDeviceNumber() {
     }
     
     std::this_thread::sleep_for(std::chrono::milliseconds(10));
-    directed_in = popen(directed_dir.c_str(), "r");
-    if(fgets(buff, 512, directed_in) != NULL){
+    in = popen(directed_dir, "r");
+    if(fgets(buff, 512, in) != NULL){
         cout << buff;
-        pclose(directed_in);
+        pclose(in);
         return atoi(buff);
     }
     //Check at same level
-    in = popen(dir.c_str(), "r");
+    in = popen(dir, "r");
     if(fgets(buff, 512, in) != NULL){
         cout << buff;
         pclose(in);
         return atoi(buff);
     }
     
-    pclose(directed_in);
     pclose(in);
     return 1;
 }
 
 // Query the current device ID remotely to hipEnvVar
 void getDevicePCIBusNumRemote(int deviceID, char* pciBusID) {
-    FILE* directed_in;
     FILE* in;
     
     string directed_dir;
     string dir;
-    if (PLATFORM_NAME == "windows"){
+    if (PLATFORM_NAME == "WINDOWS"){
         directed_dir = "directed_tests\\hipEnvVar -d ";
         dir = "hipEnvVar.exe -d ";
     } else {
@@ -95,10 +85,10 @@ void getDevicePCIBusNumRemote(int deviceID, char* pciBusID) {
     directed_dir += std::to_string(deviceID);
     std::this_thread::sleep_for(std::chrono::milliseconds(10));
     
-    directed_in = popen(directed_dir.c_str(), "r");
-    if(fgets(pciBusID, 100, directed_in) != NULL){
+    in = popen(directed_dir.c_str(), "r");
+    if(fgets(pciBusID, 100, in) != NULL){
         cout << pciBusID;
-        pclose(directed_in);
+        pclose(in);
         return;
     }
     //Check at same level
@@ -108,7 +98,6 @@ void getDevicePCIBusNumRemote(int deviceID, char* pciBusID) {
         pclose(in);
         return;
     }
-    pclose(directed_in);
     pclose(in);
     return;
 }
@@ -123,14 +112,9 @@ void getDevicePCIBusNum(int deviceID, char* pciBusID) {
 }
 
 int main() {
-    if (PLATFORM_NAME == "windows"){
-        _putenv("HIP_VISIBLE_DEVICES=");
-        _putenv("CUDA_VISIBLE_DEVICES=");
-    } else {
-        unsetenv("HIP_VISIBLE_DEVICES");
-        unsetenv("CUDA_VISIBLE_DEVICES");
-    }
-
+    unsetenv(HIP_VISIBLE_DEVICES);
+    unsetenv(CUDA_VISIBLE_DEVICES);
+    
     std::vector<std::string> devPCINum;
     char pciBusID[100];
     // collect the device pci bus ID for all device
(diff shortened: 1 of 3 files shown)


tests/src/hipEnvVarDriver.cpp
```diff
@@ -81,12 +123,17 @@ void getDevicePCIBusNum(int deviceID, char* pciBusID) {
...
 int main() {
-    unsetenv("HIP_VISIBLE_DEVICES");
-    unsetenv("CUDA_VISIBLE_DEVICES");
+    if (PLATFORM_NAME == "windows"){
+        _putenv("HIP_VISIBLE_DEVICES=");
+        _putenv("CUDA_VISIBLE_DEVICES=");
+    } else {
+        unsetenv("HIP_VISIBLE_DEVICES");
+        unsetenv("CUDA_VISIBLE_DEVICES");
+    }
```
emankov: Add a new single define for _putenv/unsetenv in test_common.h. Here use it unconditionally.

tests/src/hipEnvVarDriver.cpp
```diff
@@ -126,8 +173,13 @@ int main() {
...
         // test if CUDA_VISIBLE_DEVICES will be accepted by the runtime
-        unsetenv("HIP_VISIBLE_DEVICES");
-        unsetenv("CUDA_VISIBLE_DEVICES");
+        if (PLATFORM_NAME == "windows"){
+            _putenv("HIP_VISIBLE_DEVICES=");
+            _putenv("CUDA_VISIBLE_DEVICES=");
+        } else {
+            unsetenv("HIP_VISIBLE_DEVICES");
+            unsetenv("CUDA_VISIBLE_DEVICES");
+        }
```
emankov: The same as above.

tests/src/hipEnvVarDriver.cpp
```diff
@@ -32,43 +39,78 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
...
+    
+    string directed_dir;
+    string dir;
+    if (PLATFORM_NAME == "windows"){
+        directed_dir = "directed_tests\\hipEnvVar -d ";
+        dir = "hipEnvVar.exe -d ";
+    } else {
+        directed_dir = "./directed_tests/hipEnvVar -d ";
+        dir = "./hipEnvVar -d ";
+    }
```
emankov: Use const declarations instead.

tests/src/hipEnvVarDriver.cpp
jiabaxie: Hi, I would like clarification from something.
_putenv() and unsetenv() have slightly different arguments used here. Even if I define them, using them unconditionally won't work unless I also define the arguments too. What is preferred?

emankov: Just use const declaration for `HIP_VISIBLE_DEVICES(=)` and `CUDA_VISIBLE_DEVICES(=)` under the same platform dependant conditional define.

jiabaxie:
Hi, I've made the changes requested, and have pushed it here, please let me know if there's any other issues.

tests/src/test_common.h
```diff
@@ -104,8 +104,15 @@ THE SOFTWARE.
...
 #define setenv(x,y,z) _putenv_s(x,y)
+#define unsetenv _putenv
+#define PLATFORM_NAME "WINDOWS"
+#define HIP_VISIBLE_DEVICES "HIP_VISIBLE_DEVICES="
+#define CUDA_VISIBLE_DEVICES "CUDA_VISIBLE_DEVICES="
 #else
 #define aligned_free(x) free(x)
+#define PLATFORM_NAME "OTHER"
+#define HIP_VISIBLE_DEVICES "HIP_VISIBLE_DEVICES"
+#define CUDA_VISIBLE_DEVICES "CUDA_VISIBLE_DEVICES"
```
emankov: 1. Remove redefine: `#define PLATFORM_NAME "..."`, and use convenient `#ifdef _WIN64 ... #else ... #endif`.
2. Use `extern const char* HIP_VISIBLE_DEVICES;` and `const char* HIP_VISIBLE_DEVICES="HIP_VISIBLE_DEVICES"`; instead of defines. Btw, not `std::string` type, cause the header is C.
3. I'd rename` HIP(CUDA)_VISIBLE_DEVICES` to `HIP(CUDA)_VISIBLE_DEVICES_STR` to avoid possible collisions.

jiabaxie: I'm not clear on 1. and 2. in your feedback.
For 1. the defines are to have the conditional statements in hipEnvVarDriver.cpp know which OS is being used. I'm not aware on how to tell with OS it is with just _WIN64. 
For 2. By putting extern const char* HIP_VISIBLE_DEVICES; in test_common.h and then defining const char* HIP_VISIBLE_DEVICES="HIP_VISIBLE_DEVICES"; in hipEnvVarDriver.cpp, When do I declare it to be  "HIP_VISIBLE_DEVICES=" vs "HIP_VISIBLE_DEVICES"? How can I tell which one to use?

jiabaxie:
I'm not entirely sure if these are the changes that you've requested, please let me know if these are ok.

tests/src/hipEnvVarDriver.cpp
```diff
@@ -32,43 +32,74 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
...
+    const char* directed_dir;
+    const char* dir;
+    
+    if (_WIN64 == 1){
+        directed_dir = "directed_tests\\hipEnvVar -c";
+        dir = "hipEnvVar -c";
+    } else {
+        directed_dir = "./directed_tests/hipEnvVar -c";
+        dir = "./hipEnvVar -c";
+    }
```
emankov: Define slash and backslash in test_common.h(.cpp) and use it here unconditionally.

tests/src/hipEnvVarDriver.cpp
```diff
@@ -32,43 +32,74 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
...
+    
+    string directed_dir;
+    string dir;
+    if (_WIN64 == 1){
+        directed_dir = "directed_tests\\hipEnvVar -d ";
+        dir = "hipEnvVar.exe -d ";
+    } else {
+        directed_dir = "./directed_tests/hipEnvVar -d ";
+        dir = "./hipEnvVar -d ";
+    }
```
emankov: Duplication/common code.

tests/src/hipEnvVarDriver.cpp
```diff
@@ -81,12 +112,12 @@ void getDevicePCIBusNum(int deviceID, char* pciBusID) {
...
-    unsetenv("HIP_VISIBLE_DEVICES");
-    unsetenv("CUDA_VISIBLE_DEVICES");
-
+    unsetenv(HIP_VISIBLE_DEVICES_STR);
+    unsetenv(CUDA_VISIBLE_DEVICES_STR);
+    
     std::vector<std::string> devPCINum;
     char pciBusID[100];
-    // collect the device pci bus ID for all devices
+    // collect the device pci bus ID for all device
```
emankov: ?

jiabaxie:
The strs are now made unconditionally, the comment change before was a mistake, I've fixed that too.

tests/src/hipEnvVarDriver.cpp
```diff
@@ -32,43 +32,59 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
...
 
+
 using namespace std;
 
 int getDeviceNumber() {
     FILE* in;
     char buff[512];
-    string str;
+    string directed_dir = "directed_tests" + string(PATH_SEPERATOR_STR) + "hipEnvVar -c";
+    string dir = "." + string(PATH_SEPERATOR_STR) + "hipEnvVar -c";
```
emankov: Make it global const

tests/src/hipEnvVarDriver.cpp
```diff
@@ -32,43 +32,59 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
 #include <thread>
 #include "test_common.h"
 
+
 using namespace std;
 
 int getDeviceNumber() {
     FILE* in;
```
emankov: Here and below use declaration along with the assignment.

tests/src/hipEnvVarDriver.cpp
```diff
@@ -32,43 +32,59 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
 #include <thread>
 #include "test_common.h"
 
+
```
emankov: Formating

tests/src/hipEnvVarDriver.cpp
```diff
@@ -81,9 +97,9 @@ void getDevicePCIBusNum(int deviceID, char* pciBusID) {
 }
 
 int main() {
-    unsetenv("HIP_VISIBLE_DEVICES");
-    unsetenv("CUDA_VISIBLE_DEVICES");
-
+    unsetenv(HIP_VISIBLE_DEVICES_STR);
+    unsetenv(CUDA_VISIBLE_DEVICES_STR);
+
```
emankov: Formatting

tests/src/hipEnvVarDriver.cpp
jiabaxie: What might be the purpose in making it global? I cannot use directed_dir and dir for both getDeviceNumber and getDevicePCIBusNumRemote as they have different arguments (namely, getDeviceNumber requires -c and getDevicePCIBusNumRemote needs -d)

emankov: -c | -d the only difference, all the rest - copy/paste of string constants.

jiabaxie:
directed_dir and dir are now global and const. I cleaned up some code in the two functions as well to avoid dupe code.

PR #1620, opened 2019-11-04
fix race condition in hipEventRecord

jeffdaily:
The following lines were causing the race condition when two threads update the same event:

https://github.com/ROCm-Developer-Tools/HIP/blob/815a601dcba213d5040dab8be7ac824b573cedd2/src/hip_event.cpp#L135-L142

Thread A creates marker 1. 
Thread B creates marker 2 after marker 1. 
Thread B stores marker 2 in Event. 
Thread A stores marker 1 in Event.

The deadlock mentioned in the comment was not observed.  Previously, hipEventQuery would lock the hipStream because of resource cleanup inside the HCC runtime, but this is no longer the case.  I do not believe the deadlock condition in the comment is currently reproducible.

### DIFF
--- a/src/hip_event.cpp
+++ b/src/hip_event.cpp
@@ -44,28 +44,22 @@ void ihipEvent_t::attachToCompletionFuture(const hc::completion_future* cf, hipS
 }
 
 
-std::pair<hipEventStatus_t, uint64_t> ihipEvent_t::refreshEventStatus() {
-    auto ecd = locked_copyCrit();
-    if (ecd._state == hipEventStatusRecording) {
-        bool isReady1 = ecd._stream->locked_eventIsReady(this);
-        if (isReady1) {
-            LockedAccessor_EventCrit_t eCrit(_criticalData);
-
-            if ((eCrit->_eventData._type == hipEventTypeIndependent) ||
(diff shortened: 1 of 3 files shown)


dagamayank:
@mangupta please ensure this makes it to the next ROCm release.

jeffdaily:
Additional issues addressed in last two commits.

First, a possible though unlikely race in `refreshEventStatus` called by `hipEventElapsedTime` if two threads are using the same events.  The event lock is held while checking `is_ready`, then released, then reacquired with the assumption that the event state hasn't changed.

https://github.com/ROCm-Developer-Tools/HIP/blob/514180b83c4860a3a95fa34275f5369192e2a221/src/hip_event.cpp#L47-L52

The next issue is a performance concern in `hipEventQuery`.  The event lock is held during the `locked_copyCrit`, released, and reacquired in `locked_eventIsReady`.  The event lock does not need to be acquired twice, instead we can use the marker directly in the copied critical data.

https://github.com/ROCm-Developer-Tools/HIP/blob/514180b83c4860a3a95fa34275f5369192e2a221/src/hip_event.cpp#L277-L279

PR #1622, opened 2019-11-05
Deprecate HIP Markers

gargrahul:


CMakeLists.txt
```diff
@@ -184,7 +184,10 @@ if(NOT DEFINED COMPILE_HIP_ATP_MARKER)
         set(COMPILE_HIP_ATP_MARKER 0)
     else()
         set(COMPILE_HIP_ATP_MARKER $ENV{COMPILE_HIP_ATP_MARKER})
+        message(WARNING "HIP Markers are deprecated, please use roctracer/rocTX marker APIs.")
     endif()
+else()
+    message(WARNING "HIP Markers are deprecated, please use roctracer/rocTX marker APIs.")
```
mangupta: Do we need the else? We would get the warning even if we didn't enable markers during HIP build.

leofang:
Hi @gargrahul @mangupta Could you shed some light on the deprecation here? Isn't HIP meant to provide a compatibility layer for what's available in CUDA? It seems to imply that downstream libraries need to split the code path for CUDA and ROCm once `hipProfilerStart()` and `hipProfilerStop()` are removed? I think there is no deprecation plan announced for the CUDA counterparts `cudaProfilerStart()` and `cudaProfilerStop()` as of CUDA 11.0, so what's the rationale here? We're maintaining support for both CUDA and ROCm, and I find this deprecation warning both annoying and confusing.

PR #1630, opened 2019-11-06
__half2 should walk like CUDA and talk like CUDA

AlexVlx:
The __half2 comparison intrinsics had surprising behaviour which did not match CUDA, this corrects the oversight.

### DIFF
--- a/include/hip/hcc_detail/hip_fp16.h
+++ b/include/hip/hcc_detail/hip_fp16.h
@@ -1126,8 +1126,7 @@ THE SOFTWARE.
             {
                 auto r = static_cast<__half2_raw>(x).data ==
                     static_cast<__half2_raw>(y).data;
-                return __half2_raw{_Float16_2{
-                    static_cast<_Float16>(r.x), static_cast<_Float16>(r.y)}};
+                return __half2_raw{__builtin_convertvector(-r, _Float16_2)};
             }
             inline
             __device__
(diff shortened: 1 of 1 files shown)


include/hip/hcc_detail/hip_fp16.h
```diff
@@ -1126,53 +1126,47 @@ THE SOFTWARE.
...
             }
             inline
             __device__
             __half2 __hne2(__half2 x, __half2 y)
             {
                 auto r = static_cast<__half2_raw>(x).data !=
                     static_cast<__half2_raw>(y).data;
-                return __half2_raw{_Float16_2{
-                    static_cast<_Float16>(r.x), static_cast<_Float16>(r.y)}};
+                return __builtin_convertvector(-r, _Float16_2);
```
emankov: @AlexVlx, why `-r`?

AlexVlx: Because:
1. "For vector types, these operators return 0 if the specified relation is false and -1 (i.e., all bits set) if the specified relation is true" (http://www.informit.com/articles/article.aspx?p=1732873&seqNum=10);
2. The corresponding half results are set to 1.0 for true, or 0.0 for false (https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH____HALF2__COMPARISON.html#group__CUDA__MATH____HALF2__COMPARISON_1gf62360cbc3cb48077823cc19a9d2dd69).
Clang extended vector types behave like (1), and we want (2)

emankov: @AlexVlx, thanks for your explanation!

AlexVlx: My pleasure!

PR #1643, opened 2019-11-07
Remove redundant declarations.

mhbliao:
- The revised `hip/hip_ext.h` have that declarations.

### DIFF
--- a/include/hip/hcc_detail/hip_runtime.h
+++ b/include/hip/hcc_detail/hip_runtime.h
@@ -518,25 +518,6 @@ hc_get_workitem_absolute_id(int dim)
 #pragma pop_macro("__CUDA__")
 #endif // ndef _OPENMP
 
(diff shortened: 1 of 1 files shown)


mhbliao:
So far, without this patch, several libraries failed to be compiled due to that redundant declarations.

gargrahul:
So apps would need to include hip/hip_ext.h (for now hip/hip_hcc.h) ?

mhbliao:
they needs revising their code to remove the including of hip_hcc.h, which still works so far with deprecated warning message. If any application fails to be compiled after this, it probably includes that hip_hcc.h conditionally with HCC checking. Their needs revising as well to include hip_ext.h directly.

mhbliao:
BTW, even if they includes hip_ext.h directly, we still have the redundant declaration issues (due to the default argument initializers) as they are defined twice for hip-clang.

PR #1644, opened 2019-11-07
moving the extractkernel tool from hcc to hip

scchan:


gargrahul:
@scchan is it possible to add any test that uses this? Otherwise difficult to validate and track changes.

bin/extractkernel
```diff
@@ -0,0 +1,249 @@
...
+  $tools_path_prefix = File::Spec->catfile($ENV{'HCC_HOME'}, "bin");
+  $llvm_objdump = File::Spec->catfile($tools_path_prefix, "llvm-objdump");
+  $clang_offload_bundler = File::Spec->catfile($tools_path_prefix, "clang-offload-bundler");
+}
+else {
+  $tools_path_prefix = dirname(realpath($0));
+  $llvm_objdump = File::Spec->catfile($tools_path_prefix, "llvm-objdump");
+  $clang_offload_bundler = File::Spec->catfile($tools_path_prefix, "clang-offload-bundler");
+  if (!(-f $llvm_objdump)) {
+    $tools_path_prefix = realpath($tools_path_prefix."/../../hcc/bin");
```
pfultz2: This should use the `HCC_HOME` path set by cmake. That is this should be set to `$tools_path_prefix = realpath("@HCC_HOME@")` and then use cmake's `configure_file` to set the file.

pfultz2: It cannot use a relative path because hcc is a separate package and can be installed in a different location completely unrelated to hip.

pfultz2:
Why is this even being moved to hip? It only uses components in the compiler and not any in hip. This should be part of hcc.

gargrahul:
> 
> 
> Why is this even being moved to hip? It only uses components in the compiler and not any in hip. This should be part of hcc.

We have added support of hip-clang in this and hip-clang won't use HCC repo.

pfultz2:
> We have added support of hip-clang in this and hip-clang won't use HCC repo.

But shouldn't this go into the compiler and not into the runtime? It is completely unrelated to the runtime, and directly depends on the components in the compiler.

srinivamd:
Refer https://github.com/ROCmSoftwarePlatform/MIOpen/blob/master/include/miopen/config.h.in#L51
Looks like this change would impact such dependencies in modules.

mangupta:
@scchan What's the plan for the extractkernel tool? Do we still plan to move this to hip repo or to some other package?

PR #1647, opened 2019-11-08
[packaging] Conform to debian package naming rules

mangupta:


pfultz2:
This should create metapackages for a short time(ie 6 months or so) so that `apt install hip_hcc` can still work until all the scripts and dependent packages are updated to use the name.

emankov:
And a few words about it should be said in that case.

mangupta:
@pfultz2 I guess it is fine if we have backward compatibility only for hip_hcc package. Anyone directly installing hip_base, hip_samples or hip_doc would need to fix their scripts to refer to the new packages.

PR #1660, opened 2019-11-16
Remove logics and unit tests for CUDA_VISIBLE_DEVICES support

sunway513:
This PR is to remove the support of CUDA_VISIBLE_DEVICES environment variable support in HIP runtime.
The GPU selection env vars for CUDA path should have been picked up by the CUDA runtime, not in HIP runtime. 
The current implementation was added in the early HIP development stage to work around some limitations, it should be ready to be removed now.


gargrahul:
@sunway513 Your changes are failing on both HCC and NVCC paths. 
I think we may still need CUDA_VISIBLE_DEVICES while testing hipEnvVar tests on NVCC path.

sunway513:
@gargrahul , let me try those tests locally, thanks for the heads up.

gargrahul:
@sunway513 Your last commit fixed HCC path however as I thought we may still need to use CUDA_VISIBLE_DEVICES in hipEnvVar* tests for NVCC path.

sunway513:
@gargrahul , if the hipEnvVar* tests are only for CUDA_VISIBLE_DEVICES for NV path, can we just disable it there?

gargrahul:
> 
> 
> @gargrahul , if the hipEnvVar* tests are only for CUDA_VISIBLE_DEVICES for NV path, can we just disable it there?

@sunway513 , No what I meant was, hipEnvVar* tests use HIP_VISIBLE_DEVICES on HCC path and CUDA_VISIBLE_DEVICES when on NVCC path. We can add a HIP_PLATFORM guard in tests and continue to run them on both HCC and NVCC paths- though they have been working on both paths without that guard so far.

tests/src/hipEnvVar.cpp
```diff
@@ -67,7 +67,6 @@ int main(int argc, char** argv) {
 
     if (!env.empty()) {
         setenv("HIP_VISIBLE_DEVICES", env.c_str(), 1);
-        setenv("CUDA_VISIBLE_DEVICES", env.c_str(), 1);
```
mangupta: The test needs to run on nvcc path as well. So what you probably need is something like
```c++
#ifdef __HIP_PLATFORM_HCC__
setenv("HIP_VISIBLE_DEVICES", env.c_str(), 1);
#else
setenv("CUDA_VISIBLE_DEVICES", env.c_str(), 1);
#endif
```

tests/src/hipEnvVarDriver.cpp
```diff
@@ -82,7 +82,6 @@ void getDevicePCIBusNum(int deviceID, char* pciBusID) {
 
 int main() {
     unsetenv("HIP_VISIBLE_DEVICES");
-    unsetenv("CUDA_VISIBLE_DEVICES");
```
mangupta: same changes as in hipEnvVar.cpp

sunway513:
@mangupta , can you review the changes?

tests/src/hipEnvVarDriver.cpp
```diff
@@ -102,7 +106,6 @@ int main() {
     // query the returned device pci bus number, check if match the database
     for (int i = 0; i < totalDeviceNum; i++) {
         setenv("HIP_VISIBLE_DEVICES", (char*)std::to_string(i).c_str(), 1);
-        setenv("CUDA_VISIBLE_DEVICES", (char*)std::to_string(i).c_str(), 1);
```
gargrahul: @sunway513 This test is still failing on NVCC path. We would need to set CUDA_VISIBLE_DEVICES ( when __HIP_PLATFORM_NVCC__) , everywhere we are setting HIP_VISIBLE_DEVICES.

PR #1665, opened 2019-11-18
Fix mathlib and app builds with hip-clang.

mhbliao:
few critical apps and mathlibs cannot be built without this fix.


aaronenyeshi:
HIP-VDI needs to guard against these declarations

PR #1669, opened 2019-11-18
change hipGetDevice to return hipSuccess error, use nonnull deviceId for return value

srinivamd:
The change modifies hipGetDevice() function to always return hipSuccess and use the outgoing argument, deviceId, when not NULL to return the device ID to the caller. The caller must check the deviceId returned. A -1 is returned on null device context.


srinivamd:
Need to cleanup commits

PR #1670, opened 2019-11-18
hipGetDevice return hipSuccess, return ID if deviceId not null

srinivamd:
The hipGetDevice() will return hipSuccess and use the outgoing argument deviceId to indicate valid device or not (a minus -1 is returned when null device context).
A null outgoing argument, deviceID, will be treated as a dont-care return that can be used in diagnostics/tests.

### DIFF
--- a/src/hip_device.cpp
+++ b/src/hip_device.cpp
@@ -35,16 +35,15 @@ hipError_t hipGetDevice(int* deviceId) {
     hipError_t e = hipSuccess;
 
     auto ctx = ihipGetTlsDefaultCtx();
+    int devid = 0;
 
-    if (deviceId != nullptr) {
-        if (ctx == nullptr) {
-            e = hipErrorInvalidDevice;  // TODO, check error code.
-            *deviceId = -1;
-        } else {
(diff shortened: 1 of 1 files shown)


src/hip_device.cpp
```diff
@@ -35,16 +35,15 @@ hipError_t hipGetDevice(int* deviceId) {
...
-        } else {
-            *deviceId = ctx->getDevice()->_deviceId;
-        }
+    if (ctx == nullptr) {
+        devid = -1;
     } else {
-        e = hipErrorInvalidValue;
+        devid = ctx->getDevice()->_deviceId;
+    }
+    if (deviceId != nullptr) {
```
SarbojitAMD: You can use else part to report 'hipErrorInvalidValue' which was there in the older implementation.

gargrahul:
@srinivamd CUDA doesn't seem to return success in such case. I don't think we need this change.

srinivamd:
@gargrahul @SarbojitAMD closing this PR per @gargrahul suggestion that it is not needed.

PR #1672, opened 2019-11-19
Added check for large blocksize > 256 (SWDEV 201525)

SarbojitAMD:
1. The kernel fails if block size greater than 256 hence added a check to make sure the requested bock size can be launched with given resources.
2. The current check does not cover all the scenarios, I have opened a communication with the compiler team to get more info in order to conclude successful launch of the kernel.

### DIFF
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -145,6 +145,12 @@ hipError_t ihipModuleLaunchKernel(TlsData *tls, hipFunction_t f, uint32_t global
     auto ctx = ihipGetTlsDefaultCtx();
     hipError_t ret = hipSuccess;
 
+    std::size_t blockSize = hip_impl::get_program_state().get_kernattribute(f->_name);
+    if (blockSize < (localWorkSizeX * localWorkSizeY * localWorkSizeZ))
+    {
+        return hipErrorLaunchFailure;
+    }
+
     if (ctx == nullptr) {
         ret = hipErrorInvalidDevice;
 
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -156,14 +156,6 @@ hipError_t ihipModuleLaunchKernel(TlsData *tls, hipFunction_t f, uint32_t global
 
     } else {
 
-        // Kernel launch may fail for (no. of VGPRs X blocksize >= 65536)
-        if(f != nullptr && f->_header != nullptr){
-            uint32_t usedVGPRCount = f->_header->workitem_vgpr_count;
-            if((localWorkSizeX * localWorkSizeY * localWorkSizeZ * usedVGPRCount) >= 65536){
-                return hipErrorInvalidValue;
-            }
-        }
-
         int deviceId = ctx->getDevice()->_deviceId;
         ihipDevice_t* currentDevice = ihipGetDevice(deviceId);
         hsa_agent_t gpuAgent = (hsa_agent_t)currentDevice->_hsaAgent;
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -149,6 +149,15 @@ hipError_t ihipModuleLaunchKernel(TlsData *tls, hipFunction_t f, uint32_t global
         ret = hipErrorInvalidDevice;
 
     } else {
+
+        // Kernel launch may fail for (no. of VGPRs X blocksize >= 65536)
+        if(f != nullptr && f->_header != nullptr){
+            uint32_t usedVGPRCount = f->_header->workitem_vgpr_count;
+            if((localWorkSizeX * localWorkSizeY * localWorkSizeZ * usedVGPRCount) >= 65536){
+                return hipErrorInvalidValue;
+            }
+        }
+
         int deviceId = ctx->getDevice()->_deviceId;
         ihipDevice_t* currentDevice = ihipGetDevice(deviceId);
         hsa_agent_t gpuAgent = (hsa_agent_t)currentDevice->_hsaAgent;
--- a/src/program_state.inl
+++ b/src/program_state.inl
@@ -141,6 +141,7 @@ public:
     }
 };
 
+
 class program_state_impl {
 
 public:
@@ -172,6 +173,10 @@ public:
             std::unordered_map<
                 std::string,
                 std::vector<hsa_executable_symbol_t>>>> kernels;
+
+    std::pair<
+        std::once_flag,
+        std::unordered_map<std::string, std::size_t>> kernattrib;
 
     std::pair<
         std::once_flag,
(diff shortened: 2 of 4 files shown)


src/hip_module.cpp
```diff
@@ -149,6 +149,15 @@ hipError_t ihipModuleLaunchKernel(TlsData *tls, hipFunction_t f, uint32_t global
         ret = hipErrorInvalidDevice;
 
     } else {
+
+        // Kernel launch may fail for (no. of VGPRs X blocksize >= 65536)
+        if(f != nullptr && f->_header != nullptr){
+            uint32_t usedVGPRCount = f->_header->workitem_vgpr_count;
+            if((localWorkSizeX * localWorkSizeY * localWorkSizeZ * usedVGPRCount) >= 65536){
```
srinivamd: Please avoid hardcoded constants in code, instead use const MAX_VGPR_SIZE or the like with description of the reasoning behind that magic value.

gargrahul:
@SarbojitAMD Could you add a test ? This test could pass when it gets hipErrorLaunchFailure.

src/program_state.inl
```diff
@@ -895,6 +1057,16 @@ public:
         return it0->second;
     }
 
+    const std::size_t getKernattribute(std::string kernelName) {
+
+        auto it = get_kernattribute().find(kernelName);
+	if(it == get_kernattribute().end())
+	{
+	    hip_throw(std::runtime_error{"Undefined __global__ function."});
```
srinivamd: How about adding the `kernelName` to the exception string thrown to help test and debug?

src/program_state.inl
```diff
@@ -844,6 +849,163 @@ public:
...
+
+            static constexpr decltype(kernels_md.size()) name_sz{5};
+            dx = kernels_md.find_first_not_of(" '", dx + name_sz);
+
+            auto fn =
+                kernels_md.substr(dx, kernels_md.find_first_of("'\n", dx) - dx);
+            dx += fn.size();
+
+            std::size_t m = kernels_md.find("MaxFlatWorkGroupSize:",dx);
+            auto n = kernels_md.substr(m+21, kernels_md.find_first_of("'\n", m) - m - 21);
```
srinivamd: A description of the syntax of regexp that is being parsed would be helpful. The use of hardcoded constants such as "21" could be replaced by sizeof(MaxFlatWorkGroupSize_STRING) - 1, assuming that is what is meant above, and defining const for the hardcoded strings.

src/program_state.inl
```diff
@@ -844,6 +849,163 @@ public:
...
+            size_t block_size = std::stoul(n);
```
srinivamd: Couldn't stoul throw an exception? Should that be handled here?

src/program_state.inl
```diff
@@ -844,6 +849,163 @@ public:
...
+
+        amd_comgr_metadata_node_t metadata;
+        if (amd_comgr_get_data_metadata(dataIn, &metadata)
+            != AMD_COMGR_STATUS_SUCCESS)
+            return;
+
+        amd_comgr_metadata_node_t kernels_md;
+        if (amd_comgr_metadata_lookup(metadata, "Kernels", &kernels_md)
+            != AMD_COMGR_STATUS_SUCCESS) {
+            if (amd_comgr_metadata_lookup(metadata,
```
srinivamd: Could the nexted if-statements be combined into one?

src/program_state.inl
```diff
@@ -844,6 +849,163 @@ public:
...
+                                          "amdhsa.kernels",
+                                          &kernels_md)
+                != AMD_COMGR_STATUS_SUCCESS)
+                return;
+        }
+
+        size_t kernel_count = 0;
+        if (amd_comgr_get_metadata_list_size(kernels_md, &kernel_count)
+            != AMD_COMGR_STATUS_SUCCESS)
```
srinivamd: It looks like all of the if-statements above with calls to amd_comgr could be combined into one - is there a plan to add debug logs for the different fails?

src/program_state.inl
```diff
@@ -844,6 +849,163 @@ public:
...
+            return;
+
+        for (size_t i = 0; i < kernel_count; i++) {
+            amd_comgr_metadata_node_t kernel_md;
```
srinivamd: There's a `kernel_md` defined outside the for loop also. Using a different name here could help readability.

src/program_state.inl
```diff
@@ -844,6 +849,163 @@ public:
...
+            kernattribute[kernel_name_str] = std::stoul(metadata_to_string(max_flat_workgroup_size));
+
+			if (amd_comgr_destroy_metadata(max_flat_workgroup_size)
+                != AMD_COMGR_STATUS_SUCCESS)
+                continue;
+        }
+
+        if (amd_comgr_destroy_metadata(kernels_md) != AMD_COMGR_STATUS_SUCCESS
+            || amd_comgr_destroy_metadata(metadata) != AMD_COMGR_STATUS_SUCCESS)
+            return;
```
srinivamd: The calls are combined into one if-condition here.
If these are not used in the for-loop above (kind of hard to tell), could this be moved before the for-loop?

src/program_state.inl
```diff
@@ -844,6 +849,163 @@ public:
...
+
+        amd_comgr_release_data(dataIn);
```
srinivamd: Could this also be moved up, say, before the for-loop?

src/program_state.inl
```diff
@@ -844,6 +849,163 @@ public:
...
+            return read_kernattribute_metadata_v2(tmp, dx + 8u, kernattrib); // Skip "Kernels:".
+        }
+    }
+
+    const std::unordered_map<std::string, std::size_t>& get_kernattribute() {
+
+        std::call_once(kernattrib.first, [this]() {
+            for (auto&& name_and_isa_blobs : get_code_object_blobs()) {
+                for (auto&& isa_blobs : name_and_isa_blobs.second) {
+                    for (auto&& blob : isa_blobs.second) {
```
srinivamd: This three-nested for-loops looks daunting, even though it is called once. Anyway to optimize this?

src/hip_module.cpp
```diff
@@ -145,10 +145,17 @@ hipError_t ihipModuleLaunchKernel(TlsData *tls, hipFunction_t f, uint32_t global
...
+    if (blockSize < (localWorkSizeX * localWorkSizeY * localWorkSizeZ))
+    {
+        return hipErrorLaunchFailure;
+    }
+
     if (ctx == nullptr) {
         ret = hipErrorInvalidDevice;
 
     } else {
+
```
srinivamd: How about simply returning after line 155 with the error, and removing the `} else {` unnecessary indentation?

src/hip_module.cpp
```diff
@@ -145,10 +145,17 @@ hipError_t ihipModuleLaunchKernel(TlsData *tls, hipFunction_t f, uint32_t global
     auto ctx = ihipGetTlsDefaultCtx();
     hipError_t ret = hipSuccess;
 
+    std::size_t blockSize = hip_impl::get_program_state().get_kernattribute(f->_name);
+    if (blockSize < (localWorkSizeX * localWorkSizeY * localWorkSizeZ))
+    {
+        return hipErrorLaunchFailure;
```
srinivamd: Could we log a message on error that can be enabled for debug?

src/program_state.inl
```diff
@@ -844,6 +849,163 @@ public:
...
+            if (name != "AMD") continue; // TODO: switch to using NT_AMD_AMDGPU_HSA_METADATA.
+
+            std::string tmp{
+                static_cast<char*>(desc), static_cast<char*>(desc) + desc_size};
+
+            auto dx = tmp.find("Kernels:");
+
+            if (dx == std::string::npos) continue;
+
+            return read_kernattribute_metadata_v2(tmp, dx + 8u, kernattrib); // Skip "Kernels:".
```
srinivamd: Perhaps use something like `sizeof("kernels:_STRING) - 1` instead of `8u` ?

rwvo:
I think this is a terrible solution that should be rejected. Hardware and the HIP runtime allow blocksizes of up to 1024 threads, and blocksizes larger than 256 threads are used routinely. A proper solution would be to fix the compiler such that it no longer "optimizes" code based on the assumption that the blocksize can be at most 256. It is just an incorrect assumption.

The proposed solution will break tons of hipified cuda codes that uses blocksizes > 256.

b-sumner:
@rwvo there is a serious chicken+egg problem here.  If the compiler assumes the worst case (of maximum work group size) by default, then by default the worst case code will be generated, and this will cause performance of existing apps to plummet.  So instead the compiler assumes a more reasonable size and asks developers to say something only whey they go beyond.  If the compiler assumes the worst case then developers will always need to say something unless they are plan to run with the worst case size.

rwvo:
@b-sumner Thanks, I see the point. Ideally, the compiler would optimize for the max blocksize = 256 case, without doing optimizations that give incorrect results for larger blocksizes. Such an ideal may be impossible.

I'm not happy with the solution of returning an error code at runtime if the programmer chooses a larger blocksize and does not specify launch bounds. With this solution, valid CUDA code becomes invalid code when hipified, and if the return code from the kernel launch in the original CUDA code is not tested (which, while not being a good programming practice, happens frequently), then the hipified version may silently lead to incorrect results where the original CUDA code gives correct results.

I would rather see a compilation or linking error, or if that is not possible, something like a "throw std::runtime_error("You need to specify launch bounds for kernels that are launched with blocksizes > 256");". A std::exception cannot be ignored, and the error message points the programmer to the (simple) fix.

AlexVlx:
This being a hard failure is an absolute and unequivocal error. The cardinality of the set of kernels that happen to be dispatched with a block size that is greater than 256, do not specify `__launch_bounds__` but work just fine is strictly greater than that of those that break. This is the wrong way to address the underlying behavioural mismatch, and HIP should absolutely match CUDA here. This does not require damaging the BE or impacting any other languages. I am strongly against this being merged.

SarbojitAMD:
I strongly disagree with the suggestion of HIP runtime throw runtime exception (std::exception) under the check. It is because of the same reason as today we have many applications that are running happily with >256 block size but no __launhc_bounds__ set. Those apps will start seeing failure which will create more noise.
Current change adds the error in the trace hence any user who is debugging such kernel misbehaving issue should be able to see 'hipErrorLaunchFailure' in the trace.

Moreover this approach helps the user to identify the problem easily.

rwvo:
**If** the HIP compiler breaks a CUDA interface by requiring poorly documented additional preconditions, and generates kernels that possibly give incorrect results if they are called with a workgroup size > 256 and without launch bounds, **then** making lots of noise if users use the interface incorrectly is the right thing to do, especially if the noise included information on how to remedy the issue in two minutes per kernel involved. 

Hoping that users notice the incorrect results in the first place, and then inspect an error log that says that a specific kernel returned an error code, and then assuming that the user will figure out the cause of the error code without any hints, will lead to lots of wasted time on the user's end.

**Not breaking a CUDA interface by requiring additional preconditions would be even better**. There may be validity in optimizing for a workgroup size of 256. If this leads to suboptimal performance if kernels are launched with a larger workgroup size without specifying launch bounds on the kernel: fine. On the other hand, if it leads to kernels that may silently give incorrect results: ugh.

b-sumner:
OK, got it.  We're going to be discussing this internally.

SarbojitAMD:
Thanks for all valuable suggestions and feedback
Dropping the PR as of now will resume if anything needs to be done at the runtime level.

PR #1673, opened 2019-11-19
General sync memcpy improvements. Add `hipMemcpyWithStream`

AlexVlx:


iotamudelta:
I've tested this improvement and the new API it delivers in PyTorch. It works and provides a significant end-to-end performance uplift for some critical applications that are dominated by the async+stream sync pattern the new API addresses. I would like to see this released ASAP.

AlexVlx:
@mangupta, you will note that there are no explicit tests for the new interface. If needed we can add this subsequently, but note that all sync memcpy calls now go through the same innermost function, which is used by `hipMemcpyWithStream` as well i.e. all existing tests cover it already.

src/hip_memory.cpp
```diff
@@ -1367,6 +1619,15 @@ hipError_t hipMemcpy3DAsync(const struct hipMemcpy3DParms* p, hipStream_t stream
     return ihipLogStatus(e);
 }
 
+hipError_t hipExtMemcpyWithStream(void* dst, const void* src, size_t sizeBytes,
```
iotamudelta: Is this still needed?

bensander: the new API should be called hipMemcpyWithStream and should have AMD and CUDA implementations.

AlexVlx: Good eyes gents, that's just an accidental leftover though:)

Issue #1675, opened 2019-11-20
HIP vector types do not link properly with a combination of gcc and hcc 


jglaser:
With the development version of HIP (3a7eb694f508517b3389ed8a9295bf41f8d16685), I get linker errors in a project which mixes gcc-compiled objects with hipcc-compiled device methods. The reason is a change in hcc_detail/hip_vector_types.h, which defines e.g. double4 differently with gcc and clang. The linker doesn't recognize they're the same anymore. Reverting back to roc 2.9.

jglaser:
The changes introduced in 4a635ad are really problematic.  For instance, the definition of the `Scalar_accessor` functions as members of the vector types introduces problems with template argument deduction,  in particular when e.g. `double4` members are passed to atomicMax functions or other templates (like in serialization libraries). These functions expect e.g. the double member `.x` to be of data type double, but now it is of data type `hip_impl::Scalar_accessor`. Moreover, some operators like bitwise shift which are not present in CUDA are introduced here and can cause compilation errors. Could these changes be reverted?

PR #1676, opened 2019-11-20
Implement accurate max block size in hipFuncGetAttributes()

### LINKED ISSUE #1662: hipFuncGetAttributes returns inaccurate maxThreadsPerBlock
With the HCC backend, using `hipFuncGetAttributes` on a `__global__` function does not always return an accurate upper bound of the maximum block size. Neither does it return a multiple of the warp size. As a consequence, when iterating over possible kernel parameters allowed by `attr.maxThreadsPerBlock`, I get an error:

`### HCC STATUS_CHECK Error: HSA_STATUS_ERROR_INVALID_ISA (0x100f) at file:mcwamp_hsa.cpp line:1194`

I believe this is because because the code in `hip_module.cpp` hardcodes a fixed number of registers per CU (65536), and does not distinguish between SGPR and VGPR usage of a kernel function. Instead, it should do something more sensible like in `hipOccupancyMaxPotentialBlockSize`.

jglaser:
This PR takes ensures that the `maxThreadsPerBlock` returned by `hipFuncGetAttributes` is both a multiple of the warp size and that the register usage of the maximum block does not exceed the number of available registers.

Currently targeting the 2.9.x branch because I based it off of that branch (see #1675). Can probably be rebased on master.

Fixes #1662


jglaser:
It looks like the calculation was wrong in the first attempt, and I fixed it. however now my own tests are passing.

mangupta:
@jglaser Thanks for the PR. Can you please rebase the PR against the master branch?

jglaser:
@mangupta It would be great if someone could review, thanks. This will be a blocker once we enable CI for HOOMD blue w/HIP https://github.com/glotzerlab/hoomd-blue/pull/541

jglaser:
Any progress on this?

PR #1680, opened 2019-11-21
Fix hip_nvcc package name (underscore not dash)

smsaladi:
I think the package name should have an underscore (not dash):

```bash
(base) ➜ ~ apt-cache search hip-nvcc
(base) ➜  ~ apt-cache search hip_nvcc
hip_nvcc - HIP: Heterogenous-computing Interface for Portability [NVCC]
```


gargrahul:
@smsaladi This change was done in #1647 , to be conformant with debian package naming rules.

smsaladi:
I see -- didn't catch that

PR #1685, opened 2019-11-22
Swallow -x argument to hipcc

jglaser:
Resolve a problem where hipcc cannot be used as a backend for CMake's CUDA langauge, because CMAKE passes the `-x cu` option for all *.cu files, which hipcc complains about. Therefore, this patch makes `hipcc` ignore the `-x` option altogether.

Background: we do not intend to go back to FindHIP's compile framework in our package (HOOMD), as we recently switched from FindCUDA to CMake3's integrated support for the CUDA language. Going forward, it would be good if HIP was a recognized language as well.


jglaser:
This will not work in all circumstances, as "-x=cu" is an alternative way to formulate this flag. I opted to modify our CMake logic to not pass this argument to `hipcc` in the first place.

PR #1686, opened 2019-11-23
Simplify laying out the kernarg + generic tweaks.

AlexVlx:
We don't need to use a dynamically allocated vector for laying out the kernarg, since we already have all the type info we need. Similarly, we do not need to add implicit arguments unless cooperative groups are used.


iotamudelta:
This substantially improves performance in my tests for PyTorch in models with limited GPU queue depth.

src/hip_module.cpp
```diff
@@ -143,135 +143,116 @@ hipError_t ihipModuleLaunchKernel(TlsData *tls, hipFunction_t f, uint32_t global
...
-        if (kernelParams) {
-            if (extra) return hipErrorInvalidValue;
+    std::vector<char> kernargs{};
+    if (kernelParams && *kernelParams) {
+        if (extra) return hipErrorInvalidValue;
 
-            for (auto&& x : f->_kernarg_layout) {
-                const auto p{static_cast<const char*>(*kernelParams)};
+        for (auto&& x : f->_kernarg_layout) {
+            const auto p{static_cast<const char*>(*kernelParams)};
```
gargrahul: @AlexVlx I think it will be good to check *kernelParams here. For example passing void* args[2] = {&Ad, nullptr}; will miss check at line 154 and will lead to seg fault.

AlexVlx: I don't think so. On one hand, the body of that branch is still broken anyway, so anybody using it would have a different set of problems. On the other, passing nullptr to this looks like UB, so you can get anything from nasal demons, segfaults or nuclear armageddon. We're choosing the option that makes the common case fast. You can, of course, follow-up and insert per element checks, just note that those would come at a cost.

gargrahul: Ok but I tried that on CUDA that case and it throws an error however we seg faults.

AlexVlx: Yes...

AlexVlx: My basic point is that if needs be the check can be added at a later date, and since this is UB segfaulting is a perfectly kosher thing to do, orthogonally to what CUDA does or does not do. Indeed, adding the check would be a most excellent subsequent contribution.

src/hip_module.cpp
```diff
@@ -144,137 +144,121 @@ hipError_t ihipModuleLaunchKernel(TlsData *tls, hipFunction_t f, uint32_t gridSi
...
-            }
+            kernargs.insert(
+                kernargs.cend(),
+                round_up_to_next_multiple_nonnegative(
+                    kernargs.size(), x.second) - kernargs.size(),
+                '\0');
+            kernargs.insert(kernargs.cend(), p, p + x.first);
 
-        } 
-        else if (f->_kernarg_layout.size() != 0) {
```
jeffdaily: I think last merge of master into this PR branch lost this check.  That's why the test `directed_tests/gcc/LaunchKernel.tst` is now reporting OTHER_FAULT in the CI.

If I add this case back in, I end up with a seg fault instead, because `extra` is null and is unconditionally dereferenced later.

mangupta:

PR #1687, opened 2019-11-23
Expose support for non-returning atomic FADD.

AlexVlx:


mangupta:
Closing this since this will be handled in rocm-device-libs for now.

PR #1690, opened 2019-11-24
fix linking of vector types with gcc

### LINKED ISSUE #1675: HIP vector types do not link properly with a combination of gcc and hcc

jglaser:
Fix linker issues where incompatible definitions of the vector types are used in gcc and clang.

Now there is a single consistent data type, which either uses the clang `ext_vector_type` extension, or the equivalent gcc `vector_size`.

Eliminate specialization for size 3 data types and round up to nearest power of two. Likely some code could be saved by automatically computing the next power of two instead of manually specifying it as a template argument.

There still are some issues with these data types and C++ projects which assume that the members of, e.g., `double4` are strictly `double`. Implicit type conversion does not work in all situations. This is the case e.g. for serialization libraries (cereal) applied to the HIP vector types, or python bindings (pybind11). The HIP data types are therefore not strictly compatible to the CUDA ones. However, these issues are pretty localized in our project code (at least) and could be fixed there.

Fixes #1675

### DIFF
--- a/include/hip/hcc_detail/hip_vector_types.h
+++ b/include/hip/hcc_detail/hip_vector_types.h
@@ -42,6 +42,7 @@ THE SOFTWARE.
 #endif
 
 #if defined(__cplusplus)
+    #include <array>
     #include <iosfwd>
     #include <type_traits>
 
@@ -464,7 +465,11 @@ THE SOFTWARE.
         inline __host__ __device__
         HIP_vector_type(Us... xs) noexcept
         {
-            new (&data) Native_vec_{static_cast<T>(xs)...};
+            #if __has_extension(ext_vector_type)
+                new (&data) Native_vec_{static_cast<T>(xs)...};
+            #else
+                new (&data) std::array<T, rank>{static_cast<T>(xs)...};
+            #endif
         }
         inline __host__ __device__
         HIP_vector_type(const HIP_vector_type&) = default;
--- a/include/hip/hcc_detail/hip_vector_types.h
+++ b/include/hip/hcc_detail/hip_vector_types.h
@@ -52,9 +52,11 @@ THE SOFTWARE.
     
     namespace std {
         template<typename T, typename U, unsigned int n>
-        struct is_integral<Scalar_accessor<T, U, n>> : is_integral<T> {};
+        struct is_integral<hip_impl::Scalar_accessor<T, U, n>>
+            : is_integral<T> {};
         template<typename T, typename U, unsigned int n>
-        struct is_floating_point<Scalar_accessor<T, U, n>> : is_floating_point<T> {};
+        struct is_floating_point<hip_impl::Scalar_accessor<T, U, n>> 
+            : is_floating_point<T> {};
     } // Namespace std.
     
     namespace hip_impl {
--- a/include/hip/hcc_detail/hip_vector_types.h
+++ b/include/hip/hcc_detail/hip_vector_types.h
@@ -34,17 +34,19 @@ THE SOFTWARE.
 
 #include "hip/hcc_detail/host_defines.h"
 
-#if !defined(_MSC_VER) || __clang__
-#if defined(__clang__)
-    #define __NATIVE_VECTOR__(n, ...) __attribute__((ext_vector_type(n)))
+#if __has_attribute(ext_vector_type) || __has_attribute(vector_size)
+#if __has_attribute(ext_vector_type)
+    #define __NATIVE_VECTOR__(n, T, n_power_of_two) __attribute__((ext_vector_type(n_power_of_two),aligned(n_power_of_two*sizeof(T))))
+#elif __has_attribute(vector_size)
+    #define __NATIVE_VECTOR__(n, T, n_power_of_two) __attribute__((vector_size(n_power_of_two*sizeof(T)),aligned(n_power_of_two*sizeof(T))))
 #endif
 
-#if defined(__cplusplus) && defined(__clang__)
+#if defined(__cplusplus)
     #include <iosfwd>
     #include <type_traits>
 
     namespace hip_impl {
-        template<typename T, typename Vector, unsigned int idx>
+        template<typename T, unsigned int n, unsigned int n_power_of_two, unsigned int idx>
         struct Scalar_accessor {
             struct Address {
                 const Scalar_accessor* p;
@@ -67,6 +69,7 @@ THE SOFTWARE.
                     return &reinterpret_cast<T*>(
                         const_cast<Scalar_accessor*>(p))[idx];
                 }
+
             };
 
             friend
@@ -87,23 +90,14 @@ THE SOFTWARE.
             }
 
             // Idea from https://t0rakka.silvrback.com/simd-scalar-accessor
+            typedef T Vector __NATIVE_VECTOR__(n, T, n_power_of_two);
             Vector data;
 
             __host__ __device__
             operator T() const noexcept { return data[idx]; }
-            __host__ __device__
-            operator T() const volatile noexcept { return data[idx]; }
 
             __host__ __device__
-            operator T&() noexcept {
-                return reinterpret_cast<
-                    T (&)[sizeof(Vector) / sizeof(T)]>(data)[idx];
-            }
-            __host__ __device__
-            operator volatile T&() volatile noexcept {
-                return reinterpret_cast<
-                    volatile T (&)[sizeof(Vector) / sizeof(T)]>(data)[idx];
-            }
+            operator T() const volatile noexcept { return data[idx]; }
 
             __host__ __device__
             Address operator&() const noexcept { return Address{this}; }
@@ -113,7 +107,8 @@ THE SOFTWARE.
                 data[idx] = x;
 
                 return *this;
-            }
+                }
+
             __host__ __device__
             volatile Scalar_accessor& operator=(T x) volatile noexcept {
                 data[idx] = x;
@@ -179,7 +174,7 @@ THE SOFTWARE.
                 typename std::enable_if<
                     std::is_convertible<U, T>{}>::type* = nullptr>
             __host__ __device__
-            Scalar_accessor& operator/=(T x) noexcept {
+            Scalar_accessor& operator/=(U x) noexcept {
                 data[idx] /= x;
                 return *this;
             }
@@ -245,11 +240,14 @@ THE SOFTWARE.
 
     template<typename T>
     struct HIP_vector_base<T, 1> {
-        typedef T Native_vec_ __NATIVE_VECTOR__(1, T);
+        enum { n = 1,
+               n_power_of_two = 1
+        };
+        typedef T Native_vec_ __NATIVE_VECTOR__(n, T, n_power_of_two);
 
         union {
             Native_vec_ data;
-            hip_impl::Scalar_accessor<T, Native_vec_, 0> x;
+            hip_impl::Scalar_accessor<T, n, n_power_of_two, 0> x;
         };
 
         using value_type = T;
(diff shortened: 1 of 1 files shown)


jglaser:
it would be good to know which checks are failing so that I can fix. I can confirm that with the new accessor methods and the correspondingly improved assembly code for the vector data types we are getting quite a significant speed bump in MD simulations which use them a lot.

AlexVlx:
Thank you for the PR. LGTM in general, but there are a couple of issues:
- the 3-component specialisation is not there by accident, but rather because other use cases wanted a 3-component vector of T to have size equal to 3 * sizeof(T) i.e. they preferred suboptimal alignment if it meant spatial savings. Note that this specialisation was added relatively recently, and under duress;
- the address-of overloads are necessary for some relatively gnarly use cases that would be difficult to modify
- I think that relying on this sort of interop between Clang and GCC is fraught with peril, since 'vector_size' and 'ext_vector_type' are not guaranteed to be interoperable by definition.
Note that there is a relatively cleaner way to define the native vector even for GCC, which prevents the need for an enum and an extra non-type template argument, which was used before removing the GCC path - might be worth taking a peek.

AlexVlx:
@jglaser sadly, `vector_size` is not what we want to do in general, since `ext_vector_type` is more fully featured in terms of the basis functions that are exposed, and it is also the preferred, actively developed, extension in Clang, which is our main platform. It is somewhat unclear what the expected benefit is with inter-linking GCC and Clang compiled objects that expose non-trivial types in interfaces (I assume that this is tied to the NVIDIA centric model where there is a separate device compiler), but it would be somewhat preferable to just build the entirety of the codebase with HCC/HIP-Clang, since those are fully capable end-to-end compilers. 

Is there an externally visible version of HOOMD that can be built using HIP, which exposes the issue that you are trying to address?

jglaser:
> @jglaser sadly, `vector_size` is not what we want to do in general, since `ext_vector_type` is more fully featured in terms of the basis functions that are exposed, and it is also the preferred, actively developed, extension in Clang, which is our main platform. 

Hmm, ok. What are the extra basis functions that are exposed and how is it more actively developed? `vector_size` has been around in clang since at least version 3.4.


> It is somewhat unclear what the expected benefit is with inter-linking GCC and Clang compiled objects that expose non-trivial types in interfaces (I assume that this is tied to the NVIDIA centric model where there is a separate device compiler), but it would be somewhat preferable to just build the entirety of the codebase with HCC/HIP-Clang, since those are fully capable end-to-end compilers.
>
OK, in that case I believe only a minor fix of the implicit conversion operators for the vector types  is necessary to make HOOMD compile with clang. They always need to return references, not values. We would then give up on gcc when compiling for an AMD backend. I can modify the PR accordingly to reflect that minimal change.

> Is there an externally visible version of HOOMD that can be built using HIP, which exposes the issue that you are trying to address?

Yes, https://github.com/glotzerlab/hoomd-blue branch `hip`. Compile with -D ENABLE_HIP=ON.

AlexVlx:
An example of something not exposed by vector_size is the .{x || y || z || w} access syntax. Also, whilst it is made available for compatibility with GCC, that does not entail focused development.

Converting an accessor to a reference is only there for the aforementioned gnarly code, and it leads to suboptimal codegen. They would essentially negate any benefit from the whole Scalar_accessor ritual, so it might as well be removed (but then suboptimal code always obtains). So it would be preferable to only rely on that as a last resort - could you say more about what having only convert-to-reference would fix (possibly via email if it is more comfortable).

jglaser:
> An example of something not exposed by vector_size is the .{x || y || z || w} access syntax. Also, whilst it is made available for compatibility with GCC, that does not entail focused development.
> 
> Converting an accessor to a reference is only there for the aforementioned gnarly code, and it leads to suboptimal codegen. They would essentially negate any benefit from the whole Scalar_accessor ritual, so it might as well be removed (but then suboptimal code always obtains). So it would be preferable to only rely on that as a last resort - could you say more about what having only convert-to-reference would fix (possibly via email if it is more comfortable).

OK, I understand. We have a few places in the code where functions take vector types by reference, such as this function signature:

```
   //! Construct a vec3 from a Scalar3
    /*! \param a Scalar3 to copy
        This is a convenience function for easy initialization of vec3s from hoomd memory data structures
    */
    DEVICE explicit vec3(const Scalar3& a) : x(a.x), y(a.y), z(a.z)
        {
        }
```

these do not compile with current HIP master. Here, `Scalar3` is a typedef to `double3` or `float3`, and `vec3` is an internal data type that supports some math operations. I understand you in the sense that these function signatures should *NOT* be used for performance reasons anyway. When I remove the `&`, the code indeed compiles and works with the master branch of HIP and with `clang++` as a host compiler (but of course not with GCC).

I think the intention to not support vector types across the gcc/clang interface could be documented explicitly, as it breaks code that works with CUDA, but apart from this, this PR can then be ignored.

AlexVlx:
Hmm, I would actually expect that to work. Would it be possible to share the error that you are observing? I suspect I know what may be happening and where the challenge lies.

As for the latter point, even though I am somewhat opposed to it on the grounds of it potentially leading to arcane errors, the train has somewhat left the station in what regards that sort of expected interop. So we will probably converge on something close to your first take in this PR - let's first ensure that we address all of your issues.

jglaser:
OK, this issue occurs if I revert the `Scalar_accessor` interface to the following:

```
            __host__ __device__
            operator T() const noexcept { return data[idx]; }
            __host__ __device__
            operator T() const volatile noexcept { return data[idx]; }

             __host__ __device__
            operator T&() noexcept {
                return reinterpret_cast<
                    T (&)[sizeof(Vector) / sizeof(T)]>(data)[idx];
            }
            __host__ __device__
            operator volatile T&() volatile noexcept {
                return reinterpret_cast<
                    volatile T (&)[sizeof(Vector) / sizeof(T)]>(data)[idx];
            }
```

(note, this interface is different from current HEAD of `fix_vector_types_gcc`).

Interestingly, I get an error only with gcc 7.4.0, but not with clang 10. The error is:

```/home/michigan/hoomd-blue/hoomd/VectorMath.h:744:56: error: passing 'const hip_impl::Scalar_accessor<double, 4, 3>' as 'this' argument discards qualifiers [-fpermissive]
     DEVICE explicit quat(const Scalar4& a) : s(a.x), v(vec3<Real>(a.y, a.z, a.w))
                                                        ^~~~~~~~~~~~~~~~~~~~~~~~~
```

I am not sure which of the two compilers is right, here. Maybe it's a gcc bug, and I can test with different versions.

jglaser:
OK, this issue occurs if I revert the `Scalar_accessor` interface to the following:

```
            __host__ __device__
            operator T() const noexcept { return data[idx]; }
            __host__ __device__
            operator T() const volatile noexcept { return data[idx]; }

             __host__ __device__
            operator T&() noexcept {
                return reinterpret_cast<
                    T (&)[sizeof(Vector) / sizeof(T)]>(data)[idx];
            }
            __host__ __device__
            operator volatile T&() volatile noexcept {
                return reinterpret_cast<
                    volatile T (&)[sizeof(Vector) / sizeof(T)]>(data)[idx];
            }
```

(note, this interface is different from current HEAD of `fix_vector_types_gcc`).

Interestingly, I get an error only with gcc 7.4.0, but not with clang 10. The error is:

```/home/michigan/hoomd-blue/hoomd/VectorMath.h:744:56: error: passing 'const hip_impl::Scalar_accessor<double, 4, 3>' as 'this' argument discards qualifiers [-fpermissive]
     DEVICE explicit quat(const Scalar4& a) : s(a.x), v(vec3<Real>(a.y, a.z, a.w))
                                                        ^~~~~~~~~~~~~~~~~~~~~~~~~
```

I am not sure which of the two compilers is right, here. Maybe it's a gcc bug, and I can test with different versions.

jglaser:
Note, the offending `vec3` constructor is in this case:
```
    //! Construct a vec3
    /*! \param _x x-component
        \param _y y-component
        \param _z z-component
    */
    DEVICE vec3(const Real& _x, const Real& _y, const Real& _z) : x(_x), y(_y), z(_z)
        {
        }
```

AlexVlx:
I believe that this is because the vec3 constructor that takes 3-elements takes them by reference-to-const, which leads to an ambiguity as both conversion operators get an equal score. I'd have to think through why Clang and GCC differ in their assessment. I would suggest (without assessing how noisy this'd be, so take it with a grain of salt) that changing things to `vec3<Real>((Real)a.y, (Real)a.z, (Real)a.w)` would suffice to decide the matter, and it would not negatively impact the CUDA pathway. Alternatively, since vec3 is intended to take built-in arithmetic types (AFAICT) as its component types, it would seem reasonable for its constructors to take their arguments by value, as this'd never pessimise things (even for a double it's still 8-bytes) and could improve some cases (e.g. a float is only 4 bytes so passing it via reference is mildly excessive). Granted, this optimises away more often than not, but it'd also address the present concern.

jglaser:
OK, that fixes it of course. I updated the branch with more minimal changes to support clang/GCC interop, along the lines what we agreed upon.

AlexVlx:
@jglaser this is significantly tidier, thank you. I do believe that there may be a slightly more parsimonious formulation of the idea - may I commit to your branch directly?

jglaser:
Yes, of course, it should allow "edits from maintainers", so please go ahead. Be aware that

1. the `__attribute__((vector_size(N)))` in gcc only works on a `typedef` if the variable is defined in the same scope, it is **not** passed along with the typedef, e.g., through a template parameter,

2. some rearrangement of the three-component native vector types was necessary, because the return type of the comparison operator (`operator==`) for those was returning a `vector_size(3*sizeof(int))` which is a non-power of two and leads to a compile error in gcc.

AlexVlx:
Thanks. I'm aware of the sad issue with the attribute grammar / parsing in GCC. My preference would be to not deal with `vector_size` at all - I will suggest an alternate formulation which hopefully addresses your issue, and let's progress from there (I've noticed the copy ctor issue as well, mea culpa, that was my oversight).

AlexVlx:
@jglaser I've pushed an alternate formulation that appears to address the problem. It would be neat if you'd assess it and see if it addresses the concern expressed herein.

jglaser:
the alternate approach compiles and runs, too. On a side note, do you have a reason to not use `__attribute__((aligned(sizeof(T)*n)))`? It feels a little unsafe without it. There'll be conflicts in merging this up into #1693 ... in which we cannot simply 'assign' the C style arrays inside the constructors anymore, these have to be copied element by element. I'd like to leave it up to you how to fix the copy constructors, then.

AlexVlx:
The copy ctor matter is next after this gets merged, and yes, it's on my plate to fix since I'm to blame for the initial oversight. Thank you for your patience with iterating through this.

AlexVlx:
@mangupta, do you think it would be possible to merge this so that I may fix this gentleman's asignment issue? As far as I can tell, these are the two remaining blockers for the HOOMD-on-HIP effort.

mangupta:
> 
> 
> @mangupta, do you think it would be possible to merge this so that I may fix this gentleman's asignment issue? As far as I can tell, these are the two remaining blockers for the HOOMD-on-HIP effort.

We have a pile up of over 20 PRs trying to make their way through staging but not making through due to various regressions. So for the past couple of days I am merging in only those PRs that address any of the regressions. We look close to getting things moving again. I will merge in all PRs that look good at that time. I apologize for the delay, but trying to balance sanity across multiple teams here. I will definitely merge this PR within the next 24-36 hours.

AlexVlx:
That's more than fine, thank you.

AlexVlx:
@jglaser I've updated this to hopefully address the assignment issue as well - could you please check that after the latest changes it still does the right thing for HOOMD? Thank you.

jglaser:
Hi @AlexVlx and @mangupta, thanks for reviewing & merging. HOOMD-blue compiles with `gcc` on hip master.

PR #1691, opened 2019-11-24
 Enable hip-clang based testing for Github PR

ansurya:
SWDEV-207903:
These changes are to test HIP-HCC with HIP-clang compiler for every Github PR.
* Update HIP repo jenkins file to run unit tests on hip-clang path.
* Disabled tests that are failing temporarily on clang path.


CMakeLists.txt
```diff
@@ -296,7 +296,7 @@ if(HIP_PLATFORM STREQUAL "hcc")
     endif()
     add_library(hip_hcc SHARED ${SOURCE_FILES_RUNTIME})
     add_library(hip_hcc_static STATIC ${SOURCE_FILES_RUNTIME})
-    if(HIP_COMPILER STREQUAL "hcc")
+    if(HIP_COMPILER STREQUAL "hcc" OR HIP_COMPILER STREQUAL "clang")
```
aaronenyeshi: Why are we adding HIP-Clang here? I didn't add clang here because it shouldn't have hc_am, and also it doesn't support hiprtc. We have our own implementation in VDI.

Jenkinsfile
```diff
@@ -386,6 +386,40 @@ rocm_head:
     */
   }
 },
+rocm_clang:
+{
+  node('hip-rocm')
+  {
+    String hcc_ver = 'rocm-head'
+    String from_image = 'ci_test_nodes/rocm-clang/ubuntu-16.04:latest'
```
aaronenyeshi: Are you building hip-clang from scratch?

tests/src/runtimeApi/module/hipFuncGetAttributes.cpp
```diff
@@ -18,7 +18,7 @@ THE SOFTWARE.
 */
 
 /* HIT_START
- * BUILD: %t %s ../../test_common.cpp NVCC_OPTIONS -std=c++11
+ * BUILD: %t %s ../../test_common.cpp NVCC_OPTIONS -std=c++11 EXCLUDE_HIP_PLATFORM clang
```
aaronenyeshi: Why are we skipping these tests? they should be passing in HIP-Clang.

yxsamliu:
this is building hip-clang with HIP/HCC runtime? Is it possible to build hip-clang with HIP/VDI runtime?

gargrahul:
> 
> 
> this is building hip-clang with HIP/HCC runtime? Is it possible to build hip-clang with HIP/VDI runtime?

I think that would be done once HIP-CI is taken over by dev-ops - there is another ticket for that.

CMakeLists.txt
ansurya: CMakeLists.txt is currently being used only for building HIP-HCC. To enabling testing of HIP-HCC with HIP-clang compiler, enabled generation of hiprtc for clang path.

Jenkinsfile
ansurya: New docker image with all libraries required to test HIP-HCC with HIP-clang compiler.

tests/src/runtimeApi/module/hipFuncGetAttributes.cpp
ansurya: These tests currently fail when compiled with HIP-clang compiler with HIP-HCC runtime. Even though testcases are shared , HIT environment is not being using in HIP-VDI. Disabled tests temporarily on HIP-HCC path.

mangupta:
Closing this since we plan on redoing this.

PR #1693, opened 2019-11-25
Fix vector types accessor assignment

jglaser:
This fixes a bug where assignment of a single vector component assigns the entire vector.

The behavior was already anticipated [here](https://t0rakka.silvrback.com/simd-scalar-accessor) (see the big WARNING! paragraph)

For better comparability, I am separating this PR from #1690, which should be merged first, but am including its changes in here.


AlexVlx:
Fixed directly in #1690.

PR #1697, opened 2019-11-26
fix hipMempcy precondition, test for zero size first

jeffdaily:


jeffdaily:
@gargrahul please review.

PR #1698, opened 2019-11-26
[hip] refactoring hipExtLaunchMultiKernelMultiDevice API

asalmanp:
[Background] it was found that if lazy linking used for a library that calls  hipExtLaunchMultiKernelMultiDevice API then this API can get the wrong program_state object for looking up device kernels leading to a "No device code available" error in this API.

To fix this issue, the API was refactored to be inline and get and pass the correct program_state to an internal hip API to request a multi-device kernel launch.

### DIFF
--- a/include/hip/hcc_detail/functional_grid_launch.hpp
+++ b/include/hip/hcc_detail/functional_grid_launch.hpp
@@ -34,6 +34,9 @@ THE SOFTWARE.
 #include <type_traits>
 #include <utility>
 
+hipError_t ihipExtLaunchMultiKernelMultiDevice(hipLaunchParams* launchParamsList, int numDevices,
+                                               unsigned int flags, hip_impl::program_state& ps);
+
 #pragma GCC visibility push(hidden)
 
 namespace hip_impl {
@@ -176,4 +179,15 @@ void hipLaunchKernelGGL(F kernel, const dim3& numBlocks, const dim3& dimBlocks,
                                      numBlocks, dimBlocks, sharedMemBytes,
                                      stream, &config[0]);
 }
+
+inline
+__attribute__((visibility("hidden")))
+hipError_t hipExtLaunchMultiKernelMultiDevice(hipLaunchParams* launchParamsList,
+                                              int  numDevices, unsigned int  flags) {
+    hip_impl::hip_init();
+    auto& ps = hip_impl::get_program_state();
+    return ihipExtLaunchMultiKernelMultiDevice(launchParamsList, numDevices, flags, ps);
+
+}
+
 #pragma GCC visibility pop
--- a/include/hip/hcc_detail/hip_runtime_api.h
+++ b/include/hip/hcc_detail/hip_runtime_api.h
@@ -2960,6 +2960,7 @@ hipError_t hipOccupancyMaxActiveBlocksPerMultiprocessor(
 hipError_t hipOccupancyMaxActiveBlocksPerMultiprocessorWithFlags(
    uint32_t* numBlocks, hipFunction_t f, uint32_t blockSize, size_t dynSharedMemPerBlk, unsigned int flags);
 
+#if defined(__clang__) && defined(__HIP__)
 /**
  * @brief Launches kernels on multiple devices and guarantees all specified kernels are dispatched
  * on respective streams before enqueuing any other work on the specified streams from any other threads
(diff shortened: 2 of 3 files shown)


include/hip/hcc_detail/functional_grid_launch.hpp
```diff
@@ -176,4 +179,15 @@ void hipLaunchKernelGGL(F kernel, const dim3& numBlocks, const dim3& dimBlocks,
                                      numBlocks, dimBlocks, sharedMemBytes,
                                      stream, &config[0]);
 }
+
+inline
+__attribute__((visibility("hidden")))
+hipError_t hipExtLaunchMultiKernelMultiDevice(hipLaunchParams* launchParamsList,
+                                              int  numDevices, unsigned int  flags) {
+    hip_impl::hip_init();
+    auto& ps = hip_impl::get_program_state();
```
srinivamd: Isn't  hip_impl::hip_init() required whether or not stub library is used?

wenkaidu:
What is the impact of removal of ihipLogStatus? Can we still get HIP API trace from this call?

include/hip/hcc_detail/hip_runtime_api.h
```diff
@@ -2974,6 +2975,7 @@ hipError_t hipOccupancyMaxActiveBlocksPerMultiprocessorWithFlags(
 hipError_t hipExtLaunchMultiKernelMultiDevice(hipLaunchParams* launchParamsList,
```
gargrahul: So with this change, hipExtLaunchMultiKernelMultiDevice - as a C API won't work?

srinivamd: Is that what  `#if defined(__clang__) && defined(__HIP__)`  does? 
Also, the #if cpp pattern in other places seem to be different. For example, https://github.com/ROCm-Developer-Tools/HIP/blob/master/include/hip/hip_common.h#L29. Should there be an `#elif` check for `__HCC__` also?

include/hip/hcc_detail/functional_grid_launch.hpp
asalmanp: hip_impl::hip_init() should be called at the beginning of every HIP API regardless of using stub library or not. In the previous implementation of  the hipExtLaunchMultiKernelMultiDevice API the HIP_INIT_API() macro was doing the initialization by calling hip_impl::hip_init().

include/hip/hcc_detail/hip_runtime_api.h
asalmanp: @srinivamd  The "hip_runtime_api.h" header is also used by HIP-CLANG and the current code makes sure that the definition of hipExtLaunchMultiKernelMultiDevice is untouched for HIP-CLANG. 
There is no #elf __HCC__ required here because the hipExtLaunchMultiKernelMultiDevice is defined in "functional_grid_launch.h" header for HIP-HCC and this header is not used in HIP-CLANG.

scchan: @gargrahul yes, this won't have an C API on HIP/HCC.

scchan: @srinivamd that guard means the code path is reserved for the hip-clang implementation

asalmanp:
> What is the impact of removal of ihipLogStatus? Can we still get HIP API trace from this call?
@wenkaidu a call to hipExtLaunchMultiKernelMultiDevice wont be explicitly appear in the API trace after this change but you will still get "hipLaunchKernel" message gets printed for the number of devices used in the hipExtLaunchMultiKernelMultiDevice for the launch.

For example if you call hipExtLaunchMultiKernelMultiDevice to launch kernels into 4 devices then in the hip API trace you will see that hipLaunchKernel gets printed 4 times and for each entry, the name of the kernel and it's launch parameters will be printed for debugging purpose.

include/hip/hcc_detail/functional_grid_launch.hpp
srinivamd: Should similar changes be applied to functions like `hipLaunchCooperativeKernel()`?

asalmanp: @srinivamd yes similar change would be required for hipLaunchCooperativeKernel and hipLaunchCooperativeKernelMultiDevice APIs which will be done in a separate PR later as these APIs are not ready yet and they are work in progress.

PR #1699, opened 2019-11-28
Fix __hip_is_shared implementation for HIP-Clang.

mhbliao:
- Still prefer `__builtin_amdgcn_is_shared` if available. That enables
  always-false branches due to checking `__hip_is_shared`. It avoids
  generating invalid `addrspacecast`. Also, that improves the code
  quality a lot.

### DIFF
--- a/include/hip/hcc_detail/hip_atomic.h
+++ b/include/hip/hcc_detail/hip_atomic.h
@@ -77,14 +77,17 @@ __device__
 inline
 bool __hip_is_shared(const __attribute__((address_space(0))) void* ptr) noexcept
 {   // TODO: this is ersatz for __builtin_amdgcn_is_shared.
-    #if defined(__HIP_DEVICE_COMPILE__)
-        const unsigned int gp = reinterpret_cast<unsigned long long>(ptr);
+#if defined(__HIP_DEVICE_COMPILE__)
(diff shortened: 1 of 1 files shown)


mhbliao:
It's verified locally with PyTorch build. Otherwise, PyTorch build failed.

mangupta:
The PR is not needed anymore since #1701 is reverting the changes of PR #1591 and PR #1695.

include/hip/hcc_detail/hip_atomic.h
```diff
@@ -74,18 +74,10 @@ float atomicAdd_impl(float* address, float val)
...
-        const unsigned int gp = reinterpret_cast<unsigned long long>(ptr);
-
-        return
-            gp ==(__builtin_amdgcn_s_getreg((15 << 11) | (16 << 6) | 15) << 16);
-    #else
-        return false;
-    #endif
-}
+bool __hip_is_shared(const __attribute__((address_space(0))) void*) noexcept
+    asm("llvm.amdgcn.is.shared");
```
mhbliao: why we cannot use builtin in HCC? If flat-addrspace is missing in HCC, it should enable it. Directly using LLVM IR intrinsics should not be encouraged considering they are not the interface at language level.

PR #1700, opened 2019-11-28
add error checking for code object loading

scchan:
Errors from these API calls should be checked and it would help to debug some of the "missing kernel/metadata" errors.

### DIFF
--- a/src/program_state.inl
+++ b/src/program_state.inl
@@ -398,13 +398,19 @@ public:
                                            move(file), move(tmp));
         }
 
-        hsa_code_object_reader_create_from_memory(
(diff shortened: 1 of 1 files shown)


src/program_state.inl
```diff
@@ -398,13 +398,19 @@ public:
                                            move(file), move(tmp));
         }
 
-        hsa_code_object_reader_create_from_memory(
-            it->first.data(), it->first.size(), it->second.get());
+        auto check_hsa_error = [](hsa_status_t s) {
```
emankov: LGTM, otherwise `check_hsa_co_error`

scchan:
@mangupta could we merge this?

PR #1703, opened 2019-11-30
[packaging] Generate meta packages for old package names

mangupta:
This is to allow existing scripts to continue working until they switch
to the new package names


mangupta:
Closing this since we are reverting package name change for now via #1688.

PR #1707, opened 2019-12-03
Fix hipMemcpyWithStream

gargrahul:


tests/src/runtimeApi/memory/hipMemcpyWithStream.cpp
```diff
@@ -0,0 +1,56 @@
...
+    HIPCHECK(hipStreamCreate(&stream));
+
+    HIPCHECK(hipMemcpyWithStream(A_d, A_h, Nbytes, hipMemcpyHostToDevice, stream));
+    HIPCHECK(hipMemcpyWithStream(B_d, B_h, Nbytes, hipMemcpyHostToDevice,stream));
+    hipLaunchKernelGGL(HipTest::vectorADD, dim3(blocks), dim3(threadsPerBlock), 0, stream,
+                        static_cast<const int*>(A_d), static_cast<const int*>(B_d), C_d, N);
+    HIPCHECK(hipStreamSynchronize(stream));
+    HIPCHECK(hipMemcpy(C_h, C_d, Nbytes, hipMemcpyDeviceToHost));
+    HipTest::checkVectorADD(A_h, B_h, C_h, N);
+
```
SarbojitAMD: I think HipTest::freeArrays()  has been missed.

tests/src/runtimeApi/memory/hipMemcpyWithStream.cpp
```diff
@@ -0,0 +1,56 @@
...
+    unsigned blocks = HipTest::setNumBlocks(blocksPerCU, threadsPerBlock, N);
+    HipTest::initArrays(&A_d, &B_d, &C_d, &A_h, &B_h, &C_h, N, false);
+
+    hipStream_t stream;
+    HIPCHECK(hipStreamCreate(&stream));
+
+    HIPCHECK(hipMemcpyWithStream(A_d, A_h, Nbytes, hipMemcpyHostToDevice, stream));
+    HIPCHECK(hipMemcpyWithStream(B_d, B_h, Nbytes, hipMemcpyHostToDevice,stream));
+    hipLaunchKernelGGL(HipTest::vectorADD, dim3(blocks), dim3(threadsPerBlock), 0, stream,
+                        static_cast<const int*>(A_d), static_cast<const int*>(B_d), C_d, N);
```
SarbojitAMD: I feel it is always good to add hipGetLastError() after each kernel launch call to make sure it was clean like we do for all other hip API call, what do you say?

mangupta: Sure. Since we need to do this across all the tests, it is fine to take that up as a separate PR.

tests/src/runtimeApi/memory/hipMemcpyWithStream.cpp
mangupta: Agree. However let's do this clean as a separate PR. We need to get this in for branching.

PR #1724, opened 2019-12-07
Fix late-coming issues.

AlexVlx:
Joint work with @jeffdaily.


gargrahul:
@AlexVlx you forgot to add NVCC path related changes. Either add those or disable hipMemcpyWithStream test on NVCC for now.

AlexVlx:
Thank you @gargrahul, that is a very useful contribution indeed.

tests/src/runtimeApi/memory/hipMemcpyWithStream.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include "test_common.h"
+
+int main() {
+    size_t Nbytes = N * sizeof(int);
+    int numDevices = 0;
+    int *A_d, *B_d, *C_d;
+    int *A_h, *B_h, *C_h;
+
+    unsigned blocks = HipTest::setNumBlocks(blocksPerCU, threadsPerBlock, N);
+    HipTest::initArrays(&A_d, &B_d, &C_d, &A_h, &B_h, &C_h, N, false);
```
SarbojitAMD: The last argument of initArrays() is 'usePinnedHost' which is default set to 'false' hence it can be skipped here.

tests/src/runtimeApi/memory/hipMemcpyWithStream.cpp
```diff
@@ -0,0 +1,56 @@
...
+    HIPCHECK(hipStreamCreate(&stream));
+
+    HIPCHECK(hipMemcpyWithStream(A_d, A_h, Nbytes, hipMemcpyHostToDevice, stream));
+    HIPCHECK(hipMemcpyWithStream(B_d, B_h, Nbytes, hipMemcpyHostToDevice,stream));
+    hipLaunchKernelGGL(HipTest::vectorADD, dim3(blocks), dim3(threadsPerBlock), 0, stream,
+                        static_cast<const int*>(A_d), static_cast<const int*>(B_d), C_d, N);
+    HIPCHECK(hipStreamSynchronize(stream));
+    HIPCHECK(hipMemcpy(C_h, C_d, Nbytes, hipMemcpyDeviceToHost));
+    HipTest::checkVectorADD(A_h, B_h, C_h, N);
+
```
SarbojitAMD: Test has missed HipTest::freeArrays()

tests/src/runtimeApi/memory/hipMemcpyWithStream.cpp
jeffdaily: The current use is consistent with other tests.  wontfix.

tests/src/runtimeApi/memory/hipMemcpyWithStream.cpp
jeffdaily: This fixed by 6bf31e96051bfce1010b9f9727216058e10ca5d1 .

PR #1727, opened 2019-12-09
use launch_fence instead of barrier_bit

jeffdaily:


jeffdaily:
@AlexVlx or @gargrahul, since I don't have direct permissions to get these two commits into #1724, can someone merge them for me?  This PR is targeting the other PR branch.

PR #1728, opened 2019-12-10
[dtests] Added few Negative tests 

satyanveshd:
Added few of the negative scenarios as part of HIP dtests.

### DIFF
--- a/tests/src/Negative/memory/hipMemGetInfo.cpp
+++ b/tests/src/Negative/memory/hipMemGetInfo.cpp
@@ -1,0 +1,39 @@
+/*
+Copyright (c) 2015-Present Advanced Micro Devices, Inc. All rights reserved.
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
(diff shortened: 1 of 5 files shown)


tests/src/Negative/memory/hipMemGetInfo.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+/* HIT_START
+ * BUILD: %t %s ../../test_common.cpp EXCLUDE_HIP_PLATFORM nvcc
+ * TEST: %t
+ * HIT_END
+ */
+
+#include "test_common.h"
+#include <iostream>
+#include <hip/hip_runtime.h>
```
SarbojitAMD: test_common.h has both files included hence can be skipped

tests/src/Negative/memory/hipMemGetInfo.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+int main(){
+   hipError_t e;
+   size_t free,total;
+
+   e = hipMemGetInfo(NULL,NULL);
+  
+   HIPASSERT(e!=hipSuccess);
```
SarbojitAMD: I feel it is good to check exact error than just checking !hipSuccess, e.g. in this case check should be hipErrorInvalidValue

tests/src/Negative/memory/hipMemcpyToSymbolAsync.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+    HIPCHECK(hipMalloc(&Sd,100));
+
+    hipStream_t stream;
+    HIPCHECK(hipStreamCreate(&stream));
+
+    e = hipMemcpyToSymbolAsync(HIP_SYMBOL(Sd),S,100,0, hipMemcpyHostToDevice, stream);
+
+    HIPCHECK(hipFree(Sd));  
+    HIPASSERT(e!=hipSuccess);
```
SarbojitAMD: Better to check specific errors

tests/src/Negative/memory/hipMemcpyToSymbolAsync.cpp
```diff
@@ -0,0 +1,49 @@
...
+    char S[100]="This is not a device symbol";
+    
+    HIPCHECK(hipSetDevice(0));
+
+    HIPCHECK(hipMalloc(&Sd,100));
+
+    hipStream_t stream;
+    HIPCHECK(hipStreamCreate(&stream));
+
+    e = hipMemcpyToSymbolAsync(HIP_SYMBOL(Sd),S,100,0, hipMemcpyHostToDevice, stream);
```
SarbojitAMD: I think both error cases (hipErrorInvalidValue, hipErrorInvalidSymbol)  should be covered under -ve test cases

satyanveshd:
Raised #1735 after addressing the review comments. So, closing this.

PR #1731, opened 2019-12-10
[dtest] RCCL multi device launch test

gandryey:
Simple test for hipLaunchCooperativeKernelMultiDevice API.

### DIFF
--- a/tests/src/runtimeApi/module/hipLaunchCoopMultiKernel.cpp
+++ b/tests/src/runtimeApi/module/hipLaunchCoopMultiKernel.cpp
@@ -1,0 +1,211 @@
+/*
+Copyright (c) 2019 Advanced Micro Devices, Inc. All rights reserved.
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANNTY OF ANY KIND, EXPRESS OR
+IMPLIED, INNCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANNY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER INN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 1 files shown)


tests/src/runtimeApi/module/hipLaunchCoopMultiKernel.cpp
```diff
@@ -0,0 +1,212 @@
...
+  hipModule_t Module;
+  hipStream_t stream[MaxGPUs];
+
+  uint32_t* init = new uint32_t[BufferSizeInDwords];
+  for (uint32_t i = 0; i < BufferSizeInDwords; ++i) {
+    init[i] = i;
+  }
+
+  int nGpu = 0;
+  HIPCHECK(hipGetDeviceCount(&nGpu));
```
asalmanp: please add a check here to run the test only if nGpu>1 otherwise skip the test.

tests/src/runtimeApi/module/hipLaunchCoopMultiKernel.cpp
```diff
@@ -0,0 +1,212 @@
...
+  size_t copySizeInDwords = BufferSizeInDwords / nGpu;
+  hipDeviceProp_t deviceProp[MaxGPUs];
+
+  for (int i = 0; i < nGpu; i++) {
+    HIPCHECK(hipSetDevice(i));
+
+    // Calculate the device occupancy to know how many blocks can be run concurrently
+    hipGetDeviceProperties(&deviceProp[i], 0);
+    if (!deviceProp[i].cooperativeLaunch) {
```
asalmanp: here we should check for cooperativeMultiDeviceLaunch property to be supported for each device.

tests/src/runtimeApi/module/hipLaunchCoopMultiKernel.cpp
gandryey: I left it on purpose to run with one gpu as well, since 1 gpu case can be functional and the app won't need to create another code path. It was working fine in HIP-VDI.

tests/src/runtimeApi/module/hipLaunchCoopMultiKernel.cpp
```diff
@@ -0,0 +1,212 @@
...
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANNY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER INN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR INN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+// Simple test for hipLaunchCooperativeKernelMultiDevice API.
+
+/* HIT_START
+ * BUILD: %t %s ../../test_common.cpp EXCLUDE_HIP_PLATFORM all
```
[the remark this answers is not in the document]
gandryey: It can be built manually and tested on local systems. If KFD enables the feature in the public ROCM releases, then it can be enabled by default.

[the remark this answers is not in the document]
asalmanp: @mangupta  Can you please help to merge this test to the master branch? We want to use this test internally to test the cooperative multi-device launch. This test will be enabled once the base driver enables the cooperative feature (please also note that the single device cooperative launch hip test is also disabled by default)

PR #1732, opened 2019-12-10
[dtest] Adding VectorTest

cjatin:
Adding tests for vector types

### DIFF
--- a/tests/src/deviceLib/hip_floatnTM.cpp
+++ b/tests/src/deviceLib/hip_floatnTM.cpp
@@ -161,8 +161,9 @@ void testType(int msize) {
 
     hipMemcpy(d_fa, fa, sizeof(T) * msize, hipMemcpyHostToDevice);
     hipMemcpy(d_fb, fb, sizeof(T) * msize, hipMemcpyHostToDevice);
+    auto kernel = testOperationsGPU<T, D>;
 
-    hipLaunchKernelGGL(testOperationsGPU<T, D>, 1, msize, 0, 0, d_fa, d_fb, msize);
+    hipLaunchKernelGGL(kernel, 1, msize, 0, 0, d_fa, d_fb, msize);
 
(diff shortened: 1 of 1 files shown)


tests/src/deviceLib/hip_floatnTM.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+template <typename T, typename M>
+inline constexpr int count() {
+    return sizeof(T) / sizeof(M);
+}
+
+template <typename T, typename M>
+__device__ constexpr int countGPU() {
+    return sizeof(T) / sizeof(M);
+}
```
mangupta: Do we need both? We can just combine the two by using ```__host__ __device__``` keyword.

tests/src/deviceLib/hip_floatnTM.cpp
```diff
@@ -0,0 +1,258 @@
...
+        if constexpr (count<T, B>() >= 3) t.z = getRandomFloat();
+        if constexpr (count<T, B>() >= 4) t.w = getRandomFloat();
+
+        a[i] = t;
+    }
+}
+
+// Test operations
+template <typename T, typename B>
+void testOperations(T& a, T& b) {
```
mangupta: If we use a ```__host__ __device__``` here, we can simply call this function from ```testOperationsGPU()``` after the ```T &b = d_b[id];``` step.

tests/src/deviceLib/hip_floatnTM.cpp
```diff
@@ -0,0 +1,258 @@
...
+    }
+    if constexpr (countGPU<T, B>() >= 3) {
+        if (a.x > 0) b.x /= a.x;
+        a.x *= b.z;
+        a.y--;
+    }
+    if constexpr (countGPU<T, B>() >= 4) {
+        b.w = a.x;
+        a.w += (-b.y);
+    }
```
mangupta: All of this can be replaced by call to ```testOperations()``` if it has the ```__device__``` keyword added.

PR #1733, opened 2019-12-10
Fix the return type of demangle function so that its compatible across ABIs

saleelk:


include/hip/hcc_detail/hiprtc.h
```diff
@@ -102,8 +102,9 @@ namespace hip_impl
         {
             if (!result) return HIPRTC_ERROR_INVALID_INPUT;
 
-            *result = hip_impl::demangle(typeid(T).name());
-
+            char * res= hip_impl::demangle(typeid(T).name());
+            result->assign(res == nullptr ? std::string() : res);
+            std::free(res);
```
mhbliao: by have different types in the select ternary ops, it in facts converts 'res' into std::string first and assign that instead of char *. How about, replace std::string() with "".

scchan: `free` would segfault if `res==nullptr`.

src/hiprtc.cpp
```diff
@@ -143,7 +143,9 @@ struct _hiprtcProgram {
     {
         using namespace std;
 
-        name = hip_impl::demangle(name.c_str());
+        char* demangled = hip_impl::demangle(name.c_str());
+        name.assign(demangled == nullptr ? std::string() : demangled);
+        std::free(demangled);
```
scchan: `free` would segfault if `demangled==nullptr`.

saleelk: As opposed the the our previous assumptions, free(nullptr) is a no-op. So a ifcheck is redundant

include/hip/hcc_detail/hiprtc.h
saleelk: Sure i will change that to "". I overlooked the ternary operator trap

src/hiprtc.cpp
scchan: just looked up the behaviour for std::free and yes indeed

include/hip/hcc_detail/hiprtc.h
emankov: > free would segfault if res==nullptr.

Nope, all will be ok.

PR #1735, opened 2019-12-12
[dtests] Added few Negative tests

satyanveshd:
Added few of the negative scenarios as part of HIP dtests.


satyanveshd:
This is updated version of #1728.  Hence closing it.

PR #1739, opened 2019-12-13
[HIP] Unify hipError_t (Step 1)

emankov:
Step 1. Set the same values for RT error codes as for analogous Driver's

[Reason] RT's and Driver's error codes unification in CUDA 10.2


emankov:
Ping

PR #1744, opened 2019-12-16
Fix the return type of demangle function so that its compatible across ABIs

saleelk:
Sorry, my branch had some issues. This is the same as 1733

### DIFF
--- a/include/hip/hcc_detail/hiprtc.h
+++ b/include/hip/hcc_detail/hiprtc.h
@@ -84,7 +84,7 @@ hiprtcResult hiprtcGetCodeSize(hiprtcProgram prog, std::size_t* codeSizeRet);
 
 namespace hip_impl
 {
-    std::string demangle(const char* mangled_expression);
+    char* demangle(const char* mangled_expression);
 }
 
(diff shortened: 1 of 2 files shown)


include/hip/hcc_detail/hiprtc.h
```diff
@@ -84,7 +84,7 @@ hiprtcResult hiprtcGetCodeSize(hiprtcProgram prog, std::size_t* codeSizeRet);
 
 namespace hip_impl
 {
-    std::string demangle(const char* mangled_expression);
+    char* demangle(const char* mangled_expression);
 }
```
emankov: Maybe to change the type of `mangled_expression` to `const std::string&` instead?

saleelk: I believe I had tried that and it wasnt happy. Just added you to the local thread.

src/hiprtc.cpp
```diff
@@ -143,7 +143,9 @@ struct _hiprtcProgram {
     {
         using namespace std;
```
emankov: Please remove it or remove all `std::` namespace prefixes.

PR #1747, opened 2019-12-17
fix for hipcomplex operators failure on g++

satyanveshd:
Fixes the failure of few of the operators like unary minus(-) on g++, which resulted in hipSparse compilation failure, as well (SWDEV-216415).


gargrahul:
@satyanveshd Could you add that simple hipComplex test as a g++ dtest?

AlexVlx:
This does the wrong thing, it is going to break exactly what we were trying to fix (ABI compatibility with GCC).

mangupta:
Closing this since we have made the change in hipsparse.

PR #1748, opened 2019-12-17
[hip] Add macro guarding the enum conversion for scalar accessor.

mhbliao:
- That's a high overhead part, which needs enabling ONLY if necessary.

So far, rocFFT cannot be built due to that overhead. Previous 1m compilation time is increased up to 6+ minutes. For other sources, that overheads are too large to be measurable.


mhbliao:
Any chance to merge this change?

PR #1749, opened 2019-12-18
Remove filesystem dependency in hipRTC

cjatin:
Removing dependency on filesystem, so libstdc++fs is no longer required to link

### DIFF
--- a/src/hiprtc.cpp
+++ b/src/hiprtc.cpp
@@ -307,6 +307,17 @@ struct _hiprtcProgram {
         return true;
     }
 
+    void replaceExtension(std::string& fileName, std::string ext) const {
+        auto res = std::find(fileName.begin(), fileName.end(), '.');
+        if(res == std::end(fileName)) {
+            fileName += ext;
+        } else {
+            std::string s(fileName.begin(), res);
+            s += ext;
+            fileName = s;
+        }
+    }
+
     // ACCESSORS
     std::string writeTemporaryFiles(
         const std::string& programFolder) const
@@ -323,7 +334,7 @@ struct _hiprtcProgram {
         });
 
         auto tmp{(programFolder + '/' + name)};
-        tmp += ".cpp"; // Just add extension for now.
+        replaceExtension(tmp, ".cpp");
         ofstream{tmp}.write(source.data(), source.size());
 
         return tmp;
--- a/src/hiprtc.cpp
+++ b/src/hiprtc.cpp
@@ -35,7 +35,6 @@ THE SOFTWARE.
 #include <cassert>
 #include <cstdio>
 #include <cstdlib>
-#include <experimental/filesystem>
 #include <fstream>
 #include <future>
 #include <iterator>
@@ -43,11 +42,18 @@ THE SOFTWARE.
 #include <mutex>
 #include <stdexcept>
 #include <string>
+#include <sstream>
 #include <unordered_map>
 #include <utility>
 #include <vector>
 
 #include <iostream>
+
+#include <sys/stat.h>
+
+#if defined(_WIN32)
+#include <direct.h>
+#endif
 
 const char* hiprtcGetErrorString(hiprtcResult x)
 {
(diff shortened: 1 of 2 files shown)


src/hiprtc.cpp
```diff
@@ -80,6 +86,39 @@ const char* hiprtcGetErrorString(hiprtcResult x)
...
+    int ret = _mkdir(path.c_str());
+#else
+    mode_t mode = 0755;
+    int ret = mkdir(path.c_str(), mode);
+#endif
+    if (ret == 0) return true;
+    return false;
+}
+inline bool fileExists (const std::string& name) {
+  struct stat buffer;
```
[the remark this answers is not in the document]
gargrahul: Current implementation on github will not be used for Windows. HIP/VDI doesn't use the same code.

src/hiprtc.cpp
```diff
@@ -354,27 +387,25 @@ namespace
...
 
         Unique_temporary_path(const Unique_temporary_path&) = default;
         Unique_temporary_path(Unique_temporary_path&&) = default;
 
         ~Unique_temporary_path() noexcept
         {
-            std::experimental::filesystem::remove_all(path_);
+            std::string s("rm -r " + path_);
+            if (path_.substr(0, 5) == "/tmp/") {
+                system(s.c_str());
```
scchan: Is there a better way to remove the tmp directory other than spawn a new process?

src/hiprtc.cpp
```diff
@@ -80,6 +81,27 @@ const char* hiprtcGetErrorString(hiprtcResult x)
...
+    return (info.st_mode & S_IFDIR) != 0;
+}
+
+bool create_directory(const std::string& path) {
+    mode_t mode = 0755;
+    int ret = mkdir(path.c_str(), mode);
+    if (ret == 0) return true;
+    return false;
+}
+inline bool fileExists (const std::string& name) {
```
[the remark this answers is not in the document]
cjatin: Removed it.

src/hiprtc.cpp
```diff
@@ -286,22 +307,34 @@ struct _hiprtcProgram {
         return true;
     }
 
+    void replaceExtension(std::string& fileName, std::string ext) const {
+        auto res = std::find(fileName.begin(), fileName.end(), '.');
```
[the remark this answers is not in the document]
cjatin: great catch. changed.

src/hiprtc.cpp
```diff
@@ -286,22 +300,33 @@ struct _hiprtcProgram {
         return true;
     }
 
+    void replaceExtension(std::string& fileName, const std::string &ext) const {
+        auto res = std::find(fileName.rbegin(), fileName.rend(), '.');
+        auto sloc = std::find(fileName.rbegin(), fileName.rend(), '/');
+        if(res == fileName.rend() || res > sloc) {
```
scchan: Could you replace these with `std::string::rfind` instead?  It would make the code a lot more readable, especially for the non-obvious usages of `reverse_iterators` in the `res > sloc` comparison and in `replace` below.

src/hiprtc.cpp
```diff
@@ -286,22 +300,33 @@ struct _hiprtcProgram {
...
+            fileName += ext;
+        } else {
+            fileName.replace(res.base(), fileName.end(), ext);
```
scchan: to be updated when the above are replaced with `rfind`

src/hiprtc.cpp
cjatin: Done

src/hiprtc.cpp
```diff
@@ -301,12 +301,12 @@ struct _hiprtcProgram {
     }
 
     void replaceExtension(std::string& fileName, const std::string &ext) const {
-        auto res = std::find(fileName.rbegin(), fileName.rend(), '.');
-        auto sloc = std::find(fileName.rbegin(), fileName.rend(), '/');
-        if(res == fileName.rend() || res > sloc) {
-            fileName += ext;
+        auto res = fileName.rfind('.');
+        auto sloc = fileName.rfind('/');
+        if (res != std::string::npos && res > sloc) {
```
scchan: This needs to catch the case where `/` is not found, the condition should be
`if (res != npos && (res > sloc || sloc == npos))`

cjatin: if `/` is not found, `sloc `should have the maximum possible value for `size_t `and the condition will fail.

scchan: if '.' is found but '/' is not found, don't you want to go to this path as well?

cjatin: fixed it.

PR #1751, opened 2019-12-18
Reduce GPU copying based on arch it runs on

ansurya:
SWDEV-213230-:
[Perf] Reduce the amount of copying for GPU objects in the HIP runtime.


src/hip_hcc.cpp
```diff
@@ -2514,6 +2515,14 @@ hipError_t hipHccGetAcceleratorView(hipStream_t stream, hc::accelerator_view** a
 // TODO - add a contect sequence number for debug. Print operator<< ctx:0.1 (device.ctx)
 
 namespace hip_impl {
+    std::unordered_set<std::string> get_all_gpuarch() {
+        std::unordered_set<std::string> r{};
```
cjatin: For inserting and searching on a small set of data, set should give you a slight better performance against unordered_set

http://quick-bench.com/LZsAxAb4LGq5AjOSZeiQFun4oz8

scchan: Everytime `get_all_gpuarch` is called, it constructs a new `set`.  One option would be to have a static `set` in `get_all_gpuarch`, which would get initialized once and make the function to return its reference instead.

lpl_ca/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ target_include_directories(ca PUBLIC ${PROJECT_SOURCE_DIR}/src)
 find_library(
     hsart NAMES libhsa-runtime64.so libhsa-runtime64.so.1 HINTS ${HSA_PATH}/lib)
 target_link_libraries(ca PUBLIC ${hsart})
-target_compile_options(ca PUBLIC -Wall)
+target_compile_options(ca PUBLIC -DDISABLE_REDUCED_GPU_BLOB_COPY -Wall)
```
scchan: Should we remove this? Otherwise this patch is disabled and will not get tested.

include/hip/hcc_detail/code_object_bundle.hpp
```diff
@@ -126,11 +125,16 @@ class Bundled_code_header {
                 it += sizeof(y.header.cbuf);
 
                 y.triple.assign(it, it + y.header.triple_sz);
-
+                #ifdef DISABLE_REDUCED_GPU_BLOB_COPY
                 std::copy_n(f + y.header.offset, y.header.bundle_sz, std::back_inserter(y.blob));
-
+                #else
+                auto gpuArch = get_all_gpuarch();
+                auto itgpuArch =std::find(gpuArch.begin(),gpuArch.end(),y.triple);
```
scchan: code style - missing space after `=`

lpl_ca/CMakeLists.txt
ansurya: This is disabled for lpl_ca . It is enabled for HIP runtime.

CMakeLists.txt
```diff
@@ -301,6 +301,7 @@ if(HIP_PLATFORM STREQUAL "hcc")
         target_link_libraries(hip_hcc_static PRIVATE hc_am)
 
         add_library(hiprtc SHARED src/hiprtc.cpp src/code_object_bundle.cpp)
+        target_compile_options(hiprtc PRIVATE -DDISABLE_REDUCED_GPU_BLOB_COPY)
```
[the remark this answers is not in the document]
ansurya: It is disabled for hiprtc. It is enabled for hip runtime.

src/hip_hcc.cpp
ansurya: done.

include/hip/hcc_detail/code_object_bundle.hpp
ansurya: done

include/hip/hcc_detail/code_object_bundle.hpp
```diff
@@ -126,11 +125,16 @@ class Bundled_code_header {
                 it += sizeof(y.header.cbuf);
 
                 y.triple.assign(it, it + y.header.triple_sz);
-
+                #ifdef DISABLE_REDUCED_GPU_BLOB_COPY
                 std::copy_n(f + y.header.offset, y.header.bundle_sz, std::back_inserter(y.blob));
-
+                #else
+                auto gpuArch = get_all_gpuarch();
```
scchan: should be `auto&` to avoid a new copy

lpl_ca/CMakeLists.txt
scchan: reason?

src/hip_hcc.cpp
AlexVlx: > For inserting and searching on a small set of data, set should give you a slight better performance against unordered_set
> 
> http://quick-bench.com/LZsAxAb4LGq5AjOSZeiQFun4oz8

That suggested benchmark is not showing what you think it does. Associative data structures are interesting because they allow more efficient searching than a linear search on their underlying linked data structure. Furthermore, there are no conflicts, and conflict resolution is an important aspect here. Finally, a vector is almost always going to be faster.

CMakeLists.txt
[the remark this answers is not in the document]
ansurya: hiprtc generates code_object_bundle at runtime for the GPU it runs on. Reduce GPU copy is not applicable here.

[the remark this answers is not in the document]
ansurya: hiprtc uses code_object_bundle to compile code at run time. Reduced GPU copy is not applicable.

lpl_ca/CMakeLists.txt
ansurya: lpl_ca is used to generate HSA code objects. Reduced GPU copy is not applicable. and get_all_gpuarch() defined in hip_hcc.cpp is not visible for lpl_ca and hiprtc.

scchan: >lpl_ca is used to generate HSA code objects. Reduced GPU copy is not applicable.
This is speculative.  Note that the interface is a vector of targets https://github.com/ROCm-Developer-Tools/HIP/blob/master/lpl_ca/lpl.hpp#L67
 
>get_all_gpuarch() defined in hip_hcc.cpp is not visible for lpl_ca and hiprtc.
can we fix that?

mangupta: lpl/ca generate hsaco fat binaries that are stored to disk for use in ```hipModuleLoad()``` on the actual target system. So lpl/ca should not drop GPU targets based on hardware on the build system. Hence ```get_all_gpuarch()``` is not applicable to lpl/ca. Also ```hipModuleLoad()``` would drop the GPU targets when we load the hsaco fat binary in the target system.

PR #1766, opened 2019-12-31
Detect when an explicit printf buffer flush is required

scchan:
Detect when an explicit printf buffer flush is required in a device/stream synchronization function.

### DIFF
--- a/src/hip_hcc.cpp
+++ b/src/hip_hcc.cpp
@@ -304,18 +304,28 @@ void ihipStream_t::wait(LockedAccessor_StreamCrit_t& crit) {
 
 //---
 // Wait for all kernel and data copy commands in this stream to complete.
-void ihipStream_t::locked_wait() {
+inline void ihipStream_t::locked_wait(bool& waited) {
     // create a marker while holding stream lock,
     // but release lock prior to waiting on the marker
     hc::completion_future marker;
     {
         LockedAccessor_StreamCrit_t crit(_criticalData);
         // skipping marker since stream is empty
-        if (crit->_av.get_is_empty()) return;
+        if (crit->_av.get_is_empty()) {
+            waited = false;
+            return;
+        }
         marker = crit->_av.create_marker(hc::no_scope);
     }
 
     marker.wait(waitMode());
+    waited = true;
+    return;
+};
+
+void ihipStream_t::locked_wait() {
+    bool waited;
+    locked_wait(waited);
 };
 
 // Causes current stream to wait for specified event to complete:
@@ -1040,6 +1050,7 @@ void ihipCtx_t::locked_syncDefaultStream(bool waitOnSelf, bool syncHost) {
     // Vector of ops sent to each stream that will complete before ops sent to null stream:
     std::vector<hc::completion_future> depOps;
 
+    bool last_stream_waited = false;
     for (auto streamI = crit->const_streams().begin(); streamI != crit->const_streams().end();
          streamI++) {
         ihipStream_t* stream = *streamI;
@@ -1051,7 +1062,8 @@ void ihipCtx_t::locked_syncDefaultStream(bool waitOnSelf, bool syncHost) {
 
         if (HIP_SYNC_NULL_STREAM) {
             if (waitThisStream) {
-                stream->locked_wait();
+                last_stream_waited = false;
+                stream->locked_wait(last_stream_waited);
             }
         } else {
             if (waitThisStream) {
@@ -1082,6 +1094,14 @@ void ihipCtx_t::locked_syncDefaultStream(bool waitOnSelf, bool syncHost) {
             defaultCf.wait();  // TODO - account for active or blocking here.
         }
     }
+    else if ( (HIP_SYNC_NULL_STREAM && !last_stream_waited) ||
+              (!HIP_SYNC_NULL_STREAM && depOps.empty()) ) {
+        // This catches all the conditions where the printf buffer
+        // need to be explicitly flushed
+        if (syncHost) {
+            Kalmar::getContext()->flushPrintfBuffer();
+        }
+    }
 
     tprintf(DB_SYNC, "  syncDefaultStream depOps=%zu\n", depOps.size());
 }
@@ -1101,9 +1121,18 @@ void ihipCtx_t::locked_waitAllStreams() {
     LockedAccessor_CtxCrit_t crit(_criticalData);
 
     tprintf(DB_SYNC, "waitAllStream\n");
+    bool need_printf_flush = false;
     for (auto streamI = crit->const_streams().begin(); streamI != crit->const_streams().end();
          streamI++) {
-        (*streamI)->locked_wait();
+        bool waited;
+        (*streamI)->locked_wait(waited);
+        need_printf_flush = !waited;
+    }
+
+    // When synchronizing with the last stream, if we didn't do an explicit wait,
+    // then we need to an extra flush to the printf buffer
+    if (need_printf_flush) {
+        Kalmar::getContext()->flushPrintfBuffer();
     }
 }
 
@@ -1492,7 +1521,11 @@ hipError_t ihipStreamSynchronize(TlsData *tls, hipStream_t stream) {
         ctx->locked_syncDefaultStream(true /*waitOnSelf*/, true /*syncToHost*/);
     } else {
         // note this does not synchornize with the NULL stream:
-        stream->locked_wait();
+        bool waited;
+        stream->locked_wait(waited);
+        if (!waited) {
+            Kalmar::getContext()->flushPrintfBuffer();
+        }
         e = hipSuccess;
     }
 
--- a/src/hip_hcc_internal.h
+++ b/src/hip_hcc_internal.h
@@ -570,7 +570,7 @@ class ihipStream_t {
     LockedAccessor_StreamCrit_t lockopen_preKernelCommand();
     void lockclose_postKernelCommand(const char* kernelName, hc::accelerator_view* av, bool unlockNotNeeded = 0);
 
-
+    void locked_wait(bool& waited);
     void locked_wait();
 
     hc::accelerator_view* locked_getAv() {
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -28,6 +28,7 @@ THE SOFTWARE.
 #include "hip/hip_ext.h"
 #include "program_state.inl"
 #include "trace_helper.h"
+#include "hc_am.hpp"
 
 #include <hsa/amd_hsa_kernel_code.h>
 #include <hsa/hsa.h>


scchan:
@jeffdaily Would you please review this PR?  Thanks

jeffdaily:
I vote to either (1) revert #1667, or (2) simply to add the `flushPrintfBuffer` to the `ihipStream_t::locked_wait`  in the cases where the marker is skipped.  The latter option effectively restores the previous behavior while also preserving the change to not add the marker.

@scchan I understand why you are only flushing the printf buffer if the last stream does not `wait`.  My suggested change (2) above would call the printf flush (unnecessarily) more often, but it would simplify the code a lot.

Aside: are there any race conditions with the printf buffer?  Each HIP stream will lock its mutex, but the printf buffer is an implementation detail of the HCC runtime and I did not see any mutex used there in case multiple streams via multiple threads are flushing the printf buffer simultaneously.

scchan:
> Thanks Siu Chi, for taking over.
> 
> This change looks overly complicated to me. Why not just call flushPrintfBuffer() wherever CUDA says that a printf becomes visible? There is no reason to tie this to a wait on empty streams. Or given the end-of-life status of HCC, what if we just restore the final flush that is currently commented out? It's not perfect but it's guaranteed to work.

The issue is that if a stream is empty, then the runtime would avoid waiting on it but with a side effect of bypassing the printf flush, which is still needed even if the stream is empty.  This patch is to detect that situation and then flush the printf buffer explicitly.  
The final flush during shutdown is a final catch-all fallback but the printf behaviour would still be unpredictable since it's not being flushed at explicit synchronization points.

scchan:
> I vote to either (1) revert #1667, or (2) simply to add the `flushPrintfBuffer` to the `ihipStream_t::locked_wait` in the cases where the marker is skipped. The latter option effectively restores the previous behavior while also preserving the change to not add the marker.
> 
> @scchan I understand why you are only flushing the printf buffer if the last stream does not `wait`. My suggested change (2) above would call the printf flush (unnecessarily) more often, but it would simplify the code a lot.
> 

Do we observe significant gain with PR#1667?
The complexity in this patch is try to minimize the number of flushes since probably the flush that happens after the last stream the runtime tries to synchronize with really matters.

> Aside: are there any race conditions with the printf buffer? Each HIP stream will lock its mutex, but the printf buffer is an implementation detail of the HCC runtime and I did not see any mutex used there in case multiple streams via multiple threads are flushing the printf buffer simultaneously.

That looks like to be a problem, I'll address that in a separate PR

jeffdaily:
Unfortunately, this change was made as part of a set of related changes that benefit certain PyTorch models.  The various changes were not all tested in isolation, so I cannot point to any specific benchmark(s) that measure this PR #1667 directly.

I do recall one case where PyTorch was calling `hipMemcpyAysnc` followed by `hipStreamSynchronize`, mixed with various tiny kernels.  In general, we looked for places where we could reduce any host overhead.  We were trying to optimize roundtrips like the one in #1667 , if we could otherwise determine that the stream was empty.

jeffdaily:
Even though I can't point to a benchmark result, I'm still in favor of keeping #1667 and fixing it with this PR.  Although this PR is somewhat complex, it does look like it will fix the broken printf and do so without calling the flush too often.

ssahasra:
I agree with @jeffdaily about keeping #1667. My only question is why not always flush the buffer ... is it a significant cost? Also @scchan, do you intend to introduce a new mutex in HCC to remove race conditions on the printf buffer? That would be a pre-requisite for this patch, right?

jeffdaily:
@ssahasra No, a new printf mutex in HCC is not a prerequisite for this patch.  Whether we trigger flushes directly from HIP or from the HIP->HCC->`accelerator_view::wait` path, we are currently no worse off than before since IIRC the HCC printf buffer has never had a mutex.  Just something I noticed when reviewing this PR.

scchan:
As discussed, adding a mutex for the printf buffer: https://github.com/RadeonOpenCompute/hcc/pull/1357

ssahasra:
@scchan, what's your call on the making the flush unconditional? I think it makes the code simpler, but I am not opposed to the current patch.

scchan:
@ssahasra I'm inclined to keep the conditional flush in this patch because I think it may benefit some workloads that create numerous streams but that are idle most of the time (e.g stream pools).

src/hip_hcc.cpp
```diff
@@ -1051,7 +1062,8 @@ void ihipCtx_t::locked_syncDefaultStream(bool waitOnSelf, bool syncHost) {
 
         if (HIP_SYNC_NULL_STREAM) {
             if (waitThisStream) {
```
srinivamd: Should `waitThisStream` be the first order check instead of being nested inside?

src/hip_hcc.cpp
```diff
@@ -1082,6 +1094,14 @@ void ihipCtx_t::locked_syncDefaultStream(bool waitOnSelf, bool syncHost) {
             defaultCf.wait();  // TODO - account for active or blocking here.
         }
     }
+    else if ( (HIP_SYNC_NULL_STREAM && !last_stream_waited) ||
+              (!HIP_SYNC_NULL_STREAM && depOps.empty()) ) {
+        // This catches all the conditions where the printf buffer
+        // need to be explicitly flushed
+        if (syncHost) {
```
srinivamd: Should `syncHost` check be the first order check instead of nested inside? Could make the code simpler to read.

src/hip_hcc.cpp
```diff
@@ -304,18 +304,28 @@ void ihipStream_t::wait(LockedAccessor_StreamCrit_t& crit) {
 
 //---
 // Wait for all kernel and data copy commands in this stream to complete.
-void ihipStream_t::locked_wait() {
+inline void ihipStream_t::locked_wait(bool& waited) {
```
srinivamd: Would defining a new function such as `locked_waited()` to return `bool` break compatibility, for example,
`bool ihipStream_t::locked_waited() {  // returns true when waited on a non-empty stream`
could be used instead of overloading `locked_wait()` and avoid defining temporary `bool`s to get result inside `for` loops etc. 
Was there a reason to overload `locked_wait()` instead of defining a new function?

src/hip_hcc.cpp
```diff
@@ -304,18 +304,28 @@ void ihipStream_t::wait(LockedAccessor_StreamCrit_t& crit) {
...
     }
 
     marker.wait(waitMode());
+    waited = true;
+    return;
+};
+
+void ihipStream_t::locked_wait() {
+    bool waited;
+    locked_wait(waited);
```
srinivamd: Nit: It would be good to avoid this by defining a new function that returns `bool` such as `locked_waited()` and ignore the return value instead.

src/hip_hcc.cpp
```diff
@@ -1101,9 +1121,18 @@ void ihipCtx_t::locked_waitAllStreams() {
     LockedAccessor_CtxCrit_t crit(_criticalData);
 
     tprintf(DB_SYNC, "waitAllStream\n");
+    bool need_printf_flush = false;
     for (auto streamI = crit->const_streams().begin(); streamI != crit->const_streams().end();
          streamI++) {
-        (*streamI)->locked_wait();
+        bool waited;
+        (*streamI)->locked_wait(waited);
+        need_printf_flush = !waited;
```
srinivamd: With a `bool locked_waited()` function, the for loop could be rewritten as:
`for (auto streamI = crit->const_streams().begin(); `
`      streamI != crit->const_streams().end();`
`      ++streamI) { `
`            need_printf_flush |= (*streamI)->locked_waited()`
` }`

src/hip_hcc_internal.h
```diff
@@ -570,7 +570,7 @@ class ihipStream_t {
     LockedAccessor_StreamCrit_t lockopen_preKernelCommand();
     void lockclose_postKernelCommand(const char* kernelName, hc::accelerator_view* av, bool unlockNotNeeded = 0);
 
-
+    void locked_wait(bool& waited);
```
srinivamd: A new function that returns `bool` for waited condition would be better suited.

src/hip_hcc.cpp
jeffdaily: I don't disagree that the first order check might need revision, but not in this PR.

src/hip_hcc.cpp
jeffdaily: Same comment as above.  Please open a new PR to address this.

src/hip_hcc.cpp
jeffdaily: Using `|=` isn't correct here.  We are only concerned with whether the last stream actually performed the wait+flush.  Using `|=` would result in more print flushes than just checking the last stream.

srinivamd: @jeffdaily Thanks for pointing that out. Instead of `|=` the original code's `!` of `(*streamI)->locked_waited()` should work then, indicating whether the last (in the iteration order) non-empty stream had a wait marker, right?
Does the iteration order from `begin()` to `end()` correspond to oldest to newest time order? IOW, would iterating through the streams in reverse order matter?

src/hip_hcc.cpp
scchan: Keeping `syncHost` nested here is preferred since it follows the same structure as in the previous `if` branch

src/hip_hcc.cpp
scchan: We are only interested in whether the last iteration did a flush through `locked_waited`.  `|=` doesn't provide that information.  The traversal order through the list of stream doesn't matter here.

src/hip_hcc.cpp
scchan: it's to avoid breaking compatibility and you can't overload by the return type

srinivamd: @scchan Not sure extending the interface by adding a new function and using it in the new/modified logic would break compatibility - compatibility with what?

scchan: I'm not sure if I understand your question correctly.  What I meant was that having both `void locked_waited()` and `bool locked_waited()` doesn't work

srinivamd: @scchan The suggestion was to create a new function `bool locked_wait**ed**()` (not `wait` but `waited` as in waited?), not overloading. (Sorry the names are similar, so easy to have missed that.)

scchan: sorry I missed the name difference but I still prefer having the overloaded version to show that both are semantically identical and that one gives you the status.

srinivamd: @scchan Ok, I understand your preference.
I think the use of overload here, and the resulting code constructs of requiring the use of a temporary variable instead of checking the return value from a call makes the code complicated/less readable. Good to know that there is no compatibility issue preventing the addition of a new function should that approach be considered later.

src/hip_hcc.cpp
srinivamd: @scchan While consistency is a good thing, in this case, it seems might be more efficient to test `syncHost` or `withThisStream` first. and could also make the code more readable.

PR #1767, opened 2020-01-02
Fix hipcc warning related to hipVersion

gargrahul:
Fixes following warnings:-
Name "main::HIP_VERSION_MAJOR" used only once: possible typo at /bin/hipcc line 122.
Name "main::HIP_VERSION_MINOR" used only once: possible typo at /bin/hipcc line 122.


bin/hipcc
```diff
@@ -119,7 +119,7 @@ sub delete_temp_dirs {
 #HIP_PLATFORM controls whether to use NVCC or HCC for compilation:
 $HIP_PLATFORM= `$HIP_PATH/bin/hipconfig --platform` // "hcc";
 $HIP_VERSION= `$HIP_PATH/bin/hipconfig --version`;
-($HIP_VERSION_MAJOR, $HIP_VERSION_MINOR, $HIP_VERSION_PATCH) = split(/\./, $HIP_VERSION);
+$HIP_VERSION_PATCH = (split(/\./, $HIP_VERSION))[-1];
 my ($HIP_VERSION_GITDATE, $HIP_VERSION_GITHASH) = split(/-/, $HIP_VERSION_PATCH);
```
mangupta: We don't need any of these lines anymore. We used to split HIP_VERSION into MAJOR, MINOR, PATCH, GITDATE and GITHASH only for setting the preprocessor macros. Since we don't do that anymore, we can remove these lines entirely.

CMakeLists.txt
```diff
@@ -354,7 +354,7 @@ set(_versionInfoHeader
 #define HIP_VERSION       (HIP_VERSION_MAJOR * 100 + HIP_VERSION_MINOR)\n
 #endif\n
 ")
-file(WRITE "${CMAKE_CURRENT_SOURCE_DIR}/include/hip/hipVersion.h" ${_versionInfoHeader})
+file(WRITE "${CMAKE_CURRENT_SOURCE_DIR}/include/hip/hip_version.h" ${_versionInfoHeader})
```
SarbojitAMD: There is a suggestion of not generating the file under the source folder but to generate it under the build folder so that git will not track it as new addition to the repro.

mangupta: The way to handle this is to add it to .gitignore similar to what we have done for include/hip/hcc_detail/hip_prof_str.h (which is also a generated file)

PR #1771, opened 2020-01-03
Changes for multiple ROCm installation

paulfreddy:


packaging/hip-targets-release.cmake
```diff
@@ -5,41 +5,48 @@
 # Commands may need to know the format version.
 set(CMAKE_IMPORT_FILE_VERSION 1)
 
+# Set ROCM PATH from envinorment and default to /opt/rocm if not provided.
+if( DEFINED ENV{ROCM_PATH} )
```
pfultz2: An environment variable should not be used here. Users should not required to set an environment variable to use `find_package(hip)`.  The path should be computed based on the relative path of the file. Usually, its computed like this:

```cmake
# Compute the installation prefix relative to this file.
get_filename_component(_IMPORT_PREFIX "${CMAKE_CURRENT_LIST_FILE}" REALPATH)
get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
```

You can adjust the number of `get_filename_component` calls based on the relative location of the file to its installation directory.

packaging/hip-targets.cmake
```diff
@@ -39,22 +39,25 @@ unset(_expectedTargets)
 
 
 # The installation prefix configured by this project.
-set(_IMPORT_PREFIX "/opt/rocm/hip")
-
+if( DEFINED ENV{ROCM_PATH} )
```
pfultz2: An environment variable should not be used here either.

packaging/hip-targets.cmake
```diff
@@ -39,22 +39,25 @@ unset(_expectedTargets)
...
+else()
+     set(_IMPORT_PREFIX "/opt/rocm/hip")
+endif()
 # Create imported target hip::hip_hcc_static
 add_library(hip::hip_hcc_static STATIC IMPORTED)
 
 set_target_properties(hip::hip_hcc_static PROPERTIES
-  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;/opt/rocm/hsa/include"
-  INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;/opt/rocm/hsa/include"
+  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;$ENV{ROCM_PATH}/hsa/include"
```
pfultz2: The hip-config.cmake should search for hsa and use the path found by cmake. You can see how hcc does it [here](https://github.com/RadeonOpenCompute/hcc/blob/clang_tot_upgrade/lib/hcc-config.cmake.in#L7).

packaging/hip-targets-release.cmake
paulfreddy: hipcc itself use ROCM_PATH environment variable to pick the path, hence we thought it would be OK. 
I believe ROCM_PATH  in hipcc is provided for a reason, 

I guess we have different options here.

1. If we really want to remove all references of ENV and we can do it as you mentioned above and use relative path. 

2. If ROCM_PATH in hipcc is given for a purpose I would assume same design principle here also.

3. We can give option to user to control the path by env and if anyone uses hipcc without ROCM_PATH in that case will fall back to relative path to find the ROCM path.

mangupta: @paulfreddy : bin/hipcc is generic file. It should be independent of HIP or ROCm version. So we need ROCM_PATH there. However this file is specific to a certain HIP build. However since our ROCm build system does not guarantee that HIP is relative to ROCm in the build system we cannot use Paul's solution either. So we need something similar to lines 74-79 in hipcc. So ROCM_PATH is to be derived as ENV{ROCM_PATH} if defined. Else relative to HIP_PATH as long as HIP is installed as part of ROCm. Else default to /opt/rocm.

packaging/hip-targets.cmake
mangupta: Same comments as packaging/hip-targets-release.cmake

packaging/hip-targets.cmake
mangupta: @pfultz2 The issue is that our build system (rocm-ci) does not have hsa and hcc installed in the same paths as the target system. So paths have to be determined based on target install locations.
@paulfreddy Again we need consistency in ROCM_PATH, HSA_PATH, HCC_HOME evaluation.

samples/2_Cookbook/12_cmake_hip_add_executable/CMakeLists.txt
```diff
@@ -1,5 +1,11 @@
 cmake_minimum_required(VERSION 2.8.3)
 
+if ($ENV{ROCM_PATH})
+    set(opt_rocm_install_path $ENV{ROCM_PATH})
+else()
+    set(opt_rocm_install_path "/opt/rocm")
+endif()
+
```
mangupta: Why is this needed? Who uses opt_rocm_install_path?

packaging/hip-targets-release.cmake
pfultz2: > However since our ROCm build system does not guarantee that HIP is relative to ROCm in the build system we cannot use Paul's solution either.

My solution does work. Just using relative paths wont work. That is why I said it needs to search for its dependencies like what hcc does does [here](https://github.com/RadeonOpenCompute/hcc/blob/clang_tot_upgrade/lib/hcc-config.cmake.in#L7). In fact, hip already does this for hcc, but doesn't do this for hsa runtime.

packaging/hip-targets.cmake
pfultz2: > The issue is that our build system (rocm-ci) does not have hsa and hcc installed in the same paths as the target system

Yes, exactly, and its not just rocm-ci that does this, package managers like spack or cget can install everything in different directories. 

This why it needs to call `find_*` commands to find the dependencies. Hip already does this for hcc by calling `find_dependency(hcc)` [here](https://github.com/ROCm-Developer-Tools/HIP/blob/master/hip-config-hcc.cmake.in#L53). However, it missing a find command for hsa runtime headers. 

Hip needs to call `find_path(HSA_HEADER hsa/hsa.h)` to get the headers or it could rely on hcc to already find hsa and then just use either the `HSA_HEADER` variable or use the `hsa-runtime64` imported target. 

Hcc does it properly and it doesnt require relying on custom env variables or using incorrect relative paths. Hcc also works with rocm-ci and other package managers like cget.

Hip should do this properly as well at least for cmake(hipcc is just a perl script that we are not compelled to use so its not urgent to fix).

packaging/hip-base.postinst
```diff
@@ -6,8 +6,7 @@ pushd () {
 popd () {
     command popd "$@" > /dev/null
 }
-
-ROCMDIR=/opt/rocm
+ROCMDIR=@ROCM_PATH@
```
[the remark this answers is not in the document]
paulfreddy: This works on cmake and is verified.

packaging/hip-base.prerm
```diff
@@ -8,7 +8,7 @@ popd () {
     command popd "$@" > /dev/null
 }
 
-ROCMDIR=/opt/rocm
+ROCMDIR=@ROCM_PATH@
```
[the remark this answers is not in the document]
paulfreddy: This works and verified as well.

packaging/hip-nvcc.txt
```diff
@@ -5,7 +5,7 @@ project(hip_nvcc)
 # Packaging steps
 #############################
 set(CPACK_SET_DESTDIR TRUE)
-set(CPACK_INSTALL_PREFIX "/opt/rocm/hip")
+set(CPACK_INSTALL_PREFIX @CPACK_INSTALL_PREFIX@ )
```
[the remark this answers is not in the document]
paulfreddy: Did it this was to keep the changes consistent and avoid hard coding of /opt/rocm

packaging/hip-hcc.postinst
```diff
@@ -7,7 +7,8 @@ popd () {
     command popd "$@" > /dev/null
 }
 
-ROCMDIR=/opt/rocm
+
+ROCMDIR=@ROCM_PATH@
```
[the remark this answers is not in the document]
paulfreddy: Works and verified

packaging/hip-hcc.prerm
```diff
@@ -8,7 +8,7 @@ popd () {
     command popd "$@" > /dev/null
 }
 
-ROCMDIR=/opt/rocm
+ROCMDIR=@ROCM_PATH@
```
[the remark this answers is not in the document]
paulfreddy: Works and verified.

cmake/FindHIP.cmake
```diff
@@ -555,7 +560,13 @@ macro(HIP_ADD_EXECUTABLE hip_target)
         list(REMOVE_ITEM _sources ${_source_files})
     endif()
     if("x${HCC_HOME}" STREQUAL "x")
-        set(HCC_HOME "/opt/rocm/hcc")
+        if (DEFINED $ENV{ROCM_PATH})
+            set(HCC_HOME "$ENV{ROCM_PATH}/hcc")
+        elseif( DEFINED $ENV{HIP_PATH})
+            set(HCC_HOME "$ENV{HIP_PATH}/../hcc")
```
mangupta: You need to test if HIP_PATH is part of ROCm installation. So you need to test if $HIP_PATH/../.info/version exists before using ENV(HIP_PATH)/../hcc.

packaging/hip-base.txt
```diff
@@ -15,7 +15,7 @@ install(DIRECTORY @hip_SOURCE_DIR@/cmake DESTINATION .)
 # Packaging steps
 #############################
 set(CPACK_SET_DESTDIR TRUE)
-set(CPACK_INSTALL_PREFIX "/opt/rocm/hip")
+set(CPACK_INSTALL_PREFIX @CPACK_INSTALL_PREFIX@ )
```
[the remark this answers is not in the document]
paulfreddy: Added default path in CMakeList.txt

samples/2_Cookbook/12_cmake_hip_add_executable/CMakeLists.txt
paulfreddy: Removed

packaging/hip-targets-release.cmake
```diff
@@ -5,41 +5,50 @@
...
+     set(ROCM_PATH $ENV{ROCM_PATH})
+elseif(DEFINED ENV{HIP_PATH})
+     set(ROCM_PATH "$ENV{HIP_PATH}/..")
```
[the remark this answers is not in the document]
paulfreddy: checked using hipconfig

packaging/hip-targets.cmake
```diff
@@ -39,22 +39,37 @@ unset(_expectedTargets)
...
+     set(_IMPORT_PREFIX "$ENV{HIP_PATH}")
+else()
+     set(_IMPORT_PREFIX "/opt/rocm/hip")
+endif()
 # Create imported target hip::hip_hcc_static
 add_library(hip::hip_hcc_static STATIC IMPORTED)
 
+find_path(HSA_HEADER hsa/hsa.h
+  PATHS
+    "${_IMPORT_PREFIX}/../include"
```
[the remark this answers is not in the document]
paulfreddy: done

cmake/FindHIP.cmake
paulfreddy: done with checking hipconfig

packaging/hip-targets-release.cmake
```diff
@@ -7,30 +7,40 @@ set(CMAKE_IMPORT_FILE_VERSION 1)
...
-else ()
-     set(ROCM_PATH "/opt/rocm")
+     set(ROCM_PATH "$ENV{ROCM_PATH}")
 endif()
 
+get_filename_component(_DIR "${CMAKE_CURRENT_LIST_FILE}" REALPATH)
+# Set ROCM PATH from envinorment and default to /opt/rocm if not provided.
+set (HIP_ROOT_DIR)
+find_path(
+          HIP_ROOT_DIR
```
pfultz2: Find should be used to find the hip's root path. This always be relative to the cmake file. Find is only needed for hip's dependencies(such as hsa runtime).

packaging/hip-targets.cmake
```diff
@@ -38,38 +38,54 @@ unset(_targetsNotDefined)
...
+          DOC "HIP installed location"
+          )
+
+if (HIP_ROOT_DIR-NOTFOUND)
+  message (FATAL_ERROR "HIP not found! ROCM_PATH/HIP_PATH environment not set")
 endif()
 # Create imported target hip::hip_hcc_static
 add_library(hip::hip_hcc_static STATIC IMPORTED)
 
 find_path(HSA_HEADER hsa/hsa.h
```
pfultz2: Ideally, this should be moved to hip-config.cmake file.

packaging/hip-targets.cmake
```diff
@@ -38,38 +38,54 @@ unset(_targetsNotDefined)
...
 
-if (NOT HSA_FOUND)
+if (HSA_HEADER-NOTFOUND)
   message (FATAL_ERROR "HSA header not found! ROCM_PATH environment not set")
 endif()
 
 set_target_properties(hip::hip_hcc_static PROPERTIES
-  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;${HSA_HEADER}"
-  INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;${HSA_HEADER}"
+  INTERFACE_INCLUDE_DIRECTORIES "${HIP_ROOT_DIR}/include;${HSA_HEADER}"
```
pfultz2: This should use the `_IMPORT_PREFIX` so its consistent with whats get generated by cmake.

packaging/hip-targets.cmake
```diff
@@ -38,38 +38,54 @@ unset(_targetsNotDefined)
...
 
 
-# The installation prefix configured by this project.
 if( DEFINED ENV{ROCM_PATH} )
-     set(_IMPORT_PREFIX "$ENV{ROCM_PATH}/hip")
-elseif(DEFINED ENV{HIP_PATH})
-     set(_IMPORT_PREFIX "$ENV{HIP_PATH}")
-else()
-     set(_IMPORT_PREFIX "/opt/rocm/hip")
+     set(ROCM_PATH "$ENV{ROCM_PATH}")
```
pfultz2: There is no need to check for this here. Just use them both in the list of extra paths to search.

packaging/hip-targets.cmake
```diff
@@ -38,38 +38,54 @@ unset(_targetsNotDefined)
...
+
+# Set ROCM PATH from envinorment and default to /opt/rocm if not provided.
+set (HIP_ROOT_DIR)
+find_path(
+          HIP_ROOT_DIR
+          NAMES hipconfig
+          PATHS
+          ENV HIP_PATH
+          "${ROCM_PATH}/hip"
+          "${_DIR}/../../"
```
pfultz2: This should be put under `HINTS`.

PR #1774, opened 2020-01-04
hipMemcpysymbol fixes

satyanveshd:
hipMemcpyToSymbol and hipMemcpyFromSymbol are corrected to handle the error scenario.
#1735 should succeed with this.

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -1266,8 +1266,12 @@ hipError_t hipMemcpyToSymbol(void* dst, const void* src, size_t count,
 
     tprintf(DB_MEM, " symbol '%s' resolved to address:%p\n", symbol_name, dst);
 
-    return ihipLogStatus(
-        hipMemcpy(static_cast<char*>(dst) + offset, src, count, kind));
+    if (dst == nullptr) {
+        return ihipLogStatus(hipErrorInvalidSymbol);
+    }
+
+    return ihipLogStatus(hip_internal::memcpySync((char*)dst+offset, src, count, kind,
+                                                  hipStreamNull));
 }
 
 hipError_t hipMemcpyFromSymbol(void* dst, const void* src, size_t count,
@@ -1278,8 +1282,12 @@ hipError_t hipMemcpyFromSymbol(void* dst, const void* src, size_t count,
 
     tprintf(DB_MEM, " symbol '%s' resolved to address:%p\n", symbol_name, dst);
 
-    return ihipLogStatus(
-        hipMemcpy(dst, static_cast<const char*>(src) + offset, count, kind));
+    if (src == nullptr || dst == nullptr) {
+        return ihipLogStatus(hipErrorInvalidSymbol);
+    }
+
+    return ihipLogStatus(hip_internal::memcpySync(dst, (char*)src+offset, count, kind,
+                                                  hipStreamNull));
 }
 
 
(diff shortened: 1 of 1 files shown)


gargrahul:
@satyanveshd isn't it better to add this change as part of #1735 ?

src/hip_memory.cpp
```diff
@@ -1266,8 +1266,12 @@ hipError_t hipMemcpyToSymbol(void* dst, const void* src, size_t count,
 
     tprintf(DB_MEM, " symbol '%s' resolved to address:%p\n", symbol_name, dst);
 
-    return ihipLogStatus(
-        hipMemcpy(static_cast<char*>(dst) + offset, src, count, kind));
+    if (dst == nullptr) {
+        return ihipLogStatus(hipErrorInvalidSymbol);
+    }
+
+    return ihipLogStatus(hip_internal::memcpySync((char*)dst+offset, src, count, kind,
```
SarbojitAMD: I would prefer to have cpp style casting e.g. static_cast<char*> over (char*)

src/hip_memory.cpp
```diff
@@ -1266,8 +1266,12 @@ hipError_t hipMemcpyToSymbol(void* dst, const void* src, size_t count,
 
     tprintf(DB_MEM, " symbol '%s' resolved to address:%p\n", symbol_name, dst);
 
-    return ihipLogStatus(
-        hipMemcpy(static_cast<char*>(dst) + offset, src, count, kind));
+    if (dst == nullptr) {
+        return ihipLogStatus(hipErrorInvalidSymbol);
```
SarbojitAMD: I feel hipErrorInvalidValue is more appropriate than hipErrorInvalidSymbol.

src/hip_memory.cpp
```diff
@@ -1278,8 +1282,12 @@ hipError_t hipMemcpyFromSymbol(void* dst, const void* src, size_t count,
 
     tprintf(DB_MEM, " symbol '%s' resolved to address:%p\n", symbol_name, dst);
 
-    return ihipLogStatus(
-        hipMemcpy(dst, static_cast<const char*>(src) + offset, count, kind));
+    if (src == nullptr || dst == nullptr) {
```
SarbojitAMD: Since you are adding error checks don't you think symbol_name and hipMemcpyKind should also be validated?

mangupta: symbol_name is validated as part of inline implementation @ include/hip/hcc_detail/hip_runtime_api.h:1594. hipMemcpyKind is used as part of a switch case which defaults to generic copy if the kind is invalid

src/hip_memory.cpp
mangupta: The inline implementation @ include/hip/hcc_detail/hip_runtime_api.h:1594 takes in the symbolName and determines the dst pointer using hipGetSymbolAddress. So if dst is null it means the symbolName was probably invalid. Hence returning hipErrorInvalidSymbol is the right thing which is also what the CUDA equivalent API would be returning.

src/hip_memory.cpp
mangupta: Agreed.

satyanveshd:
Replaced C-style cast with static_cast as well as fixed memcpykind error codes to match with cuda for different kinds.

satyanveshd:
> @satyanveshd isn't it better to add this change as part of #1735 ?

There were other fixes as well, so raised as a separate PR.

PR #1776, opened 2020-01-06
Revert "`static inline` in a header, just like excess sugar in a diet…

ansurya:
…, causes bloat (#1692)"
SWDEV-217761
be70b9f7e711e1ae72fcfebb2fa0fdcbebbcf5fc PR breaks rocFFT build.  Reverting PR temporarily.

when __device__ and __host__ are used together without static keyword it gives below error.
__device__ __host__ static inline float hipCabsf(hipFloatComplex z) { return sqrtf(hipCsqabsf(z)); }
error:  'sqrtf':  no overloaded function has restriction specifiers that are compatible with the ambient context 'hipCabsf'

This reverts commit be70b9f7e711e1ae72fcfebb2fa0fdcbebbcf5fc.

### DIFF
--- a/include/hip/hcc_detail/channel_descriptor.h
+++ b/include/hip/hcc_detail/channel_descriptor.h
@@ -32,23 +32,23 @@ THE SOFTWARE.
 HIP_PUBLIC_API
 hipChannelFormatDesc hipCreateChannelDesc(int x, int y, int z, int w, hipChannelFormatKind f);
 
-inline hipChannelFormatDesc hipCreateChannelDescHalf() {
+static inline hipChannelFormatDesc hipCreateChannelDescHalf() {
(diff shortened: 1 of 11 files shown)


ansurya:
> I think you need to fix the actual issue, instead of bluntly reverting things. A good first step towards that being including what you use i.e. including `math_functions.h`, which that shambolic `hip_complex.h` header should do and doesn't.

Sure with  `math_functions.h`  it works fine. I thought it is only for device functions. And currently it uses sqrt from std library. Since these functions are for both host and device i didn't proceed that way. Verified with simple sample both host and device versions work fine with `math_functions.h`.

AlexVlx:
How can it use std::sqrt on the device path? `__host__` and `__device__` are disjunctive.

PR #1785, opened 2020-01-10
Fix texture tests, always pass row pitch to HSA API for linear layout…

ansurya:
… images

=> New ROCr calculates pitch as per HSA specification and addrlib is used to check whether HW can support that configuration. Hence few texture tests are failing with HSA_EXT_STATUS_ERROR_IMAGE_PITCH_UNSUPPORTED.

=> Determine pitch for linear images and always pass rowpitch to HSA API's.

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -33,62 +33,10 @@ THE SOFTWARE.
 __device__ char __hip_device_heap[__HIP_SIZE_OF_HEAP];
 __device__ uint32_t __hip_device_page_flag[__HIP_NUM_PAGES];
 
-#define IMAGE_PITCH_ALIGNMENT 256
-
 // Internal HIP APIS:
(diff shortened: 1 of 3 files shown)


src/hip_memory.cpp
```diff
@@ -33,10 +33,63 @@ THE SOFTWARE.
...
+          bytesPerPixel *= 4;
+          break;
+
+        default:
+          bytesPerPixel *= 2;
+          break;
+      }
+      return bytesPerPixel;
+    }
+
```
mangupta: We could move this to hip_hcc_internal.h. This way we don't need to duplicate this in both hip_memory.cpp and hip_texture.cpp.
Also from the below code I only see alignUp(T value, size_t alignment) variant being used. Do we need the remaining?

ansurya: done.

PR #1786, opened 2020-01-10
Clear `HIP_PATH` before version detection

aclex:
Don't allow `HIP_PATH` to be propagated to `hipconfig`, when run by CMake to detect the package version, as it leads to the wrong version is detected: when there's already HIP of some different version installed in the system and `HIP_PATH` points to its location, `hipconfig` tends to return the version of the installed HIP, rather than the value defined for the distribution. The compiled results report wrong version and spoils the rest of the stack in this case.

First found during the testing of the ebuild for HIP at https://github.com/justxi/rocm/pull/120 and fixed there with clearing `HIP_PATH` just before calling `cmake` as the first measure. But it looks, that the proper fix is like in this request.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,7 @@ endmacro()
 # Setup version information
 #############################
 # Determine HIP_BASE_VERSION
+set(ENV{HIP_PATH} "")
 execute_process(COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/bin/hipconfig --version
     OUTPUT_VARIABLE HIP_BASE_VERSION
     OUTPUT_STRIP_TRAILING_WHITESPACE)


emankov:
Investigate the failures, please.

aclex:
@emankov Yes, sure, but could you please print them here, if they're available for you? Unfortunately, the redirect on "Details" link sends me to unavailable page at hsautil.amd.com, internal resource, probably.

emankov:
Sure, @aclex, here it is, but it looks to me that you're off the hook:

[1786_error_log.zip](https://github.com/ROCm-Developer-Tools/HIP/files/4097277/1786_error_log.zip)

aclex:
@emankov thank you very much for the log and for approving it! Yes, it seems to be a couple of tests being failing. So now should I just wait for the next successful build and rebase on it?

emankov:
@aclex, no, you should wait for @mangupta's review only, not the successful build, cause the failed tests definitely not yours.

aclex:
@emankov ah, OK, thank you so much!

aclex:
Thank you very much for reviewing and merging it! Was a pleasure to participate a little.

PR #1789, opened 2020-01-14
(Maybe) Match alignment between Clang and GCC.

### LINKED ISSUE #1740: ext_vector_type GCC ABI incompatibility
Running HOOMD-blue using several GPUs over MPI with HIP 651a91b5e0e89eca190da45273d841348c73407b  results in application crashes. The culprit is a packed data structure in HOOMD-blue containing vector types.

The struct in question is
```
struct pdata_element
    {
    Scalar4 pos;               //!< Position
    Scalar4 vel;               //!< Velocity
    Scalar3 accel;             //!< Acceleration
    Scalar charge;             //!< Charge
    Scalar diameter;           //!< Diameter
    int3 image;                //!< Image
    unsigned int body;         //!< Body id
    Scalar4 orientation;       //!< Orientation
    Scalar4 angmom;            //!< Angular momentum
    Scalar3 inertia;           //!< Moments of inertia
    unsigned int tag;          //!< global tag
    Scalar4 net_force;         //!< net force
    Scalar4 net_torque;        //!< net torque
    Scalar net_virial[6];      //!< net virial
    };
```

here ScalarX is doubleX or floatX, depending on the build configuration. For now, assume it is double.

In HOOMD, .cu files get compiled using hcc (clang internally), the rest of the code with the host compiler (can be gcc or clang). I put in two `printf` statements, one in hcc compiled code, and one in gcc compiled code, to report the size of `pdata_element`:

```
sizeof(pdata_element) in .cc 328
sizeof(pdata_element) in .cu 352
```
Obviously, they disagree, explaining the memory access fault and subsequent application crash. The underlying reason is that the vector types now use ABI incompatible underlying data types (`ext_vector_type` in clang, simple C array in gcc). They are incompatible because `ext_vector_type` does not exist in gcc, so it cannot be part of the ABI. If we used the same data type in both compilers, e.g. the simple C array or `vector_size`, I would expect it to work (as it does in roc 2.10.0). I know that @AlexVlx prefers `ext_vector_type` over `vector_size` (#1675, #1690), but it seems that we should use the least common denominator between the compilers to guarantee ABI compatibility. Or better even, no special vector type at all.

Alternatively, as I suggested before, compilation with a single compiler could be enforced and using gcc would result in an error. Even though it may seem like an unnecessary restriction, I believe having no compatibility would be preferable over mysterious application crashes the user has to debug.

AlexVlx:
Should fix #1740 and the related internal bug. @jglaser, if you have time to verify it for HOOMD, it'd be greatly appreciated.


jglaser:
Hi @AlexVlx , I successfully compiled HOOMD against this branch with gcc 9.2.1. It also passes a test (`test_neighborlist`)  that failed with the wrong copy constructors for the `Scalar_accessors`. I am running the entire set of tests now, and would let you know if anything fails.

If everything is indeed OK, will this be part of ROCm 3.1?? I would then update the CI images to the next ROCm release that includes this fix once available.

AlexVlx:
@jglaser I think we can give it a shot, but it will be tight:)

jglaser:
No worries. We will update once everything is in place. There is also this blocking issue/proposed solution still.

https://github.com/ROCm-Developer-Tools/HIP/pull/1676

mangupta:
> If everything is indeed OK, will this be part of ROCm 3.1?? I would then update the CI images to the next ROCm release that includes this fix once available.
@jglaser It has missed the boat for ROCm 3.1. Only PRs marked with the rocm-3.1 milestone would potentially make it to the release branch.

PR #1794, opened 2020-01-15
Revert fae85cf since it breaks binary operators for vectors elements.

kpyzhov:


pfultz2:
Can you add unit tests for where it breaks, so future solutions will not stumble on this same issue?

PR #1795, opened 2020-01-15
[HIPIFY][cmake] Packaging for hipify-clang

aakanksha555:
Package selects the correct CMAKE_VERSION depending on the type of build:
- If hipify-clang is built along with HIP, it uses the HIP_VERSION (cmake produces a .hipversion file when HIP is built, extracting the HIP_VERSION data from it)
- If hipify-clang is built stand-alone, it uses the LLVM version.


aakanksha555:
> LGTM, if it finally works

Yeah, packages are generated with proper version now.

aakanksha555:
Merging is blocked for me, what should be done about it?

PR #1800, opened 2020-01-20
Added support to disable test based on HIP Runtime(HCC/VDI) and Compi…

ansurya:
…ler(hcc/clang)


tests/hit/HIT.cmake
```diff
@@ -203,10 +295,10 @@ macro(HIT_ADD_FILES _dir _label _parent)
         string(REGEX REPLACE "\n" ";" _contents "${_contents}")
         foreach(_cmd ${_contents})
             string(REGEX REPLACE " " ";" _cmd "${_cmd}")
-            parse_build_command(_target _sources _hipcc_options _hcc_options _nvcc_options _link_options _exclude_platforms _depends ${_dir} ${_cmd})
+            parse_build_command(_target _sources _hipcc_options _hcc_options _nvcc_options _link_options _exclude_platforms _exclude_runtime _exclude_compiler _depends ${_dir} ${_cmd})
             string(REGEX REPLACE "/" "." target ${_label}/${_target})
-            insert_into_map("_exclude" "${target}" "${_exclude_platforms}")
-            if(_exclude_platforms STREQUAL "all" OR _exclude_platforms STREQUAL ${HIP_PLATFORM})
+            if(_exclude_platforms STREQUAL "all" OR _exclude_platforms STREQUAL ${HIP_PLATFORM} OR (_exclude_runtime STREQUAL ${HIP_RUNTIME} AND (NOT _exclude_compiler OR _exclude_compiler STREQUAL ${HIP_COMPILER})))
```
mangupta: Why do we need ```AND (NOT _exclude_compiler OR _exclude_compiler STREQUAL ${HIP_COMPILER})```? Can't we just have ```OR _exclude_compiler STREQUAL ${HIP_COMPILER}```?

ansurya: This is to disable test in combination of HIP_RUNTIME and HIP_COMPILER(e.g.. exclude only for HCC runtime and clang compiler, allow other combinations). Both flags has to be checked together. And this allows EXCLUDE_HIP_COMPILER to be empty.

PR #1806, opened 2020-01-23
fix a minor bug while computing this.grid()::thread_rank()

mshivama:
* There was a bug while computing this_grid().thread_rank(), fixed it.
* The semantics of this_grid().thread_rank() computation is as follows.
   1. Compute number of work-groups to be passed before arriving at current work-group
   2. Using above, compute total number of threads being passed till current work-group
   3. Computer local thread-rank (of current thread) within current work-group
   4. Return sum of above 2 and 3.


mshivama:
I need some more fix on top of existing one. I will prepare new PR.

PR #1808, opened 2020-01-23
SWDEV-220503:  this_grid().thread_rank() gives incorrect result

mshivama:
* There was a bug while computing this_grid().thread_rank(), fixed it.
* The semantics of this_grid().thread_rank() computation is as follows.
   1.  Compute number of work-groups to be passed before arriving at current work-group
   2.  Using above, compute total number of threads being passed till current work-group
   3.  Computer local thread-rank (of current thread) within current work-group
   4.  Return sum of above 2 and 3.

### DIFF
--- a/include/hip/hcc_detail/hip_cooperative_groups_helper.h
+++ b/include/hip/hcc_detail/hip_cooperative_groups_helper.h
@@ -115,7 +115,7 @@ __CG_STATIC_QUALIFIER__ uint32_t thread_rank() {
   // Compute total number of threads being passed to reach current workgroup
   // within grid
   uint32_t num_threads_till_current_workgroup =
-           (uint32_t)(blkIdx * (hipBlockIdx_x * hipBlockIdx_y * hipBlockIdx_z));
+           (uint32_t)(blkIdx * (hipBlockDim_x * hipBlockDim_y * hipBlockDim_z));
 
   // Compute thread local rank within current workgroup
   uint32_t local_thread_rank =
--- a/include/hip/hcc_detail/hip_cooperative_groups_helper.h
+++ b/include/hip/hcc_detail/hip_cooperative_groups_helper.h
@@ -109,7 +109,7 @@ __CG_STATIC_QUALIFIER__ uint32_t thread_rank() {
   // Compute global id of the workgroup to which the current thread belongs to
   uint32_t blkIdx =
            (uint32_t)((hipBlockIdx_z * hipGridDim_y * hipGridDim_x) +
-                      (hipBlockIdx_x * hipGridDim_y) +
+                      (hipBlockIdx_y * hipGridDim_x) +
                       (hipBlockIdx_x));
 
   // Compute total number of threads being passed to reach current workgroup
@@ -120,7 +120,7 @@ __CG_STATIC_QUALIFIER__ uint32_t thread_rank() {
   // Compute thread local rank within current workgroup
   uint32_t local_thread_rank =
            (uint32_t)((hipThreadIdx_z * hipBlockDim_y * hipBlockDim_x) +
-                      (hipThreadIdx_x * hipBlockDim_y) +
+                      (hipThreadIdx_y * hipBlockDim_x) +
                       (hipThreadIdx_x));
 
   return (num_threads_till_current_workgroup + local_thread_rank);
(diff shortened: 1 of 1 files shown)


mshivama:
Hello @b-sumner @asalmanp and @jlgreathouse,

Could you please review this PR which fixes issue raised in SWDEV-220503.

include/hip/hcc_detail/hip_cooperative_groups_helper.h
```diff
@@ -106,21 +106,21 @@ __CG_STATIC_QUALIFIER__ uint32_t size() {
 }
 
 __CG_STATIC_QUALIFIER__ uint32_t thread_rank() {
-  // Compute global id of the workgroup to which the current threads belongs to
+  // Compute global id of the workgroup to which the current thread belongs to
   uint32_t blkIdx =
            (uint32_t)((hipBlockIdx_z * hipGridDim_y * hipGridDim_x) +
-                      (hipBlockIdx_y * hipGridDim_x) +
+                      (hipBlockIdx_x * hipGridDim_y) +
```
jlgreathouse: This does not seem right. I believe the previous version was correct. Could you describe why this change would be correct?

And *if* this change is correct (which I don't think it is), then why not change the line above as well? Why remove hipBlockIdx_y from the calculation, but leave hipBlockIdx_z?

include/hip/hcc_detail/hip_cooperative_groups_helper.h
```diff
@@ -106,21 +106,21 @@ __CG_STATIC_QUALIFIER__ uint32_t size() {
...
   // within grid
   uint32_t num_threads_till_current_workgroup =
-           (uint32_t)(blkIdx * (hipBlockIdx_x * hipBlockIdx_y * hipBlockIdx_z));
+           (uint32_t)(blkIdx * (hipBlockDim_x * hipBlockDim_y * hipBlockDim_z));
 
   // Compute thread local rank within current workgroup
   uint32_t local_thread_rank =
            (uint32_t)((hipThreadIdx_z * hipBlockDim_y * hipBlockDim_x) +
-                      (hipThreadIdx_y * hipBlockDim_x) +
+                      (hipThreadIdx_x * hipBlockDim_y) +
```
jlgreathouse: See my comment above about the similar change on line 122. I disagree with this change.

include/hip/hcc_detail/hip_cooperative_groups_helper.h
mshivama: Actually I got confused :) Let me think again, and correct it in case, if I made mistake.

include/hip/hcc_detail/hip_cooperative_groups_helper.h
```diff
@@ -106,21 +106,21 @@ __CG_STATIC_QUALIFIER__ uint32_t size() {
...
            (uint32_t)((hipBlockIdx_z * hipGridDim_y * hipGridDim_x) +
-                      (hipBlockIdx_y * hipGridDim_x) +
+                      (hipBlockIdx_x * hipGridDim_y) +
                       (hipBlockIdx_x));
 
   // Compute total number of threads being passed to reach current workgroup
   // within grid
   uint32_t num_threads_till_current_workgroup =
-           (uint32_t)(blkIdx * (hipBlockIdx_x * hipBlockIdx_y * hipBlockIdx_z));
+           (uint32_t)(blkIdx * (hipBlockDim_x * hipBlockDim_y * hipBlockDim_z));
```
jlgreathouse: I believe this change alone is sufficient. I tested with a 3D kernel launch and this change (and *only* this change, not the other two changes) is sufficient for correct execution.

With the two other changes as well, we get wrong results.

mshivama: Ok I will update it.

jlgreathouse:
LGTM. Before I review-approve this, a question: does the HIP project prefer to do a rebase/squash commit instead of having intermediary commits in the repo? On most large projects I work with, we don't like having broken commits in the middle of the stream, as it makes bisecting difficult.

mshivama:
Hello @mangupta, Could you please merge this PR to HIP master?

mangupta:
> LGTM. Before I review-approve this, a question: does the HIP project prefer to do a rebase/squash commit instead of having intermediary commits in the repo? On most large projects I work with, we don't like having broken commits in the middle of the stream, as it makes bisecting difficult.

@jlgreathouse Yes. We prefer to squash-commit the PRs to avoid broken commits in the history.

Issue #18, opened 2016-05-15
Incorrect conversion of "warpSize" keyword with hipify


ekondis:
Using **hipify** to convert a ".cu" file to a HIP ".cpp" file I noticed that the special variable _warpSize_ within a kernel's code is transformed to _hipWarpSize_. I think it should stay as is (_warpSize_).

adityaatluri:
Hi,
The warpSize between NVIDIA and AMD GPUs are different. Which are 32 for NVIDIA and 64 for AMD. If you take a look at HIP source, `hipWarpSize` is a macro of value 64. HIP automatically translates to platform dependent value during compilation

ekondis:
The problem is that **hipcc** does not recognise _hipWarpSize_. Trying to compile a hipified kernel returns error on _hipWarpSize_:

```
test-hipwarpsize.cpp:8:23: error: use of undeclared identifier 'hipWarpSize'
        g_data[i] = gridSize+hipWarpSize;
                             ^
1 error generated.
Died at /opt/rocm/bin/hipcc line 230.

```

adityaatluri:
@mangupta @bensander any comments?

PR #1809, opened 2020-01-23
Tweak synchronous memcpy implementation

AlexVlx:
The existing one can have issues on certain systems, therefore this limits use of direct memcpy via largeBAR to sizes where it is unequivocally better.

Also addresses SWDEV-220030.


jeffdaily:
FYI jenkins build 15 failed and 17 failed, but build 16 succeeded.  I believe this PR is actually passing CI now if we could try kicking off another try.

gargrahul:
> 
> 
> FYI jenkins build 15 failed and 17 failed, but build 16 succeeded. I believe this PR is actually passing CI now if we could try kicking off another try.

Done. #18 is in progress now.

jeffdaily:
@gargrahul did something change recently?  I'm seeing this error in the CI logs:
Error: docker: Error response from daemon: Unknown runtime specified nvidia.

src/hip_memory.cpp
```diff
@@ -204,6 +204,12 @@ namespace {
 inline
 void do_copy(void* __restrict dst, const void* __restrict src, size_t n,
              hsa_agent_t da, hsa_agent_t sa) {
+    if (da.handle == 0 || sa.handle == 0) { // Handle degenerate case.
+        throwing_result_check(hsa_memory_copy(dst, src, n),
+                              __FILE__, __func__, __LINE__);        
+        return;
+    }
```
SarbojitAMD: Not sure if we still need this change as we have observed hangs in hsa_memory_copy().

SarbojitAMD: Ignore the comment missed your latest commit :)

jedwards-AMD:
Lets merge this in.

PR #1814, opened 2020-01-24
Fix for a syntax error in deb packages

aakanksha555:


aakanksha555:
@searlmc1 could you merge the change? I don't have perms.

searlmc1:
Hi @aakanksha555 ,

The continuous-integration/jenkins/pr-merge needs to pass before it can be merged.

gargrahul:
> 
> 
> Hi @aakanksha555 ,
> 
> The continuous-integration/jenkins/pr-merge needs to pass before it can be merged.

Yes. It seems there is some issue with CI. @mangupta  could you please check ?

mangupta:
> > Hi @aakanksha555 ,
> > The continuous-integration/jenkins/pr-merge needs to pass before it can be merged.
> 
> Yes. It seems there is some issue with CI. @mangupta could you please check ?

The problem is that every time HIP master branch updates (i.e. a PR is merged into master branch), the CI will rebuild all PRs. So if there are 10 PRs open and we merge one PR, 9 builds are triggered. If we merge another one after say a few minutes, anothe 8 more builds are triggered. Considering that we have ~40 open PRs right now, you can do the math on the number of builds CI has to handle. The way around this is to batch merge all PRs that we want to within a minute or two so that CI would end up rebuilding all PRs just once.

PR #1816, opened 2020-01-24
optimize `atomicAdd` for float/double

psychocoderHPC:
Reduce the number of atomic operations performed to emulate `atomicAdd`
for data type `float` and `double`.

The optimization is also mentioned, since a few years, in the [CUDA programming guide](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomic-functions).

The way to implement `atomicAdd` with `atomicCAS` is common but I like to point to older implementations with `atomicExch` too and some old comparisons from 2010. [ref](https://devtalk.nvidia.com/default/topic/458062/atomicadd)

I did a integration test of this implementation vs. the old together with [PIConGPU](https://github.com/ComputationalRadiationPhysics/picongpu/issues). PIConGPU is doing in one kernel a scatter with N threads to a chunk of shared memory with `atomicAdd(float*,float)`.
The improvement for single precision float data (time to solution) on a Radeon VII was **>10%** and and on MI50 **>12%**.

Doing the scatter directly on global memory instead of shared results in a improvement of **>25%**.

### DIFF
--- a/include/hip/hcc_detail/hip_atomic.h
+++ b/include/hip/hcc_detail/hip_atomic.h
@@ -59,17 +59,16 @@ float atomicAdd(float* address, float val)
 {
     unsigned int* uaddr{reinterpret_cast<unsigned int*>(address)};
     unsigned int r{__atomic_load_n(uaddr, __ATOMIC_RELAXED)};
-
-    unsigned int old;
-    do {
-        old = __atomic_load_n(uaddr, __ATOMIC_RELAXED);
-
-        if (r != old) { r = old; continue; }
-
-        r = atomicCAS(uaddr, r, __float_as_uint(val + __uint_as_float(r)));
-
-        if (r == old) break;
-    } while (true);
+    unsigned  int assumed;
+    do
+    {
+        assumed = r;
+        r = atomicCAS(
+            uaddr,
+            r,
+            __float_as_uint(val + __uint_as_float(r)));
+    }
+    while(assumed != r);
 
     return __uint_as_float(r);
 }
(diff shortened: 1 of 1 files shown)


iotamudelta:
AFAIK this reverts from TTAS to TAS paradigm. We deliberately recently switched to TTAS due to some important workloads.

psychocoderHPC:
What do you mean by `TTAS` and `TAS`?

iotamudelta:
See, e.g., https://en.wikipedia.org/wiki/Test_and_test-and-set for TTAS

psychocoderHPC:
I found now also the original PR were the old implementation was removed: https://github.com/ROCm-Developer-Tools/HIP/pull/1573

@iotamudelta 

> We deliberately recently switched to TTAS due to some important workloads.

Could you please describe the workloads? Is it reducing to one memory address?

I try to understand in which cases the current HIP implementation is faster, for me the branches within the infinity loop looks hard to optimize, for a compiler, for an architecture where a warp is sharing one process counter.

psychocoderHPC:
In my tested workflow all threads within a block performed a atomicAdd to a sub-volume. So the possibility that other threads manipulating the same address in parallel is small. 
In the case where several blocks writing to the same address the current HIP implementation will perform maybe better.

@iotamudelta Can I find somewhere the performance tests used to validate TTAS vs. TAS?

psychocoderHPC:
@AlexVlx under which conditions TTAS performace better than the very old TAS implementarion?

mangupta:

PR #1825, opened 2020-01-27
Cleaned up error messages for HipEnvVarDriver test

jiabaxie:
There were several error messages that appeared even if the hipEnvVarDriver.exe test passes and executes successfully. Now it is cleaned up. The following are those instances:
- When popen searches for directed_test directory but does not find it, it outputs an error, then finds the hipEnvVar at the same level. Currently the fix will prompt the test to only output an error if both searches for hipEnvVar fails.
- When assertion is used towards the later half of the test, conditions were set to specifically hide the devices, resulting in No Hip Device detected in the latter half of the test. The fix will make these errors not appear as they are intended to not find any devices. Assertions themselves are untouched.

HipEnvVarDriver.cpp has also been refactored. Reading HipEnvVar will now happen in a helper function for getDeviceNumber and getDevicePCIBusNumRemote, as the code to read HipEnvVar were really similar in them.

### DIFF
--- a/tests/src/hipEnvVarDriver.cpp
+++ b/tests/src/hipEnvVarDriver.cpp
@@ -70,7 +70,7 @@ int getDeviceNumber(bool print_err=true) {
         }
     }
     if (print_err) {
-		std::cout << buff;
+        std::cout << buff;
     }
     return atoi(buff);
 }
--- a/tests/src/hipEnvVarDriver.cpp
+++ b/tests/src/hipEnvVarDriver.cpp
@@ -38,23 +38,22 @@ const string directed_dir = string(".") + PATH_SEPERATOR_STR + "directed_tests" 
 const string dir = string(".") + PATH_SEPERATOR_STR + "hipEnvVar";
 
 int readHipEnvVar(string flags, char* buff){
-    //Don't print errors if popen is missing file
-    int fd = dup(fileno(stderr));
-    freopen(NULL_DEVICE, "w", stderr);
+
+    std::cout << "\nFinding hipEnvVar in " << directed_dir << "...\n";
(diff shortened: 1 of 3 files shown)


jiabaxie:
Hi,
Any updates on this PR?

tests/src/hipEnvVarDriver.cpp
```diff
@@ -21,6 +21,7 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
  * HIT_END
  */
 
+#include <io.h>
```
mangupta: This breaks CI. I am not sure why you need to include io.h in this file especially on linux.

jiabaxie: Hi,
Currently, hipEnvVarDriver will try to find hipEnvVar in a few places. Under directed_test first, then on the same level. If hipEnvVar doesn't exist in directed_test (but finds it in the same level), it will output error messages about a file not found, even though it's part of the searching process. I turned off stderr for this part since it is expected behavior, and only prints an error when hipEnvVar does not exist at all.

I see that io.h breaks linux, but putting guards on this include (i.e. move it to the WIN64 macro under test_common.h that detects OS) should solve this problem.
Alternatively, I can simply not turn off these error messages, but alert the user that error messages may be expected (i.e. print a message when an error may be expected so that the user is not confused)

Which approach would you say is best?

jiabaxie:
I removed io.h and simply prints messages notifying users about the error messages. Does this look ok?

jiabaxie:
Will this be merged soon?

PR #1826, opened 2020-01-28
Device texture functions should not normalize the sampled pixel. This…

vsytch:
… is already done by HW.

### DIFF
--- a/include/hip/hcc_detail/hip_texture_types.h
+++ b/include/hip/hcc_detail/hip_texture_types.h
@@ -57,6 +57,7 @@ struct __HIP_TEXTURE_ATTRIB texture : public textureReference {
     texture(int norm = 0, enum hipTextureFilterMode fMode = hipFilterModePoint,
             enum hipTextureAddressMode aMode = hipAddressModeClamp) {
         normalized = norm;
+        readMode = hipReadModeNormalizedFloat;
         filterMode = fMode;
         addressMode[0] = aMode;
         addressMode[1] = aMode;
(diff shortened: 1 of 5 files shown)


mangupta:
@vsytch I don't think the HW does this normalization. The case from which you are removing the normalization is when the data is integer format (8/16/32-bit) but you want to read it back as float. CUDA supports this and this change was added as part of #1213

vsytch:
> @vsytch I don't think the HW does this normalization. The case from which you are removing the normalization is when the data is integer format (8/16/32-bit) but you want to read it back as float. CUDA supports this and this change was added as part of #1213

@mangupta HW will automatically do {int8, uint8, int16, uint16} -> float conversion when sampling a pixel from a image with a normalized int format. Neither Cuda (see cuTexObjectCreate docs) nor OpenCL support {int32, uint32} -> float conversion.

I discovered this bug when looking into hipNormalizedFloatValueTex test failure with HIP-VDI. Creating a texture with read mode set to hipReadModeNormalizedFloat would produce garbage pixel values, but with hipReadModeElementType I would observe the correct normalized pixel values. I checked that the T# had the correct format, so it had to be a bug in tex1D(). Like I suspected it had extra logic to mess with the pixel value.

HIP-HCC is responsible to create an HSA image with the format HSA_EXT_IMAGE_CHANNEL_TYPE_UNORM_INT8, HSA_EXT_IMAGE_CHANNEL_TYPE_SNORM_INT8, HSA_EXT_IMAGE_CHANNEL_TYPE_UNORM_INT16, HSA_EXT_IMAGE_CHANNEL_TYPE_SNORM_INT16 when creating a texture with read mode set to hipReadModeNormalizedFloat . However in the case of the above test, HIP-HCC does not handle flags set by hipTexRefSetFlags(), so UB is observed.

This header change doesn't break any tests for HIP-HCC and fixes HIP-VDI. I've also requested @gargrahul to disable the hipNormalizedFloatValueTex test untill HIP-HCC fixes its bugs.

PR #1829, opened 2020-01-30
Implement long / long long shuffles

arghdos:
Implement additional data-types for shuffles (`long` and `long long`).
Based upon the `double` implementation.


arghdos:
Find proposed tests in SWDEV-221363

PR #1831, opened 2020-01-31
Cmake package

cjatin:
Bringing package makefile in sync with genereted one's.


packaging/hip-targets.cmake
```diff
@@ -47,6 +50,9 @@ endif()
 get_filename_component(_DIR "${CMAKE_CURRENT_LIST_DIR}" REALPATH)
 get_filename_component(_IMPORT_PREFIX "${_DIR}/../../../" REALPATH)
 
+# The installation prefix configured by this project.
+set(_IMPORT_PREFIX "${CMAKE_INSTALL_PREFIX}")
```
pfultz2: This is called by other packages so `CMAKE_INSTALL_PREFIX` will not be set to hip's installation path. A relative path should be used.

pfultz2: For example, the onnx runtime will call `find_package(hip)` but the `CMAKE_INSTALL_PREFIX` will be set to `/usr/local` or some other directory. It will never be set to hip's installation path.

cjatin: Removed the hard overwrite of `_IMPORT_PREFIX`

pfultz2: Thanks!

PR #1833, opened 2020-01-31
Enable multi platform excludes

lmoriche:
Replaced _exclude_platforms' string with a list. Use IN_LIST to check
if a platform is excluded.


mangupta:
PR #1800 already provides a mechanism to disable tests based on on runtime and compiler combinations. Only enhancement in that would be to rename the EXCLUDE_HIP_PLATFORM targets to amd and nvidia instead of hcc and nvcc.

PR #1835, opened 2020-01-31
Fix C-style hipLaunchKernel

scchan:
This fixes the implementation of hipLaunchKernel and the unit test


src/hip_module.cpp
```diff
@@ -1454,21 +1455,7 @@ hipError_t hipLaunchKernel(
...
-   void* config[]{
-        HIP_LAUNCH_PARAM_BUFFER_POINTER,
-        args,
-        HIP_LAUNCH_PARAM_BUFFER_SIZE,
-	    &szKernArg,
-        HIP_LAUNCH_PARAM_END};
-
-   return ihipLogStatus(ihipModuleLaunchKernel(tls, kd, numBlocks.x * dimBlocks.x, numBlocks.y * dimBlocks.y, numBlocks.z * dimBlocks.z,
-                          dimBlocks.x, dimBlocks.y, dimBlocks.z, sharedMemBytes, stream, nullptr, (void**)&config, nullptr, nullptr, 0));
+   return hipModuleLaunchKernel(kd, numBlocks.x, numBlocks.y, numBlocks.z,
```
SarbojitAMD: @scchan, I think we don't want to call two external HIP APIs one in side other, like in this PR hipLaunchKernel calling hipModuleLaunchKernel. May I know the reason for calling it?

tests/src/gcc/LaunchKernel.c
```diff
@@ -138,14 +138,43 @@ bool LaunchKernelArg3()
...
+  dim3 blocks       = {1,1,1};
+  dim3 threads      = {1,1,1};
+
+  // Allocate Device memory
+  HIPCHECK(hipMalloc((void**)&A_d, sizeof(int)));
+
+  char c = 1;
+  short s = 10;
+  int i = 100;
+  struct things t = {2,20,200};
```
SarbojitAMD: Don't we need to allocate these variables on device?

src/hip_module.cpp
scchan: The original implementation has an incorrect assumption that all every array element from args is a pointer to a gpu buffer.  Also, the kernel arg should be constructed based on the metadata of the kernel.  I see hipLaunchKernel as a simple wrapper around hipModuleLaunchKernel.

tests/src/gcc/LaunchKernel.c
scchan: Notice the kernel signature for kernel4, these are passed by value.

PR #1836, opened 2020-01-31
Enable code object v3 codgen when compiling with HCC

scchan:


scchan:
Note that the LaunchKernel test would fail without https://github.com/ROCm-Developer-Tools/HIP/pull/1835

PR #1842, opened 2020-02-02
Check pinned status in hipMemcpyAsync

gargrahul:
This should help to take optimal path of sync copies when host buffer is unpinned for H2D and D2H copies.

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -1415,12 +1415,12 @@ hipError_t hipMemcpyAsync(void* dst, const void* src, size_t sizeBytes, hipMemcp
     bool isLocked = true;
     void *pinnedPtr= nullptr;
     if(kind == hipMemcpyHostToDevice ) {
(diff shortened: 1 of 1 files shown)


AlexVlx:
This makes everything that is unpinned synchronous. This does not meet the contract of the API and leads to surprising behaviour (my supposedly non-blocking API blocking). Even if this is pre-existing behaviour, it is ghastly and should be fixed, as opposed to reinforced. We should not piggyback on unrelated work in order to claim volatile, brittle wins.

iotamudelta:
Seconded that. It's been a recurring meme in analyzing PyTorch performance that this unexpected behavior is, well, unexpected and wastes cycles of developers. While I appreciate that the improvements originating from #1673 can be trivially used here, `hipMemcpyAsync` being truly async is the right thing to do.

For the use case that `hipMemcpyAsync` is followed by a stream synchronize, we now have `hipMemcpyWithStream` to correctly and efficiently target it.

mangupta:

PR #1625, opened 2019-11-05
Match Occupancy APIs syntax with CUDA

satyanveshd:
Raising this PR after PR#1560 was reverted.
Matched the interface for cooperative groups with CUDA. This includes the templates for hipOccupancyMaxActiveBlocksPerMultiprocessor, hipOccupancyMaxActiveBlocksPerMultiprocessorWithFlags. 

hipOccupancyMaxPotentialBlockSize is not changed as there is no C equivalent API for this in CUDA. 

Also, modified the tests related to these API's according to the changes made.

PR #1844, opened 2020-02-04
Update hipLaunchCoopMultiKernel.cpp at the line 159

lthakur007:
used reinterpret_cast<uint32_t*> for numBlocks, as expected by hipOccupancyMaxActiveBlocksPerMultiprocessor() api.


satyanveshd:
Can this be re-checked once, since in #1625 the API is modified to return the type int, the re-interpret cast might not be required.

PR #1847, opened 2020-02-05
[hip] Revise compiler wrapper for HIP-Clang.

mhbliao:
- As the builtin headers from clang are installed in the standard
  destination, there's no need to add them as system header paths.
  There's no need to add HIP headers as system headers as well.

### DIFF
--- a/bin/hipcc
+++ b/bin/hipcc
@@ -138,10 +138,6 @@ if (defined $HIP_RUNTIME and $HIP_RUNTIME eq "VDI" and !defined $HIP_VDI_HOME) {
 if (defined $HIP_VDI_HOME) {
     if (!defined $HIP_CLANG_PATH and (-e "$HIP_VDI_HOME/bin/clang" or -e "$HIP_VDI_HOME/bin/clang.exe")) {
         $HIP_CLANG_PATH = "$HIP_VDI_HOME/bin";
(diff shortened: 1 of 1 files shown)


bin/hipcc
```diff
@@ -333,7 +332,11 @@ if ($HIP_PLATFORM eq "clang") {
 }
 
 # Add paths to common HIP includes:
-$HIPCXXFLAGS .= " -isystem $HIP_INCLUDE_PATH" ;
+if ($HIP_PLATFORM eq "clang") {
+    $HIPCXXFLAGS .= " -I $HIP_INCLUDE_PATH" ;
```
yxsamliu: this may result in lots of warning msg for each HIP compilation. Should we clean that up before making this change?

mhbliao: Could elaborate in details?

yxsamliu: -isystem suppresses warnings in these headers. Do you see any extra warnings from hip_runtime.h when making this change?

mhbliao: So far, from the HIP test compilation results, that's only warning I observed.

mhbliao: Hi @yxsamliu 
Does the new change look OK?

mhbliao:
The new revision takes conservative step. Please review it ASAP as so far builds are broken without this change.

mhbliao:
could someone merge this to unbreak lots of build? Thx!

PR #1850, opened 2020-02-06
make abort and assert functions,

arghdos:
Such that they do not clobber user definitions.  Needed for Kokkos.

### DIFF
--- a/include/hip/nvcc_detail/hip_runtime.h
+++ b/include/hip/nvcc_detail/hip_runtime.h
@@ -102,15 +102,12 @@ typedef int hipLaunchParm;
 #define HIP_DYNAMIC_SHARED_ATTRIBUTE
 
 #ifdef __HIP_DEVICE_COMPILE__
-#define abort()                                                                                    \
-    { asm("trap;"); }
+inline void __device__ abort() { asm("trap;"); }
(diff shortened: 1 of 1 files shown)


arghdos:
Example test:

```
#include <iostream>
#include <stdexcept>
#include <sstream>
#include <hip/hip_runtime.h>

#define hipCheck(stmt) do {                          \
  hipError_t err = stmt;                             \
  if (err != hipSuccess)  {                          \
    char msg[256];  \
      sprintf(msg, "%s in file %s, function %s, line %d\n", #stmt,__FILE__,__FUNCTION__,__LINE__); \
      std::string errstring = hipGetErrorString(err); \
      std::cerr << msg << "\t" << errstring << std::endl;  \
      throw std::runtime_error(msg);            \
  }                                                 \
} while(0)

namespace test
{
  __device__ void abort()
  {
    // implement my own abort / assert handlers
  }
  #undef assert
  __device__ void assert(bool cond)
  {
  }
}

__global__ void kernel() {
  test::assert(true);
  test::abort();
}

int main() {
  kernel<<<dim3(1), dim3(1)>>>();
  hipCheck(hipDeviceSynchronize());
  return 0;
}
```

PR #1851, opened 2020-02-06
Use deque instead of vector for code readers so that the iterators and references will be stable

pfultz2:
Even though there is a lock for pushing back to the vector, the references and iterators to the element are used without a lock. This is a problem as other threads will push back to the vector which can invalidate the references and iterators from other threads.

Instead of locking the entire function, I switched to `std::deque` which has stable iterators and references when using `emplace_back`.

### DIFF
--- a/tests/src/runtimeApi/module/hipModuleLoadDataMultThreaded.cpp
+++ b/tests/src/runtimeApi/module/hipModuleLoadDataMultThreaded.cpp
@@ -18,7 +18,7 @@ THE SOFTWARE.
 */
 
 /* HIT_START
- * BUILD: %t %s ../../test_common.cpp EXCLUDE_HIP_PLATFORM nvcc
+ * BUILD: %t %s ../../test_common.cpp NVCC_OPTIONS -std=c++11
  * TEST: %t
  * HIT_END
  */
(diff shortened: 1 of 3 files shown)


tests/src/runtimeApi/module/hipModuleLoadDataMultThreaded.cpp
```diff
@@ -0,0 +1,148 @@
...
+IMPLIED, INNCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANNY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER INN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR INN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+/* HIT_START
+ * BUILD: %t %s ../../test_common.cpp EXCLUDE_HIP_PLATFORM nvcc
```
mangupta: Why is this disabled on cuda path? Looking at your previous changes in this ticket, I think the issue is that you are trying to use the same context on all the threads. Creating a new context for the same device on all the threads should fix the cuda path.

pfultz2: Even when creating a context on each thread, I get an error about invalid ptx from JIT compilation. That seems to be an environment issue. 

This unit test is to mainly to check the AMD backend so it seems acceptable for it not to run on the cuda-backend. Also, if it found an issue, we most likely would not fix it either.

tests/src/runtimeApi/module/hipModuleLoadDataMultThreaded.cpp
```diff
@@ -18,7 +18,7 @@ THE SOFTWARE.
 */
 
 /* HIT_START
- * BUILD: %t %s ../../test_common.cpp EXCLUDE_HIP_PLATFORM nvcc
+ * BUILD: %t %s ../../test_common.cpp
```
pfultz2: You need to add c++11 flags on nvidia.

PR #1852, opened 2020-02-06
[hipcc] Skip warning on `gfx000`.

mhbliao:
- The known target checking should skip `gfx000` as well as it won't be
  used in real compilation command formation. The avoid generating
  annoying warning on `gfx000`.


mhbliao:
So far, `rocm_agent_enumerator` always output `gfx000` as the rocm agent. Since we already skip generating compilation option for `gfx000`, we should also avoid generating warning on it. That's quite annoying.

PR #1853, opened 2020-02-07
[hipcc] - removing dependency on rocm-dkms package

pruthvistony:
-> ROCM_PATH which points to <rocm_install_dir> is defaulted to /opt/rocm and
if it is defined in the ENV and it exists, then set it to ENV var.
-> Here we are eliminating the dependency on .info/version which is installed
by rocm-dkms package which cant be installed in a docker.
-> Dont want to rely on parent of HIP_PATH since HIP can be installed into
any path.

### DIFF
--- a/bin/hipcc
+++ b/bin/hipcc
@@ -70,15 +70,20 @@ $HIPCC_LINK_FLAGS_APPEND=$ENV{'HIPCC_LINK_FLAGS_APPEND'};
 # linked by symlink, the absolute path of this script should be used to
 # derive HIP_PATH, as dirname $0 could be /opt/rocm/bin or /opt/rocm/hip/bin
 # depending on how it gets invoked.
-# ROCM_PATH which points to <rocm_install_dir> is determined based on whether
-# we find .info/version in the parent of HIP_PATH or not. If it is found,
-# ROCM_PATH is defined relative to HIP_PATH else it is hardcoded to /opt/rocm.
+#
+# ROCM_PATH which points to <rocm_install_dir> is defaulted to /opt/rocm and
+# if it is defined in the ENV and it exists, then set it to ENV var.
+# -> Here we are eliminating the dependency on .info/version which is installed
+# by rocm-dkms package which cant be installed in a docker.
+# -> Dont want to rely on parent of HIP_PATH since HIP can be installed into
+# any path.
+# ROCM_PATH is the variable used ahead for refering to all other ROCm stack
+# components.
 #
 $HIP_PATH=$ENV{'HIP_PATH'} // dirname(Cwd::abs_path("$0/../")); # use parent directory of hipcc
-if (-e "$HIP_PATH/../.info/version") {
-    $ROCM_PATH=$ENV{'ROCM_PATH'} // dirname("$HIP_PATH"); # use parent directory of HIP_PATH
-} else {
-    $ROCM_PATH=$ENV{'ROCM_PATH'} // "/opt/rocm";
+$ROCM_PATH="/opt/rocm";
+if (-e "$ENV{'ROCM_PATH'}") {
+    $ROCM_PATH=$ENV{'ROCM_PATH'};
 }
 $HIP_VDI_HOME=$ENV{'HIP_VDI_HOME'};
 $HIP_LIB_PATH=$ENV{'HIP_LIB_PATH'};


bin/hipcc
```diff
@@ -70,15 +70,20 @@ $HIPCC_LINK_FLAGS_APPEND=$ENV{'HIPCC_LINK_FLAGS_APPEND'};
...
 # derive HIP_PATH, as dirname $0 could be /opt/rocm/bin or /opt/rocm/hip/bin
 # depending on how it gets invoked.
-# ROCM_PATH which points to <rocm_install_dir> is determined based on whether
-# we find .info/version in the parent of HIP_PATH or not. If it is found,
-# ROCM_PATH is defined relative to HIP_PATH else it is hardcoded to /opt/rocm.
+#
+# ROCM_PATH which points to <rocm_install_dir> is defaulted to /opt/rocm and
+# if it is defined in the ENV and it exists, then set it to ENV var.
+# -> Here we are eliminating the dependency on .info/version which is installed
+# by rocm-dkms package which cant be installed in a docker.
```
mangupta: If we cannot rely on .info/version, we can look for .info/version-utils (installed by rocm-utils) or .info/version-dev (installed by rocm-dev). At the very least we need rocm-dev to be installed in the docker for hipcc to be able to compile any application code. So it should be fine to test for .info/version-dev.

pruthvistony: Yes, I agree rocm-dev once installed - .info/version-dev is created and can be used here.
But I am thinking of keeping the code change of hip independent of any meta packages installation.
-> ROCm can be installed without installing any rocm-dev, rocm-libs and rocm-dkms meta packages.
Please let me know if there is any reason to check this file exists condition.

mangupta: From what I understand, if you don't install the meta packages, the /opt/rocm symlink is not created. If so, the users would anyways be required to define ROCM_PATH even if they have just one ROCm version installed. So for such users neither does the existing scheme nor the proposed logic in this PR useful. They will need to define ROCM_PATH.

However for folks who install rocm-dev even if they install multiple ROCm versions side-by-side, switching to using .info/version-dev or .info/version-utils in the existing logic would enable them to no longer require setting ROCM_PATH for using the same ROCm version as the hipcc being used.

So as long as we are able to auto-detect ROCM_PATH for more folks than what we are today it is sufficient reason to make the change. Forcing more users to set ROCM_PATH is not the way to go.

bin/hipcc
```diff
@@ -70,15 +70,20 @@ $HIPCC_LINK_FLAGS_APPEND=$ENV{'HIPCC_LINK_FLAGS_APPEND'};
...
+# any path.
+# ROCM_PATH is the variable used ahead for refering to all other ROCm stack
+# components.
 #
 $HIP_PATH=$ENV{'HIP_PATH'} // dirname(Cwd::abs_path("$0/../")); # use parent directory of hipcc
-if (-e "$HIP_PATH/../.info/version") {
-    $ROCM_PATH=$ENV{'ROCM_PATH'} // dirname("$HIP_PATH"); # use parent directory of HIP_PATH
-} else {
-    $ROCM_PATH=$ENV{'ROCM_PATH'} // "/opt/rocm";
+$ROCM_PATH="/opt/rocm";
```
srinivamd: Is this reintroducing dependency on /opt/rocm or environment variable? Is there a use-case where rocm-dkms is not installed when this package is installed and used? What scenario is this fix addressing?

bin/hipcc
pfultz2: > Yes, In the proposed logic, ROCM_PATH needs to be defined irrespective of the symlink is present or not present as a required practice.

Defining custom env variable should never be considered required practice. 

> So if any application which wants to link with ROCm libs, it should give the option -L(rocm path for libs) and -l(lib name) and build.

There is more than just library paths that are needed to build against libraries. Applications should use `find_package` or pkg-config for non-cmake builds, and if they have multiple installations they can set `CMAKE_PREFIX_PATH` or `PKG_CONFIG_PATH` to their rocm installation.

Ideally, hipcc would use pkg-config for its usage requirements and build with `hcc $(pkg-config hip-hcc --cflags --ldflags) "$@"`. It can seed the pkg-config path with the relative+/opt/rocm(ie `PKG_CONFIG_PATH="$PKG_CONFIG_PATH:$PWD/.../lib/pkgconfig:/opt/rocm/lib/pkgconfig"`. So the user's path is searched first and then it falls back on relative or /opt/rocm. This can also work when staging hip in the CI or when using directly from the package.

mangupta:

PR #1854, opened 2020-02-07
Add hipDrvOccupancyMaxActiveBlocksPerMultiprocessor

gargrahul:
Equivalent to cuOccupancyMaxActiveBlocksPerMultiprocessor.


mhbliao:
Following the occupancy APIs provided through CUDA, we need to rename the existing `hipOccupancyMaxActiveBlocksPerMultiprocessor` @ line 2950 to the one you proposed instead of adding a new one. Adding new one won't solve the issue. In addition, PyTorch's hipifying script needs to replace `cuOccupancyMaxActiveBlocksPerMulitprocessor` with the one you proposed.

Furthermore, the similar renaming should be applied to `hipOccupancyMaxActiveBlocksPerMultiprocessorWithFlags` as well.

gargrahul:
> 
> 
> Following the occupancy APIs provided through CUDA, we need to rename the existing `hipOccupancyMaxActiveBlocksPerMultiprocessor` @ line 2950 to the one you proposed instead of adding a new one. Adding new one won't solve the issue. In addition, PyTorch's hipifying script needs to replace `cuOccupancyMaxActiveBlocksPerMulitprocessor` with the one you proposed.
> 
> Furthermore, the similar renaming should be applied to `hipOccupancyMaxActiveBlocksPerMultiprocessorWithFlags` as well.

Existing one is equivalent for __host__​__device__​cudaError_t **cuda**OccupancyMaxActiveBlocksPerMultiprocessor ( int* numBlocks, const void* func, int  blockSize, size_t dynamicSMemSize )

mhbliao:
The one corresponding to `cudaOccupancyMaxActiveBlocksPerMultiprocessor` is already provided at line 3374. The implementation on line 3376 needs to use the new API you proposed. `cudaOccupancyMaxActiveBlocksPerMultiprocessor` is a C++ interface and provided as function template.

mhbliao:
LGTM

PR #1860, opened 2020-02-10
Revert "Cmake package"

cjatin:
Reverts ROCm-Developer-Tools/HIP#1831

### DIFF
--- a/packaging/hip-targets-release.cmake
+++ b/packaging/hip-targets-release.cmake
@@ -20,6 +20,7 @@ set_target_properties(hip::hip_hcc_static PROPERTIES
 else()
 set_target_properties(hip::hip_hcc_static PROPERTIES
   IMPORTED_LINK_INTERFACE_LANGUAGES_RELEASE "CXX"
+  IMPORTED_LINK_INTERFACE_LIBRARIES_RELEASE "hc_am"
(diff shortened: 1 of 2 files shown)


pfultz2:
Why is this being reverted? This fixes problems with the cmake, and it makes it consistent with hip when installed with `make install`. This shouldn't be reverted.

This also affects the onnx runtime.

mangupta:
> Why is this being reverted? This fixes problems with the cmake, and it makes it consistent with hip when installed with `make install`. This shouldn't be reverted.
> 
> This also affects the onnx runtime.

#1831 broke builds of several components including MiOpen driver tests, RCCL and rocSPARSE. Since we need to freeze for 3.2 branching, reverting this for now to unblock builds. We will continue to investigate the impacted components to get them to work with #1831.

pfultz2:
> #1831 broke builds of several components including MiOpen driver tests

I find that odd, since we test MIOpen using HIP from `make install` and not from `make package`, and we do not see any such issue. Where is the info about the issues found?

gargrahul:
> 
> 
> > #1831 broke builds of several components including MiOpen driver tests
> 
> I find that odd, since we test MIOpen using HIP from `make install` and not from `make package`, and we do not see any such issue. Where is the info about the issues found?

Please check SWDEV-222155 and SWDEV-222158.

PR #1861, opened 2020-02-10
Remove free memory hiding

gargrahul:


src/hip_hcc.cpp
```diff
@@ -84,7 +84,7 @@ int HIP_FORCE_P2P_HOST = 0;
 int HIP_FAIL_SOC = 0;
 int HIP_DENY_PEER_ACCESS = 0;
 
-int HIP_HIDDEN_FREE_MEM = 256;
+int HIP_HIDDEN_FREE_MEM = 0;
```
SarbojitAMD: Could you please add comment here why default value is not needed now? This will be good for future reference.

gargrahul:
Closing since this is needed.

PR #1862, opened 2020-02-12
remove future.valid() check from locked_copy2DAsync

jeffdaily:
A recent change to HCC runtime causes hc::completion_future::valid() to always return false here.
See https://github.com/RadeonOpenCompute/hcc/pull/1317.


gargrahul:
@jeffdaily As discussed we shouldn't merge this change. Please confirm.

jeffdaily:
@gargrahul can confirm.  Closing this PR in favor of fixing HCC runtime to return an invalid future when 2D async copy fails.

PR #1866, opened 2020-02-12
Add missing __hip_pinned_shadow__ attributes to the texture global vars.

vsytch:


mangupta:
@vsytch This doesn't look right to me. Do we expect applications to add additional attributes in their code when they switch hip runtime? How about adding it here: https://github.com/ROCm-Developer-Tools/HIP/blob/master/include/hip/hcc_detail/hip_texture_types.h#L49?

vsytch:
I believe this is what we recommended R******* to do when using hip-clang. They use __hip_pinned_shadow__ in their code (not sure if it was explicitly added, or a side effect of hipify).

yxsamliu:
> @vsytch This doesn't look right to me. Do we expect applications to add additional attributes in their code when they switch hip runtime? How about adding it here: https://github.com/ROCm-Developer-Tools/HIP/blob/master/include/hip/hcc_detail/hip_texture_types.h#L49?

`__hip_pinned_shadow__` can only be applied to variables, not types. `__attribute__((device_builtin_texture_type))` was rejected by clang reviewers.

PR #1869, opened 2020-02-13
[sample] Add hipDispatchEnqueueRateMT

gargrahul:
Adds a sample to measure dispatch enqueue rate for multi thread multi stream dispatches.

### DIFF
--- a/samples/1_Utils/hipDispatchLatency/hipDispatchEnqueueRateMT.cpp
+++ b/samples/1_Utils/hipDispatchLatency/hipDispatchEnqueueRateMT.cpp
@@ -1,0 +1,174 @@
+/*
+Copyright (c) 2020-present Advanced Micro Devices, Inc. All rights reserved.
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+#include <stdio.h>
+#include "hip/hip_runtime.h"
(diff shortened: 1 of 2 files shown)


samples/1_Utils/hipDispatchLatency/hipDispatchEnqueueRateMT.cpp
```diff
@@ -0,0 +1,174 @@
...
+    std::atomic_int shared {0};
+    std::vector<std::future<void>> threads;
+    int max_threads = 1;
+};
+
+
+int main(int argc, char* argv[])
+{
+    if (argc != 3) {
+        std::cerr << "Run test as 'hipDispatchEnqueueRateMT <num_threads> <0-module_launch /1-ggl_launch /2-both>'\n";
```
hthangirala: Worth printing full names since the output of this test doesn't abbreviate. 
0-hipModuleLaunchKernel 
1-hipLaunchKernelGGL

I don't think both should be an option, since: 
a) not common usage
b) later test does not start from scratch, but instead would used HIP resources reserved from previous test.

samples/1_Utils/hipDispatchLatency/hipDispatchEnqueueRateMT.cpp
```diff
@@ -0,0 +1,174 @@
...
+    for (auto i = 0; i < TOTAL_RUN_COUNT; ++i) {
+        auto start = std::chrono::high_resolution_clock::now();
+        hipModuleLaunchKernel(function, 1, 1, 1, 1, 1, 1, 0, stream, &kernel_params, nullptr);
+        auto stop = std::chrono::high_resolution_clock::now();
+        results[i] = std::chrono::duration<double, std::milli>(stop - start).count();
+    }
+    print_timing("Thread ID : " + std::to_string(tid) + " , " + "hipModuleLaunchKernel enqueue rate", results);
+    hipStreamSynchronize(stream);
+    hipModuleUnload(module);
+    hipStreamDestroy(stream);
```
hthangirala: Cannot release resources as threads are not synchronized and perturbs other threads are still in the critical path timing enqueue operations. Since we're only allowing one test to run at a time, its okay to let it cleaned up at process exit.

samples/1_Utils/hipDispatchLatency/hipDispatchEnqueueRateMT.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+    for (auto i = 0; i < TOTAL_RUN_COUNT; ++i) {
+        auto start = std::chrono::high_resolution_clock::now();
+        hipLaunchKernelGGL((EmptyKernel), dim3(NUM_GROUPS), dim3(GROUP_SIZE), 0, stream);
+        auto stop = std::chrono::high_resolution_clock::now();
+        results[i] = std::chrono::duration<double, std::milli>(stop - start).count();
+    }
+    print_timing("Thread ID : " + std::to_string(tid) + " , " + "hipLaunchKernelGGL enqueue rate", results);
+    hipStreamSynchronize(stream);
+    hipStreamDestroy(stream);
```
hthangirala: Cannot release resources as threads are not synchronized and perturbs other threads are still in the critical path timing enqueue operations. Since we're only allowing one test to run at a time, its okay to let it cleaned up at process exit.

samples/1_Utils/hipDispatchLatency/hipDispatchEnqueueRateMT.cpp
gargrahul: Thanks updated the code.

samples/1_Utils/hipDispatchLatency/hipDispatchEnqueueRateMT.cpp
gargrahul: Removed the cleanup parts

PR #1875, opened 2020-02-16
Let HIP-Clang inline all functions by default

yxsamliu:
This is a quick workaround to match HCC behavior for performance since inlining usually
results in more optimization opportunities therefore better performance.

We will fine tuning inline threashold later.


b-sumner:
@yxsamliu while I understand that we might want to minimize the impact of a switchover to hip-clang, nothing can eliminate that completely.  And I'd say this is a move in the wrong direction.  I think we should avoid this change and instead document this difference with hcc and explain our plans for later tuning.

yxsamliu:
Our priority is to reach the same performance as HCC. After that, we can remove this and tune inline threashold.

PR #1876, opened 2020-02-16
Remove deprecated HIP markers

gargrahul:


include/hip/hip_profile.h
```diff
@@ -1,45 +0,0 @@
...
-#warning "HIP Markers are deprecated and would be removed soon."
-#include <CXLActivityLogger.h>
-#define HIP_SCOPED_MARKER(markerName, group)                                                       \
-    amdtScopedMarker __scopedMarker(markerName, group, nullptr);
-#define HIP_BEGIN_MARKER(markerName, group) amdtBeginMarker(markerName, group, nullptr);
-#define HIP_END_MARKER() amdtEndMarker();
-#else
-#define HIP_SCOPED_MARKER(markerName, group)
-#define HIP_BEGIN_MARKER(markerName, group)
-#define HIP_END_MARKER()
```
mangupta: We might want to retain these 3 lines to avoid build breakage of any applications that were adding markers. Of course markers will no longer work, but I don't see the need to break their builds.

src/hip_hcc.cpp
```diff
@@ -2481,29 +2451,6 @@ bool ihipStream_t::locked_copy2DAsync(void* dst, const void* src, size_t width,
...
 
-//-------------------------------------------------------------------------------------------------
-//-------------------------------------------------------------------------------------------------
-// Profiler, really these should live elsewhere:
-hipError_t hipProfilerStart() {
-    HIP_INIT_API(hipProfilerStart);
-#if COMPILE_HIP_ATP_MARKER
-    amdtResumeProfiling(AMDT_ALL_PROFILING);
-#endif
-
```
mangupta: Again maybe we want to retain the API as a stub similar to hipCtxCreate so that applications don't need to be modified just to be able to build them again.

src/hip_hcc.cpp
```diff
@@ -2481,29 +2451,6 @@ bool ihipStream_t::locked_copy2DAsync(void* dst, const void* src, size_t width,
...
-    return ihipLogStatus(hipSuccess);
-};
-
-
-hipError_t hipProfilerStop() {
-    HIP_INIT_API(hipProfilerStop);
-#if COMPILE_HIP_ATP_MARKER
-    amdtStopProfiling(AMDT_ALL_PROFILING);
-#endif
```
mangupta: Ditto

PR #1879, opened 2020-02-18
Fix HIPRTC headers to export C style symbols

saleelk:
Remove hiprtcGetTypeName from the headers


include/hip/hcc_detail/hiprtc.h
```diff
@@ -41,29 +45,22 @@ enum hiprtcResult {
...
-hiprtcResult hiprtcVersion(int* major, int* minor) noexcept
-{   // TODO: NVRTC versioning is somewhat unclear.
-    if (!major || !minor) return HIPRTC_ERROR_INVALID_INPUT;
-
-    // TODO: this should be generic / set by the build infrastructure.
-    *major = 9;
-    *minor = 0;
 
-    return HIPRTC_SUCCESS;
-}
```
mangupta: We are missing the implementation for hiprtcVersion() in src/hiprtc.cpp. This implementation needs to move there.

saleelk: Done

PR #1885, opened 2020-02-19
Fix hipcc for extra -mllvm option

yxsamliu:


bin/hipcc
```diff
@@ -729,7 +729,9 @@ if ($HIP_PLATFORM eq "clang") {
     }
     if ($optArg ne "-O0") {
         $HIPCXXFLAGS .= " -mllvm -amdgpu-early-inline-all=true -mllvm -amdgpu-function-calls=false";
-        $HIPLDFLAGS .= " -mllvm -amdgpu-early-inline-all=true -mllvm -amdgpu-function-calls=false";
+        if ($needLDFLAGS and not $needCXXFLAGS) {
+            $HIPLDFLAGS .= " -mllvm -amdgpu-early-inline-all=true -mllvm -amdgpu-function-calls=false";
+        }
```
emankov: Shouldn't `$HIPCXXFLAGS` be populated under `if ($needCXXFLAGS)` as well?

yxsamliu: if ($needCXXFLAGS) is false, $HIPCXXFLAGS is not used, so it does not matter.

yxsamliu:
ping

PR #1891, opened 2020-02-22
Return success when 2D copy size = 0

gargrahul:


srinivamd:
Created this https://github.com/ROCm-Developer-Tools/HIP/pull/1896 for review that covers other cases as well just in case.

mangupta:
Closing this in favor of #1896

PR #1899, opened 2020-02-27
[dtests] __shfl_up and __shfl_down tests

cjatin:


tests/src/kernel/hipShflUpDownTest.cpp
```diff
@@ -0,0 +1,97 @@
...
+        hipFree(d_a);
+        failed("Shfl Up Sum did not match.");
+    }
+    hipFree(d_a);
+}
+int main() {
+    runTestShflUp<long>();
+    runTestShflUp<long long>();
+    runTestShflDown<long>();
+    runTestShflDown<long long>();
```
mangupta: Can we also enable the test for int and float as well for completeness sake?

cjatin: done

PR #1900, opened 2020-02-28
return hipSuccess when count is zero

srinivamd:
Check size/count of bytes for zero count and return hipSuccess immediately. Used ihipLogStatus for return wherever allowed.

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -674,6 +674,7 @@ hipError_t hipMalloc(void** ptr, size_t sizeBytes) {
     hipError_t hip_status = hipSuccess;
 
     if (sizeBytes == 0) {
+        if (ptr) *ptr = NULL;
         return ihipLogStatus(hipSuccess);
     }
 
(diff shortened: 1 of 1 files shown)


srinivamd:
@emankov @gargrahul @mangupta please review this updated PR.

psychocoderHPC:
I suggest to set the pointer for malloc operations to `nullptr` so that passing the pointer later to `hipFree` is not failing. Not sure if `hipFree` supports `nullptr` but should.

srinivamd:
> I suggest to set the pointer for malloc operations to `nullptr` so that passing the pointer later to `hipFree` is not failing. Not sure if `hipFree` supports `nullptr` but should.

Isn't `nullptr` also returned on error/failure to malloc?
Also, wasn't sure about modifying caller's arg as a side-effect for nothing done.

srinivamd:
@mangupta @gargrahul Could you please review and approve for merge it if it looks good?

PR #1903, opened 2020-03-02
Match Occupancy APIs syntax with CUDA

satyanveshd:
Occupancy APIs syntax is changed to match with CUDA.

This is an updated version of PR #1625


asalmanp:
@satyanveshd can you please re-baseline and fix the merge conflict? 

After fixing the merge conflict, please make sure to change "uint32_t" to "int" in the two below locations so that when this PR gets merged it wont break HIP building

https://github.com/ROCm-Developer-Tools/HIP/blob/master/src/hip_module.cpp#L547
https://github.com/ROCm-Developer-Tools/HIP/blob/master/src/hip_module.cpp#L689

mangupta:
Closing this in favor of #1943

PR #1907, opened 2020-03-03
disable device functions malloc and free

satyanveshd:
disabling device functions __hip_malloc and __hip_free  for now as it is not thread safe. Multiple work items accessing these concurrently might cause race condition.


gargrahul:
Closing this in favor of #2009

PR #1908, opened 2020-03-03
[HIP] add hip specific properties for cooperative kernel multi device

asalmanp:


include/hip/hip_runtime_api.h
```diff
@@ -313,6 +321,14 @@ typedef enum hipDeviceAttribute_t {
...
     hipDeviceAttributeCooperativeLaunch,                 ///< Support cooperative launch
     hipDeviceAttributeCooperativeMultiDeviceLaunch,      ///< Support cooperative launch on multiple devices
+    hipDeviceAttributeCooperativeMultiDeviceUnmatchedFunc, ///< Supports cooperative launch on multiple
+                                                           ///devices with unmatched functions
+    hipDeviceAttributeCooperativeMultiDeviceUnmatchedGridDim, ///< Supports cooperative launch on multiple
+                                                              ///devices with unmatched grid dimensions
+    hipDeviceAttributeCooperativeMultiDeviceUnmatchedBlockDim,///< Supports cooperative launch on multiple
+                                                              ///devices with unmatched block dimensions
+    hipDeviceAttributeCooperativeMultiDeviceUnmatchedSharedMem,///< Supports cooperative launch on multiple
+                                                               ///devices with unmatched shared memories
```
srinivamd: Nit: Would it be ok to add/extend enums at the end as a practice to maintain compatibility? Same with extending structs above. Just a general comment/practice that might help unless there is a reason/expectation to bunch up related ones.
Versioning the structs (internal) could be good too for extensibility...

emankov: Additionally, I'd appreciate entire aligning the touched structs with NVIDIA's cause they are already incompatible. Thank's in advance.

asalmanp: moved the new enums/ peoperties to the end as requested.
@emankov entire aligning the structs/enums with the CUDA's is out of scope of this PR and it can be done in a separate PR. The purpose of this PR is just to add four new properties/enums for the cooperative groups feature.

emankov: Could you bring the rest into the line, please? Of course, in a separate PR.

PR #1909, opened 2020-03-04
Expose support for non-returning atomic FADD

mangupta:
- Implements SWDEV-215624

### DIFF
--- a/include/hip/hcc_detail/device_library_decls.h
+++ b/include/hip/hcc_detail/device_library_decls.h
@@ -72,6 +72,7 @@ extern "C" __device__ __attribute__((const)) uint __ockl_multi_grid_thread_rank(
 extern "C" __device__ __attribute__((const)) int __ockl_multi_grid_is_valid(void);
 extern "C" __device__ __attribute__((convergent)) void __ockl_multi_grid_sync(void);
 
+extern "C" __device__ void __ockl_global_atomic_add_f32(__attribute__((address_space(1))) float*, float);
 
 // Introduce local address space
 #define __local __attribute__((address_space(3)))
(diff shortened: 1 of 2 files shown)


include/hip/hcc_detail/device_library_decls.h
```diff
@@ -72,6 +72,7 @@ extern "C" __device__ __attribute__((const)) uint __ockl_multi_grid_thread_rank(
 extern "C" __device__ __attribute__((const)) int __ockl_multi_grid_is_valid(void);
 extern "C" __device__ __attribute__((convergent)) void __ockl_multi_grid_sync(void);
 
+extern "C" __device__ void __ockl_global_atomic_add_f32(float*, float);
```
b-sumner: The OCKL function requires an address_space(1) pointer for the first argument.

include/hip/hcc_detail/hip_atomic.h
```diff
@@ -73,6 +73,14 @@ float atomicAdd(float* address, float val)
 
     return __uint_as_float(r);
 }
+
+__device__
+inline
+void atomicAddNoRet(float* address, float val)
+{
+    __ockl_global_atomic_add_f32(address, val);
```
b-sumner: The OCKL function requires an address_space(1) pointer for the first argument.  Will need a cast if address here.

include/hip/hcc_detail/device_library_decls.h
mangupta: fixed via b87c508

include/hip/hcc_detail/hip_atomic.h
mangupta: fixed via b87c508

Issue #1205, opened 2019-07-05
hipStreamAddCallback, stream not blocked by callback (HIP/HCC)


tdd11235813:
Refering to this [issue](https://github.com/ComputationalRadiationPhysics/alpaka/issues/794) in Alpaka, it seems that [hipStreamAddCallback](https://rocm-documentation.readthedocs.io/en/latest/ROCm_API_References/HIP_API/Stream-Management.html#hipstreamaddcallback) callback execution does not block the hip/hcc stream, so when two callbacks are added to the same stream by two different threads, we get a race condition.
```c++
// stream is initialized like this (only disables automatic default-stream sync)
ALPAKA_HIP_RT_CHECK(           
    hipStreamCreateWithFlags(  
        &m_HipQueue,           
        hipStreamNonBlocking));
```
[callback includes a wait](https://github.com/ComputationalRadiationPhysics/alpaka/blob/6f43249c216a5cd9c041b8f8be59c6f220298905/include/alpaka/queue/QueueHipRtBlocking.hpp#L225), which however does not block the hip stream, thus our task execution becomes thread-unsafe (and our tests fail).

```c++
static void HIPRT_CB hipRtCallback(hipStream_t /*queue*/, hipError_t /*status*/, void *arg)                           
{                                                                                                                     
   /* ... */                                                                                                                 
    // Wait for the executing thread to finish the task if it has not already finished.                               
    std::unique_lock<std::mutex> lock(pCallbackSynchronizationData->m_mutex);                                         
    if(pCallbackSynchronizationData->state != CallbackState::finished)                                                
    {                                                                                                                 
        pCallbackSynchronizationData->m_event.wait( /* ... */ );
    }                                                                                                                 
}                                                                                                                     
```
```c++
// call is
ALPAKA_HIP_RT_CHECK(hipStreamAddCallback(
                    queue.m_spQueueImpl->m_HipQueue,
                    hipRtCallback,
                    pCallbackSynchronizationData.get(),
                    0u));
```
it is either a bug in the HIP implementation or a documentation error see
https://rocm-documentation.readthedocs.io/en/latest/ROCm_API_References/HIP_API/Stream-Management.html#hipstreamaddcallback: `The callback will block later work in the stream until it is finished.`

tdd11235813:
attached you find a minimal working example without Alpaka dependency. Issues reproduced with hip-clang on a Radeon VII (gfx906, Vega 20) GPU.
> (hipcc)
[...]

PR #1915, opened 2020-03-06
fix hipStreamAddCallback race

jeffdaily:


srinivamd:
Does this address this issue https://github.com/ROCm-Developer-Tools/HIP/issues/1205 also?

jeffdaily:
I wasn't aware of #1205.  I will check.  I know that this PR does fix a few TensorFlow unit tests, and that's why it was written.

jeffdaily:
I have 1 assert left from the test code attached to #1205 .
```
===============================================================================
test cases: 10 |  9 passed | 1 failed
assertions: 20 | 19 passed | 1 failed
```

The last failure might be due to HCC.  This PR is the first that I am aware of that uses `accelerator_view::acquire_locked_hsa_queue` to insert foreign AQL packets.  This is causing some failures in HCC.

srinivamd:
@jeffdaily Your results, with just 1 assertion failed, is awesome! Will start reviewing the code.

src/hip_hcc.cpp
```diff
@@ -306,17 +318,21 @@ inline void ihipStream_t::locked_wait(bool& waited) {
     // create a marker while holding stream lock,
     // but release lock prior to waiting on the marker
     hc::completion_future marker;
+    std::vector<hsa_signal_t> pending_callbacks;
```
srinivamd: initialize it `{}` just to be safe?

src/hip_stream.cpp
```diff
@@ -257,11 +257,85 @@ hipError_t hipStreamGetPriority(hipStream_t stream, int* priority) {
...
+    // stream is locked, now lock underlying HSA queue using unique_ptr RAII
+    auto av{cs->_av};
+    auto d{[=](hsa_queue_t*) mutable { av.release_locked_hsa_queue(); }};
+    std::unique_ptr<hsa_queue_t, decltype(d)> q{
+        static_cast<hsa_queue_t*>(av.acquire_locked_hsa_queue()),
+        std::move(d)};
+
+    // get first packet
+    auto b_index{hsa_queue_add_write_index_relaxed(q.get(), 1) % q->size};
+    auto b{static_cast<hsa_barrier_or_packet_t*>(q->base_address) + b_index};
```
srinivamd: nit: not sure about single letter variable identifiers 'b' and 'd'. Maybe a more meaningful/longer identifier, making it easier to spot it in code, improve readability?

jeffdaily:
@srinivamd I appreciate your review.  I would make the suggested changes, but we need to have more discussion with HCC team concerning this PR.

jeffdaily:
Closing in favor of #1921 .  This current PR cannot be accepted since it breaks a contract with HCC's runtime since HCC cannot track dependencies of foreign AQL packets.

PR #1916, opened 2020-03-06
[HIP] Refactor cooperative APIs

asalmanp:
This PR refactors cooperative APIs to use the dedicated cooperative accelerated view introduced in PR#1406 (https://github.com/RadeonOpenCompute/hcc/pull/1406) in HCC.

In addition to the above change, the current PR addresses the following issues found during internal testing:

- added support for hipLaunchCooperativeKernel() to be able to use the NULL/default stream

- added support to return hipErrorCooperativeLaunchTooLarge if requested workload cannot fit on the GPU

- added support to properly return error where two (or more) of the GPUs are the same device (and same stream) in hipLaunchCooperativeKernelMultiDevice() API

- Fixed an issue where a floating point exception occurs if we pass "0" into blockDim in multi-GPU cooperative launch

- Fixed an issue where a multi-GPU launch segfaults if the final argument passed to the kernel is a 4 bytes value

- Fixed an issue where launching a multi-device kernel to the NULL stream causes a segfault

### DIFF
--- a/src/hip_hcc.cpp
+++ b/src/hip_hcc.cpp
@@ -1619,7 +1619,7 @@ void ihipPrintKernelLaunch(const char* kernelName, const grid_launch_parm* lp,
 // Allows runtime to track some information about the stream.
 hipStream_t ihipPreLaunchKernel(hipStream_t stream, dim3 grid, dim3 block, grid_launch_parm* lp,
                                 const char* kernelNameStr, bool lockAcquired) {
-    if (stream == nullptr || stream != stream->getCtx()->_defaultStream){
+    if (stream == nullptr || stream != stream->getCtx()->_defaultStream) {
         stream = ihipSyncAndResolveStream(stream, lockAcquired);
     }
     lp->grid_dim.x = grid.x;
--- a/src/hip_hcc.cpp
+++ b/src/hip_hcc.cpp
@@ -1619,7 +1619,9 @@ void ihipPrintKernelLaunch(const char* kernelName, const grid_launch_parm* lp,
 // Allows runtime to track some information about the stream.
 hipStream_t ihipPreLaunchKernel(hipStream_t stream, dim3 grid, dim3 block, grid_launch_parm* lp,
                                 const char* kernelNameStr, bool lockAcquired) {
-    stream = ihipSyncAndResolveStream(stream, lockAcquired);
+    if (stream == nullptr || stream != stream->getCtx()->_defaultStream){
+        stream = ihipSyncAndResolveStream(stream, lockAcquired);
+    }
     lp->grid_dim.x = grid.x;
     lp->grid_dim.y = grid.y;
     lp->grid_dim.z = grid.z;
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -502,7 +502,8 @@ hipError_t ihipLaunchCooperativeKernel(const void* f, dim3 gridDim,
 
     stream = ihipSyncAndResolveStream(stream);
 
-    if (!stream->getDevice()->_props.cooperativeLaunch) {
+    if (!stream->getDevice()->_props.cooperativeLaunch ||
+        blockDimX.x * blockDimX.y * blockDimX.z > stream->getDevice()->_props.maxThreadsPerBlock) {
         return hipErrorInvalidConfiguration;
     }
 
@@ -650,7 +651,8 @@ hipError_t ihipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsL
         }
 
         if (lp.gridDim.x == 0 || lp.gridDim.y == 0 || lp.gridDim.z == 0 ||
-            lp.blockDim.x == 0 || lp.blockDim.y == 0 || lp.blockDim.z == 0){
+            lp.blockDim.x == 0 || lp.blockDim.y == 0 || lp.blockDim.z == 0 ||
+            lp.blockDim.x * lp.blockDim.y * lp.blockDim.z > currentDevice->_props.maxThreadsPerBlock){
             return hipErrorInvalidConfiguration;
         }
     }
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -192,8 +192,8 @@ hipError_t ihipModuleLaunchKernel(TlsData *tls, hipFunction_t f, uint32_t global
         if (impCoopParams) {
             const auto p{static_cast<const char*>(*impCoopParams)};
             // The sixth index is for multi-grid synchronization
-            kernargs.insert((kernargs.cend() - HIP_IMPLICIT_KERNARG_SIZE) + 6 * HIP_IMPLICIT_KERNARG_ALIGNMENT,
-                            p, p + HIP_IMPLICIT_KERNARG_ALIGNMENT);
+            copy(p, p + HIP_IMPLICIT_KERNARG_ALIGNMENT,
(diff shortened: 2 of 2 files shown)


src/hip_module.cpp
```diff
@@ -459,89 +542,133 @@ hipError_t ihipLaunchCooperativeKernel(const void* f, dim3 gridDim,
...
+    GET_TLS();
+    uint32_t numBlocksPerSm;
+    result = ihipOccupancyMaxActiveBlocksPerMultiprocessor(tls, &numBlocksPerSm, kd,
+                    stream->getDevice()->_props.warpSize, sharedMemBytes);
+    if (result != hipSuccess) {
+        return hipErrorLaunchFailure;
+    }
+    int maxActiveBlocks = numBlocksPerSm * stream->getDevice()->_props.multiProcessorCount;
+
+    //check to see if the workload fits on the GPU
```
emankov: Formatting: missing space before first annotating word.

src/hip_module.cpp
```diff
@@ -459,89 +542,133 @@ hipError_t ihipLaunchCooperativeKernel(const void* f, dim3 gridDim,
...
+
+        if (lp.stream == nullptr){
+            return hipErrorInvalidResourceHandle;
         }
-    }
 
-    hipFunction_t* gwsKds = reinterpret_cast<hipFunction_t*>(malloc(sizeof(hipFunction_t) * numDevices));
-    hipFunction_t* kds    = reinterpret_cast<hipFunction_t*>(malloc(sizeof(hipFunction_t) * numDevices));
+        auto it = find(streams.begin(), streams.end(), lp.stream);
+        if (it == streams.end()){
```
emankov: Here and below iterator is excessive, `if (find(...) == steams. end()) `.

src/hip_module.cpp
```diff
@@ -459,89 +542,133 @@ hipError_t ihipLaunchCooperativeKernel(const void* f, dim3 gridDim,
     kd->_kernarg_layout = *reinterpret_cast<const std::vector<
             std::pair<std::size_t, std::size_t>>*>(kargs.getHandle());
 
+    GET_TLS();
+    uint32_t numBlocksPerSm;
+    result = ihipOccupancyMaxActiveBlocksPerMultiprocessor(tls, &numBlocksPerSm, kd,
```
emankov: Uninitialized `numBlocksPerSm`.

src/hip_module.cpp
```diff
@@ -459,89 +542,133 @@ hipError_t ihipLaunchCooperativeKernel(const void* f, dim3 gridDim,
...
 
     stream->criticalData().unlock();
-#if (__hcc_workweek__ >= 19213)
     stream->criticalData()._av.release_locked_hsa_queue();
-#endif
+    coopAV.release_locked_hsa_queue();
+
+    // wait on the dispatch on the dedicated cooperative queue to finish
+    coopAV.wait(hc::hcWaitModeActive);
+
```
emankov: Formatting: too many empty lines.

src/hip_module.cpp
```diff
@@ -550,23 +677,42 @@ hipError_t ihipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsL
...
-            free(kds);
             return hipErrorInvalidValue;
         }
         hip_impl::kernargs_size_align kargs = ps.get_kernargs_size_align(
                 reinterpret_cast<std::uintptr_t>(lp.func));
         kds[i]->_kernarg_layout = *reinterpret_cast<const std::vector<std::pair<std::size_t, std::size_t>>*>(
                 kargs.getHandle());
+
+        uint32_t numBlocksPerSm;
+        result = ihipOccupancyMaxActiveBlocksPerMultiprocessor(tls, &numBlocksPerSm, kds[i],
```
emankov: Uninitialized `numBlocksPerSm`.

src/hip_module.cpp
```diff
@@ -192,7 +192,7 @@ hipError_t ihipModuleLaunchKernel(TlsData *tls, hipFunction_t f, uint32_t global
         if (impCoopParams) {
             const auto p{static_cast<const char*>(*impCoopParams)};
             // The sixth index is for multi-grid synchronization
-            kernargs.insert((kernargs.cend() - padSize - HIP_IMPLICIT_KERNARG_SIZE) + 6 * HIP_IMPLICIT_KERNARG_ALIGNMENT,
+            kernargs.insert((kernargs.cend() - HIP_IMPLICIT_KERNARG_SIZE) + 6 * HIP_IMPLICIT_KERNARG_ALIGNMENT,
```
scchan: - Using `HIP_IMPLICIT_KERNARG_ALIGNMENT` is odd here.  I think what we really want to say is something like `6 * HIP_IMPLICIT_KERNARG_ELEMENT_SIZE` (even though the alignment and size is the same)?
- `insert` add new elements to the vector and therefore increases its size.  I think you want to `std::copy` into the 6th slot instead?

src/hip_hcc.cpp
```diff
@@ -1619,7 +1619,9 @@ void ihipPrintKernelLaunch(const char* kernelName, const grid_launch_parm* lp,
 // Allows runtime to track some information about the stream.
 hipStream_t ihipPreLaunchKernel(hipStream_t stream, dim3 grid, dim3 block, grid_launch_parm* lp,
                                 const char* kernelNameStr, bool lockAcquired) {
-    stream = ihipSyncAndResolveStream(stream, lockAcquired);
+    if (stream == nullptr || stream != stream->getCtx()->_defaultStream){
+        stream = ihipSyncAndResolveStream(stream, lockAcquired);
```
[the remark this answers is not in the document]
scchan: Could you clarify what issue these additional checks try to address?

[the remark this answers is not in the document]
asalmanp: @scchan The ihipSyncAndResolveStream syncs a normal stream with a defaultStream and there are two cases:
1 - if a nullStream is passed, this function syncs with default stream and returns the defaultStream
2 - if a normal stream is passed, this function syncs the stream with defautlStream and returns the same normal stream.

The new additional checks make sure that only under the above two cases the ihipSyncAndResolveStream is called inside the ihipPreLaunchKernel. So, for example, if the defaultStream is passed, then it is not needed to call ihipSyncAndResolveStream here. 

This is the case where a nullStream is passed to hipLaunchCooperativeKernel API. In this case, there is already an ihipSyncAndResolveStream call inside the hipLaunchCooperativeKernel API to get the deafultStream so when we reached here in ihipPreLaunchKernel, it is not required to call ihipSyncAndResolveStream because we already have deafultStream and it is not needed to sync again with deafultStream itself.

src/hip_module.cpp
asalmanp: fixed.

src/hip_module.cpp
asalmanp: corrected.

src/hip_module.cpp
asalmanp: initialized numBlocksPerSm.

src/hip_module.cpp
asalmanp: @scchan I removed insert with copy as requested.
Regarding using HIP_IMPLICIT_KERNARG_ELEMENT_SIZE, it is not defined anywhere in HIP.. The byte size and alignment for implicit arguments are the same, so using the  HIP_IMPLICIT_KERNARG_ALIGNMENT should be fine here.

PR #1918, opened 2020-03-08
Packaging changes to bring them to sync with generated ones

cjatin:


packaging/hip-targets-release.cmake
```diff
@@ -35,12 +40,18 @@ set_target_properties(hip::hip_hcc PROPERTIES
...
-else()
+elseif(HIP_COMPILER STREQUAL "hcc")
 set_target_properties(hip::hip_hcc PROPERTIES
   IMPORTED_LINK_INTERFACE_LIBRARIES_RELEASE "hcc::hccrt;hcc::hc_am"
   IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/lib/libhip_hcc.so"
   IMPORTED_SONAME_RELEASE "libhip_hcc.so"
   )
+else()
+set_target_properties(hip::hip_hcc PROPERTIES
+  IMPORTED_LINK_INTERFACE_LIBRARIES_RELEASE "hcc::hccrt"
```
pfultz2: This should not be included. In fact, this should be the same for hip-clang and hcc.

packaging/hip-targets-release.cmake
```diff
@@ -17,12 +17,17 @@ set_target_properties(hip::hip_hcc_static PROPERTIES
   IMPORTED_LINK_INTERFACE_LANGUAGES_RELEASE "CXX"
   IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/lib/libhip_hcc_static.a"
   )
-else()
+elseif(HIP_COMPILER STREQUAL "hcc")
 set_target_properties(hip::hip_hcc_static PROPERTIES
   IMPORTED_LINK_INTERFACE_LANGUAGES_RELEASE "CXX"
   IMPORTED_LINK_INTERFACE_LIBRARIES_RELEASE "hc_am"
```
pfultz2: This should be removed as well. It should be the same between hip-clang and hcc.

PR #1919, opened 2020-03-09
separate printf declaration for vdi/clang

ssahasra:
There are now two implementations of printf in HIP:

1. The implemenation for HCC is controlled by the HC_FEATURE_PRINTF
   macro, and it works only with the HCC compiler used in combination
   with the HCC runtime.

2. The implementation for hip-clang requires the VDI runtime, and is
   always enabled with that combination.


ssahasra:
Can someone please merge this for me? I don't have the necessary permission.

PR #1921, opened 2020-03-09
fix hipStreamAddCallback race, using new HCC API

jeffdaily:
Depends on https://github.com/RadeonOpenCompute/hcc/pull/1420


jeffdaily:
@srinivamd After discussion with @scchan , my first attempt #1915 should be replaced with this PR.  I was using HCC incorrectly, circumventing HCC's dependency tracking et al.

Good news is that if RadeonOpenCompute/hcc#1420 is approved (which this current HIP PR depends on), I got 100% passing for #1205 .

srinivamd:
Nice! Thanks @jeffdaily for the update!

jeffdaily:
@srinivamd I found a solution that does not depend on HCC adding an API.  Closing in favor of that solution in #1934 .

PR #1924, opened 2020-03-10
Let hipcc not pass -mllvm option to HIP-Clang on Windows

yxsamliu:
Currently there is a clang bug on Windows causing duplicate -mllvm options in clang -cc1.

Tempoarily disable -mllvm options for HIP-Clang on Windows until the bug is fixed.

Change-Id: I3a4393ba7745989398dc6c6001722837dad18704


yxsamliu:
The link for the clang fix is https://reviews.llvm.org/D76032

It will take some time so we still need this workaround.

PR #1925, opened 2020-03-10
Fix incorrect shfl_xor for Windows

arghdos:
copy/paste error, need __shfl_xor w/ lane_mask


emankov:
Duplicate of #1923

dmcdougall:
Hmmm.  Those two patches are not the same.  Is there a fundamental difference between `_down` and `_xor`?

arghdos:
Good catch @dmcdougall -- @emankov, I left a review for the fix

emankov:
2 copy-paste typos )

PR #1928, opened 2020-03-11
Fix hipOccupancyMaxActiveBlocksPerMultiprocessor API on NVCC

jlgreathouse:
NVCC warned if you tried to use hipOccupancyMaxActiveBlocksPerMultiprocessor
because when passing in a device function pointer, "const void* func" was
insufficient to describe it accurately. Changing to a templated class type
definition to fix this.


jlgreathouse:
Oops, I hadn't sufficiently tested this on NVCC. This fails due to putting C++ in the extern C section. I'll submit a fixed PR soon.

PR #1929, opened 2020-03-11
[HIP] add cooperative kernel launch APIs on NVCC

asalmanp:


jlgreathouse:
I was chatting with Aryan about this: testing on one of my NV boxes shows that (I think) this is missing a `#define hipLaunchParams cudaLaunchParams` somewhere. Compilation with NVCC would fail with this patch applied to my HIP version because hipLaunchParams was undefined. But I'll admit my NV box has a hacked up ROCm 3.1, so I may be missing another patch somewhere.

include/hip/nvcc_detail/hip_runtime_api.h
```diff
@@ -1679,6 +1686,17 @@ inline static hipError_t hipGetChannelDesc(hipChannelFormatDesc* desc, hipArray_
     return hipCUDAErrorTohipError(cudaGetChannelDesc(desc,array));
 }
 
+inline static hipError_t hipLaunchCooperativeKernel(const void* f, dim3 gridDim, dim3 blockDimX,
```
jlgreathouse: Really minor comment, but should the blockDimX variable instead be named blockDim? The X seems to be a copy/paste from another function.

asalmanp: Agree, rename it.

PR #1935, opened 2020-03-11
Fix cooperative launch APIs to set hipGetLastError

jlgreathouse:
Previously, the cooperative launch APIs did not properly log their
errors in the global hipGetLastError variable before returning back
to the user. As such, the APIs would leave hipSuccess in the
last error, which would break some use cases.

This fixes that problem by making a trampoline function that does
the HIP_INIT_API and ihipLogStatus.

### DIFF
--- a/include/hip/hcc_detail/functional_grid_launch.hpp
+++ b/include/hip/hcc_detail/functional_grid_launch.hpp
@@ -37,14 +37,15 @@ THE SOFTWARE.
 hipError_t ihipExtLaunchMultiKernelMultiDevice(hipLaunchParams* launchParamsList, int numDevices,
                                                unsigned int flags, hip_impl::program_state& ps);
 
-hipError_t ihipLaunchCooperativeKernel(const void* f, dim3 gridDim, dim3 blockDimX, void** kernelParams,
-                unsigned int sharedMemBytes, hipStream_t stream, hip_impl::program_state& ps);
+hipError_t hipLaunchCooperativeKernel(const void* f, dim3 gridDim,
+                                    dim3 blockDim, void** args,
+                                    size_t sharedMem, hipStream_t stream,
+                                    hip_impl::program_state& ps);
 
-hipError_t ihipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsList, int  numDevices,
-                unsigned int  flags, hip_impl::program_state& ps);
-
-
-
+hipError_t hipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsList,
+                                                 int  numDevices,
+                                                 unsigned int flags,
+                                                 hip_impl::program_state& ps);
 
 #pragma GCC visibility push(hidden)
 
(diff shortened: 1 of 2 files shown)


include/hip/hcc_detail/functional_grid_launch.hpp
```diff
@@ -37,14 +37,15 @@ THE SOFTWARE.
...
-                unsigned int sharedMemBytes, hipStream_t stream, hip_impl::program_state& ps);
-
-hipError_t ihipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsList, int  numDevices,
-                unsigned int  flags, hip_impl::program_state& ps);
-
-
+hipError_t hipLaunchCooperativeKernel(const void* f, dim3 gridDim,
+                                    dim3 blockDim, void** args,
+                                    size_t sharedMem, hipStream_t stream,
+                                    hip_impl::program_state& ps);
```
SarbojitAMD: I think we should remove **hip_impl::program_state& ps** in order to match it with CUDA API signature

include/hip/hcc_detail/functional_grid_launch.hpp
```diff
@@ -37,14 +37,15 @@ THE SOFTWARE.
...
 
+hipError_t hipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsList,
+                                                 int  numDevices,
+                                                 unsigned int flags,
+                                                 hip_impl::program_state& ps);
```
SarbojitAMD: I think we should remove hip_impl::program_state& ps in order to match it with CUDA API signature

include/hip/hcc_detail/functional_grid_launch.hpp
```diff
@@ -202,22 +203,24 @@ hipError_t hipExtLaunchMultiKernelMultiDevice(hipLaunchParams* launchParamsList,
...
 inline
 __attribute__((visibility("hidden")))
-hipError_t hipLaunchCooperativeKernel(F f, dim3 gridDim, dim3 blockDimX, void** kernelParams,
-                unsigned int sharedMemBytes, hipStream_t stream) {
-
+hipError_t hipLaunchCooperativeKernel(F f, dim3 gridDim, dim3 blockDim,
+                                      void** args, size_t sharedMem,
+                                      hipStream_t stream) {
     hip_impl::hip_init();
     auto& ps = hip_impl::get_program_state();
```
SarbojitAMD: we can always get program_state in internal functions do we really need to get it in here?

src/hip_module.cpp
```diff
@@ -698,6 +711,21 @@ hipError_t ihipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsL
...
 
+__attribute__((visibility("default")))
+hipError_t hipLaunchCooperativeKernelMultiDevice(hipLaunchParams* launchParamsList,
+                                                 int  numDevices,
+                                                 unsigned int flags,
+                                                 hip_impl::program_state& ps) {
+
+    // Skipping passing in ps, because the logging function does not like it
+    HIP_INIT_API(hipLaunchCooperativeKernelMultiDevice, launchParamsList,
+                 numDevices);
```
SarbojitAMD: 'falgs' is not logged

include/hip/hcc_detail/functional_grid_launch.hpp
asalmanp: @SarbojitAMD the hipLaunchCooperativeKernel here is an overloaded internal function which is used to call the internal  ihipLaunchCooperativeKernel function. The hip_impl::program_state& ps is needed here and should not be removed.

include/hip/hcc_detail/functional_grid_launch.hpp
asalmanp: @SarbojitAMD  same as above the hipLaunchCooperativeKernelMultiDevice here is an overloaded internal function which is used to call the internal  ihipLaunchCooperativeKernelMultiDevice function. The hip_impl::program_state& ps is needed here and should not be removed.

include/hip/hcc_detail/functional_grid_launch.hpp
asalmanp: @SarbojitAMD Yes, the hip_impl::get_program_state() is intentionally used here to get the correct program state during lazy binding. Please see [PR#1698](https://github.com/ROCm-Developer-Tools/HIP/pull/1698) for more details.

include/hip/hcc_detail/functional_grid_launch.hpp
jlgreathouse: Agreed with @asalmanp for these. These overloaded functions exist as trampolines to reach the internal cooperative launch functions. ihipLogStatus() explicitly says we should not log the status of internal functions (so don't use it in ihip* locations), so these functions exist as a safe place to do HIP_INIT() and ihipLogStatus() before diving into the ihip cooperative launch functions.

As Aryan notes, `ps` is needed inside the ihip cooperative launch function, so we're passing it through here.

The actual user-visible hipLaunchCooperativeKernel() without `hip_impl::program_state& ps` still exists.

src/hip_module.cpp
jlgreathouse: Good catch. I'll submit a fix.

PR #1937, opened 2020-03-12
Fix segfault seen in hipMemcpy [SWDEV-224941]

SarbojitAMD:
Fix for segfault seen if invalid kind is passed to hipMemcpy.

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -324,6 +324,9 @@ void memcpy_impl(void* __restrict dst, const void* __restrict src, size_t n,
     auto di{info(dst)};
 
     if (!is_large_BAR){
+       // Pointer info takes presidence over hipMemcpyKind
+       // if there is mismatch b/w Memcpy kind and dst/src pointer
+       // E.g. dst(host pointer),src(device pointer) and hipMemcpyKind set as hipMemcpyHostToDevice
        if (di.size == is_cpu_owned && si.size == is_cpu_owned)
           k = hipMemcpyHostToHost;
        else if (si.size == is_cpu_owned && di.size != is_cpu_owned)
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -317,9 +317,38 @@ void generic_copy(void* __restrict dst, const void* __restrict src, size_t n,
     return do_copy(dst, src, n, di.agentOwner, si.agentOwner);
 }
 
+inline void validateCpyRequest(const void* dst, const void* src, const hipMemcpyKind k)
+{
+    if(is_large_BAR) return;
+
+    const auto di{info(dst)};
+    const auto si{info(src)};
+    switch(k){
+    case hipMemcpyHostToHost:
+         if (si.size != is_cpu_owned || di.size != is_cpu_owned)
+            throw ihipException(hipErrorInvalidValue);
+         break;
+    case hipMemcpyHostToDevice:
+         if (si.size != is_cpu_owned)
+            throw ihipException(hipErrorInvalidValue);
+         break;
+    case hipMemcpyDeviceToHost:
+         if (di.size != is_cpu_owned)
+            throw ihipException(hipErrorInvalidValue);
+         break;
+    case hipMemcpyDeviceToDevice:
+         if (si.size == is_cpu_owned || di.size == is_cpu_owned)
+            throw ihipException(hipErrorInvalidValue);
+         break;
+    default:
+         throw ihipException(hipErrorInvalidValue);
+    }
+}
+
 inline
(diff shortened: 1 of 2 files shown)


AlexVlx:
This adds unnecessary overhead, what is the point, beyond "somebody filed a bug". I don't believe CUDA goes to similar lengths to guard against user error.

SarbojitAMD:
> This adds unnecessary overhead, what is the point, beyond "somebody filed a bug". I don't believe CUDA goes to similar lengths to guard against user error.

I have verified and found CUDA returns error if invalid kind passed but HIP segfaults which is not a good user experience. 
Regarding overhead we can always check if the overhead is significant enough of causing any perf regression.

AlexVlx:
What do you define as invalid kind? Have you actually measured (we can do many things)? Why are you duplicating the switch, and expensive pointer info checks? If you really want to do a thing, just changing the default case to throw an "Invalid copy kind" / or that gnarly ihipException thing we have for no good reason should be more than enough.

SarbojitAMD:
> What do you define as invalid kind? Have you actually measured (we can do many things)? Why are you duplicating the switch, and expensive pointer info checks? If you really want to do a thing, just changing the default case to throw an "Invalid copy kind" / or that gnarly ihipException thing we have for no good reason should be more than enough.

I think error scenario is not very clear to you hence all these doubts, I added test cases along the change to make it more clear but let me try to explain it here again

Segfault happens if user calls hipMemcpy( device_ptr, host_ptr, size, **hipMemcpyDeviceToHost**); // notice src is a host pointer and dst is a device pointer but kind passed as D2H which will fail if system is not a large BAR

Same segfaults seen in other cases where 
i) src = dev_ptr, dst=host_ptr, kind=hipMemcpyHostToDevice 
....
CUDA can catch such scenario and returns invalid value.

src/hip_memory.cpp
```diff
@@ -320,20 +320,31 @@ void generic_copy(void* __restrict dst, const void* __restrict src, size_t n,
 inline
 void memcpy_impl(void* __restrict dst, const void* __restrict src, size_t n,
                  hipMemcpyKind k) {
+    auto si{info(src)};
+    auto di{info(dst)};
+
+    if (!is_large_BAR){
```
mangupta: Maybe add a comment as to why we are doing this.. something like direction determined by src/dst pointer info takes precedence over hipMemcpyKind if there is a mismatch in input parameters passed to the API.
Also why is this applicable only for small BAR systems?

SarbojitAMD: It does not fail for large BAR hence the check. Mostly failure happens because there are cases where lower functions uses std::memcpy without checking if pointers are host pointer.

src/hip_memory.cpp
```diff
@@ -320,20 +320,34 @@ void generic_copy(void* __restrict dst, const void* __restrict src, size_t n,
...
+       // Pointer info takes presidence over hipMemcpyKind
+       // if there is mismatch b/w Memcpy kind and dst/src pointer
+       // E.g. dst(host pointer),src(device pointer) and hipMemcpyKind set as hipMemcpyHostToDevice
+       if (di.size == is_cpu_owned && si.size == is_cpu_owned)
```
arghdos: Shouldn't these checks be lowered into the `h2d_copy`, `d2h_copy` and `generic_copy`'s below?  We are introducing extra overhead for no reason by running `info` and doing the checks here. 

For instance, once we've committed to a `hipMemcpyHostToDevice`, we are already determining the source / destination pointer information (e.g., on line 267).  Why not put the statement:

```
si.size == is_cpu_owned && di.size != is_cpu_owned
```

 in `h2d_copy`, and return a `hipInvalidValue` or some other error code at that point if it fails?

SarbojitAMD:
> This is not the desired behavior. If the user supplies an incorrect pointer / hipMemcpyKind combination, this should return an error code that they can check. We should not be "fixing" it for them.
> 
> Returning an error code lets a user know that there's something wrong with their code, this method will hide it until they try to run on (e.g.,) the NVCC backend.

NVCC does not catch all scenarios e.g. In D2D case if memkind passed is H2H, D2H, .... API will return success hence we can't match exact NVCC behavior.

src/hip_memory.cpp
SarbojitAMD: For API hipMemcpyHostToDevice & hipMemcpyDeviceToHost it is fine but for H2H case we need to have extra info()s anyway. So if you see only one extra info() that is being added here.

AlexVlx:
@mangupta why did you merge this? @arghdos who is the original requester had issues with the fix, and I have issues with the overhead of needlessly invoking info.

PR #1939, opened 2020-03-12
handle C++ name mangling during symbol lookup

arghdos:
Workaround for SWDEV-225389 w/ HCC.

### DIFF
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -39,6 +39,7 @@ THE SOFTWARE.
 #include <cstdint>
 #include <cstdio>
 #include <cstdlib>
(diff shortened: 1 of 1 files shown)


SarbojitAMD:
Could you please create test case for the scenario that you have fixed here. 
Thanks

AlexVlx:
I don't quite think this is the right place to fix this, since this is an implementation detail. I suspect you might have some user-facing API that is not passing the right symbol name (probably because the user is not passing the right thing), but that should be addressed there and not in the implementation, which assumes stricter preconditions.

arghdos:
Closing re: email discussion

PR #1946, opened 2020-03-13
Improve Occupancy Apis and support Clang+Hcc RT

tomsang:
1.Rewrite the following interfaces to make logic clear.
  hipOccupancyMaxActiveBlocksPerMultiprocessor
  hipOccupancyMaxActiveBlocksPerMultiprocessorWithFlags
  hipOccupancyMaxPotentialBlockSize

2.Call ihipConvertHostFuncPtr2Symbol to support clang + HIP-HCC.

PR #1947, opened 2020-03-13
enable HCC printf when using hip-clang

ssahasra:
This allows printf to work with hip-clang and HCC runtime. See comments under #1919 for a reported bug and feature request.

### DIFF
--- a/include/hip/hcc_detail/device_functions.h
+++ b/include/hip/hcc_detail/device_functions.h
@@ -34,13 +34,19 @@ THE SOFTWARE.
 #include <hip/hcc_detail/device_library_decls.h>
 #include <hip/hcc_detail/llvm_intrinsics.h>
 
-#if __HIP_CLANG_ONLY__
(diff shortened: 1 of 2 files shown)


ssahasra:
@emankov, please wait for Tao Sang to confirm that the patch actually works with clang+HCC

gargrahul:
> 
> 
> @emankov, please wait for @tomsang to confirm that the patch actually works with clang+HCC

Tao has added part of this in #1946 - https://github.com/ROCm-Developer-Tools/HIP/pull/1946/files#diff-55dee707086ed9b76f9b1774f24f918bR41 , I think after this we won't need it there.

tomsang:
Compiling in Clang+Hcc RT can work now.

ssahasra:
@emankov this PR is ready to be merged if it has passed the usual process. @tomsang confirmed that the patch works for him with clang on HCC runtime.

PR #1964, opened 2020-03-20
Fix for Segfault seen in hipMemcpyDtoD [SWDEV-227444]

SarbojitAMD:
System where Peer2Peer access is not enabled hipMemcpyDtoD & hipMemcpyPeer throw exception. Added a fallback mechanism which will be slow but complete the copy.

### DIFF
--- a/src/hip_memory.cpp
+++ b/src/hip_memory.cpp
@@ -311,14 +311,12 @@ void generic_copy(void* __restrict dst, const void* __restrict src, size_t n,
 
     hsa_status_t res = hsa_amd_agents_allow_access(1u, &si.agentOwner,
                                                    nullptr, di.agentBaseAddress);
-    if (res != HSA_STATUS_SUCCESS){
-        // If devices do not have access then fallback mechanism will be used
-        // copy will be slower
-        throwing_result_check(hsa_memory_copy(dst,src,n), __FILE__, __func__, __LINE__);
-        return;
+    if (res == HSA_STATUS_SUCCESS){
+        return do_copy(dst, src, n, di.agentOwner, si.agentOwner);
     }
-
-    return do_copy(dst, src, n, di.agentOwner, si.agentOwner);
+    // If devices do not have access then fallback mechanism will be used
+    // copy will be slower
+    throwing_result_check(hsa_memory_copy(dst,src,n), __FILE__, __func__, __LINE__);
 }
 
 inline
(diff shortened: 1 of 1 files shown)


AlexVlx:
Exceptions are not for message passing, and this is needlessly slow and verbose. Again, it's unclear if HIP ever handled this, but if it did the right thing is to call `hsa_memory_copy` on failure, at the point of failure and not after unwinding the stack.

mangupta:
HIP used to fallback to unpinned memory copy if peers is not enabled in D2D case. However I agree with Alex that we should throw exception and catch to resolve this. It should be something like ```if (res == HSA_STATUS_SUCCESS) <do_copy...> else <fallback_code>```

jeffdaily:
I agree with @mangupta 's observation.  I went back through sync memcpy code history to verify the fallback path.  We used to call `stream->locked_copySync`, and in that call it would check `canSeeMemory`.  If memory wasn't visible, `forceUnpinnedCopy=true` would be passed to HCC's `accelerator_view::copy_ext` to force the use of the unpinned copy engine.  The unpinned copy engine would use two fixed-size, pinned staging buffers and cycle between them to get some overlap as it iterated over the copied memory.

https://github.com/RadeonOpenCompute/hcc/blob/96192de51d391c07bbeb38132790d867b3469540/lib/hsa/unpinned_copy_engine.cpp#L476-L564

jeffdaily:
Pardon, my history lesson wasn't needed.  I wasn't suggesting we copy/paste/reimplement the unpinned copy engine.  I agree with @mangupta 's suggestion.

In the sync copy path for D2D, when `hsa_amd_agents_allow_access` fails, the fallback should be called instead of throwing.  The fallback should be `stream->locked_copySync` like it was before. 

Need to better understand whether the async path needs the new catch block.

jeffdaily:
LGTM

src/hip_memory.cpp
```diff
@@ -341,6 +345,14 @@ void memcpy_impl(void* __restrict dst, const void* __restrict src, size_t n,
     case hipMemcpyHostToDevice: return h2d_copy(dst, src, n, di);
     case hipMemcpyDeviceToHost: return d2h_copy(dst, src, n, si);
     case hipMemcpyDeviceToDevice: {
+        hsa_status_t res = hsa_amd_agents_allow_access(1u, &si.agentOwner,
```
AlexVlx: This is not writtent well. Assume that the `hsa_amd_agents_allow_access` call succeeds. Then you're going to call it again immediately because... reasons. Try to write this in affirmative form, it'll prevent such issues i.e. if allow access succeeds return do_copy, and subsequently handle the special case. This is also a matter of logical hygiene (follow the likeliest path first).

PR #1969, opened 2020-03-23
Fix few memory leaks in HIP

SarbojitAMD:
Fixed few memory leaks showed up in cppcheck scan.

### DIFF
--- a/src/hip_module.cpp
+++ b/src/hip_module.cpp
@@ -330,22 +330,18 @@ hipError_t ihipExtLaunchMultiKernelMultiDevice(hipLaunchParams* launchParamsList
         return hipErrorInvalidValue;
     }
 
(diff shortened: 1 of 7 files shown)


src/hip_module.cpp
```diff
@@ -376,6 +376,7 @@ hipError_t ihipExtLaunchMultiKernelMultiDevice(hipLaunchParams* launchParamsList
 
         if(globalWorkSizeX > UINT32_MAX || globalWorkSizeY > UINT32_MAX || globalWorkSizeZ > UINT32_MAX)
         {
+            free(kds);
```
scchan: it would be better to change `kds` into a `std::vector` then we don't need to explicitly free it

SarbojitAMD: Thanks, updated the PR accordingly

PR #1974, opened 2020-03-24
Kernarg offset fix for COV3

david-salinas:
With COV3 the kernarg metadata uses offset instead of alignment.  Currently, the HIP RT attempts to calculate the alignment from the size and offset of arguments; but this really cannot be done accurately, especially for kernels with many arguments with large sizes (structs).  So we should just use the offset when setting up the kernel argument buffer.  But for COV2, when we read the metadata, we need to now calculate the offset.


scchan:
Could you please resolve the conflict and remove the commit noise about merges into and-master? Thanks.

PR #1979, opened 2020-03-26
implement expected behavior for SWDEV-224941

arghdos:


arghdos:
This is a fix for !1937, in which we return an error code rather than silently changing the user's copy direction for them.

src/hip_memory.cpp
```diff
@@ -229,6 +229,9 @@ void d2h_copy(void* __restrict dst, const void* __restrict src, size_t n,
     const auto di{info(dst)};
     const auto is_locked{di.type == HSA_EXT_POINTER_TYPE_LOCKED};
 
+    if (di.size != is_cpu_owned || si.size == is_cpu_owned) {
+        throw ihipException(hipErrorInvalidMemcpyDirection);
+    }
     if (!is_locked && si.size == is_cpu_owned) {
         return do_std_memcpy(dst, src, n);
```
SarbojitAMD: Same as my previous comment

tests/src/runtimeApi/memory/hipMemcpyNegativeTests.cpp
```diff
@@ -30,24 +30,25 @@ int main() {
...
-    HIPCHECK(hipMemcpy(Ad, A, sizeof(int), hipMemcpyDeviceToHost));
-    HIPCHECK(hipMemcpy(A,  Ad, sizeof(int), hipMemcpyHostToDevice));
-    HIPCHECK(hipMemcpy(Ad, Bd, sizeof(int), hipMemcpyHostToHost));
-    HIPCHECK(hipMemcpy(A,  A, sizeof(int), hipMemcpyDeviceToDevice));
-    
+    // If the passed pointers do not match the kind, we should return a
+    // hipErrorInvalidMemcpyDirection error
+    HIPASSERT(hipMemcpy(Ad, A, sizeof(int), hipMemcpyDeviceToHost) == hipErrorInvalidMemcpyDirection);
+    HIPASSERT(hipMemcpy(A,  Ad, sizeof(int), hipMemcpyHostToDevice) == hipErrorInvalidMemcpyDirection);
+    HIPASSERT(hipMemcpy(Ad, Bd, sizeof(int), hipMemcpyHostToHost) == hipErrorInvalidMemcpyDirection);
```
SarbojitAMD: On nvcc this test passes may be we can hide it for nvcc path and enable all tests for both nvcc and hip

arghdos:
Sarbojit -- we seem to not be on the same page on this, so I wanted to have a more in depth conversation here. From my point of view:

Providing an incorrect pointer type / memory copy direction is a **coding error, regardless of whether it will cause a segfault**.  So for instance, on the D2H copy, if you provide a device pointer as the destination, this is an incorrect use of the API, regardless of whether it is technically possibly to do the copy (i.e., largebar).  This is very important to identify with an error code because **performing the copy may result in sub-optimal performance**.  

For instance, if my code is incorrectly assumes a pinned host pointer (i.e., alloc'd with hipHostMalloc) is a device pointer, I can pass it to a kernel as an argument, and everything will work as expected on a non large-BAR system.  However, the performance will be very, very poor as today I will be using zero-access copies to access elements in the kernel.

To bring this back to the the copy API, consider a typical use case where we copy an array to the device before running a kernel.  Here we technically *can* run a H2D copy where the destination is a pinned host buffer, however the resulting copy perf. would be poor.

- One route is to implement what you did in your original PR, where the pointer info determines the true copy directions.  I **strongly oppose** this direction, as there is no visibility to the programmer that the runtime changed the meaning of their code without their knowledge.  This will cause issues if the user runs this code on any other backend (e.g., NVCC, a new HIP backend, etc.).  Further this does nothing to address the situation above (accidentally using a host-pinned buffer in a kernel).  Finally, it runs completely contradictory to an error code based API -- why check error codes at all, if the runtime automagically fixes it for you (assuming we do so correctly...)?

- The other route is treat this as a coding error like in CUDA (with the caveat of the D2D copy).  This makes it easy for a programmer to determine the source of the error (i.e., by checking the error code).  This may cause regressions in existing code, but again **these regressions are coding errors**...

SarbojitAMD:
> Sarbojit -- we seem to not be on the same page on this, so I wanted to have a more in depth conversation here. From my point of view:
> 
> Providing an incorrect pointer type / memory copy direction is a **coding error, regardless of whether it will cause a segfault**. So for instance, on the D2H copy, if you provide a device pointer as the destination, this is an incorrect use of the API, regardless of whether it is technically possibly to do the copy (i.e., largebar). This is very important to identify with an error code because **performing the copy may result in sub-optimal performance**.
> 
> For instance, if my code is incorrectly assumes a pinned host pointer (i.e., alloc'd with hipHostMalloc) is a device pointer, I can pass it to a kernel as an argument, and everything will work as expected on a non large-BAR system. However, the performance will be very, very poor as today I will be using zero-access copies to access elements in the kernel.
> 
> To bring this back to the the copy API, consider a typical use case where we copy an array to the device before running a kernel. Here we technically _can_ run a H2D copy where the destination is a pinned host buffer, however the resulting copy perf. would be poor.
> 
> * One route is to implement what you did in your original PR, where the pointer info determines the true copy directions.  I **strongly oppose** this direction, as there is no visibility to the programmer that the runtime changed the meaning of their code without their knowledge.  This will cause issues if the user runs this code on any other backend (e.g., NVCC, a new HIP backend, etc.).  Further this does nothing to address the situation above (accidentally using a host-pinned buffer in a kernel).  Finally, it runs completely contradictory to an error code based API -- why check error codes at all, if the runtime automagically fixes it for you (assuming we do so correctly...)?
> * The other route is treat this as a coding error like in CUDA (with the caveat of the D2D copy).  This makes it easy for a programmer to determine the source of the error (i.e., by checking the error code).  This may cause regressions in existing code, but again **these regressions are coding errors**...

Sure, please address review comments 
1. Remove if block "if (!is_locked && si.size == is_cpu_owned) {" which is dead code now 
2. It seems some dtests are failing need to fix them

mangupta:

PR #1980, opened 2020-03-27
Initial support for bfloat16

ansurya:
Initial support  is essentially clone of the rocBLAS bfloat16 header.
Added testcase to validate different bfloat16 operations on Host and Device.


include/hip/hcc_detail/hip_bfloat16.h
```diff
@@ -0,0 +1,288 @@
...
+    }
+    constexpr __host__ __device__ bool isnan(hip_bfloat16 a)
+    {
+        return !(~a.data & 0x7f80) && +(a.data & 0x7f);
+    }
+    constexpr __host__ __device__ bool iszero(hip_bfloat16 a)
+    {
+        return !(a.data & 0x7fff);
+    }
+    inline hip_bfloat16 sin(hip_bfloat16 a)
```
b-sumner: We can remove these two math functions for now.

include/hip/hip_bfloat16.h
```diff
@@ -0,0 +1,36 @@
...
+
+#ifndef HIP_INCLUDE_HIP_HIP_BFLOAT16_H
+#define HIP_INCLUDE_HIP_HIP_BFLOAT16_H
+
+#include <hip/hip_common.h>
+
+#if defined(__HIP_PLATFORM_HCC__) && !defined(__HIP_PLATFORM_NVCC__)
+#include <hip/hcc_detail/hip_bfloat16.h>
+#elif defined(__HIP_PLATFORM_NVCC__) && !defined(__HIP_PLATFORM_HCC__)
+#else
```
b-sumner: Can't we use the same implementation for NVCC?

tests/src/deviceLib/hipBfloat16.cpp
```diff
@@ -0,0 +1,148 @@
...
+static std::mt19937 rng(dev());
+
+inline float getRandomFloat(float min = 10, float max = 230) {
+    std::uniform_real_distribution<float> gen(min, max);
+    return gen(rng);
+}
+
+__host__ __device__ bool testConversion(float a, hip_bfloat16 b) {
+  float c = float(b);
+  // float relative error should be less than 1/(2^7) since bfloat16
```
b-sumner: This function is probably misnamed.  I would call it testRelativeAccuracy or something like that.

tests/src/deviceLib/hipBfloat16.cpp
```diff
@@ -0,0 +1,148 @@
...
+#include <random>
+
+#define PI 3.14
+#define SIZE 100
+using namespace std;
+
+static std::random_device dev;
+static std::mt19937 rng(dev());
+
+inline float getRandomFloat(float min = 10, float max = 230) {
```
b-sumner: I understand why this is avoiding zero, and I assume we need max-min != max, but why not make the interval bigger?

ansurya: done.

ansurya:
Test fails on Nvidia GPU, no development GPU s to debug the issue.

PR #1798, opened 2020-01-17
Fix hipMemcpy3D

gargrahul:
Fixes #1790 and #1791. hipMemcpy3D still requires further refactoring for different input and output combinations.

PR #1987, opened 2020-03-31
Fix 2D and 3D memset

gargrahul:


psychocoderHPC:
From looking to the code it looks good. I will also perform some test within the next days.

From the performance perspective it would make more sende to use a kernel to perform the ND memset instead of having a loop and spawn multiple 1D memsets. IMO such an optimization should be shipped later as separate pull request.

psychocoderHPC:
Fix is working. 
@gargrahul I suggest to add CI test similar you added in #1798

src/hip_memory.cpp
```diff
@@ -2219,16 +2219,40 @@ hipError_t hipMemset(void* dst, int value, size_t sizeBytes) {
...
+    if ((pitch == width) && (height == setHeight)) {
+        sizeBytes = pitch * setHeight * depth;
+        if(async)
+            return ihipMemsetAsync(dst, value, sizeBytes, stream, copyDataType);
+        else
+            return ihipMemsetSync(dst, value, sizeBytes, nullptr, copyDataType);
+    } else {
+        for(size_t i = 0; i < depth; ++i) {
+            for(size_t j = 0; j < setHeight; ++j) {
+                int* dstPtr = (int*) ((unsigned char*) dst + i * height * pitch + j * pitch);
```
arghdos, line 2235: Wont this only work if dst is a 4-byte type?

arghdos, line 2235: Is there any reason to cast this to an int*? Shouldn't it be void*?

gargrahul: I have corrected this.

PR #1984, opened 2020-03-30
IPC events

jeffdaily:

PR #1996, opened 2020-04-02
add IPC event support

jeffdaily:
Replaces #1984 .

Instead of using HSA IPC signals, this implementation only relies on variables of std::atomic<int> type inside a shared memory region.

The attached test only really makes sure hipEvent IPC calls for a single process are providing correct return values.

A better test is attached, adapted from CUDA samples.
[simpleIPC.zip](https://github.com/ROCm-Developer-Tools/HIP/files/4424278/simpleIPC.zip)

### DIFF
--- a/include/hip/hcc_detail/hip_runtime_api.h
+++ b/include/hip/hcc_detail/hip_runtime_api.h
@@ -97,8 +97,6 @@ typedef int hipDevice_t;
 
 typedef struct ihipStream_t* hipStream_t;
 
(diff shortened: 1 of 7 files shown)


scchan:
@jeffdaily this should just work with hip-clang/rocclr as well?

jeffdaily:
@scchan The test program is universal to both HIP/HCC and hip-clang/rocclr.  However, all other changes are specific to the HIP/HCC runtime only.

PR #2003, opened 2020-04-06
Adding regex for comprehensive match for compiler names

cjatin:
Adding regex for comprehensive match.
It should match with hcc and /opt/rocm/bin/hcc

### DIFF
--- a/packaging/hip-targets-release.cmake
+++ b/packaging/hip-targets-release.cmake
@@ -12,7 +12,7 @@ get_filename_component(_IMPORT_PREFIX "${_DIR}/../../../" REALPATH)
 
 # Import target "hip::hip_hcc_static" for configuration "Release"
 set_property(TARGET hip::hip_hcc_static APPEND PROPERTY IMPORTED_CONFIGURATIONS RELEASE)
-if(HIP_COMPILER STREQUAL "hcc" OR HIP_COMPILER STREQUAL "clang")
+if(CMAKE_CXX_COMPILER MATCHES "(.*)hcc" OR CMAKE_CXX_COMPILER MATCHES "(.*)clang(.*)")
 set_target_properties(hip::hip_hcc_static PROPERTIES
   IMPORTED_LINK_INTERFACE_LANGUAGES_RELEASE "CXX"
   IMPORTED_LINK_INTERFACE_LIBRARIES_RELEASE "hc_am"
@@ -30,7 +30,7 @@ list(APPEND _IMPORT_CHECK_FILES_FOR_hip::hip_hcc_static "${_IMPORT_PREFIX}/lib/l
 
 # Import target "hip::hip_hcc" for configuration "Release"
 set_property(TARGET hip::hip_hcc APPEND PROPERTY IMPORTED_CONFIGURATIONS RELEASE)
-if(HIP_COMPILER STREQUAL "hcc" OR HIP_COMPILER STREQUAL "clang")
+if(CMAKE_CXX_COMPILER MATCHES "(.*)hcc" OR CMAKE_CXX_COMPILER MATCHES "(.*)clang(.*)")
 set_target_properties(hip::hip_hcc PROPERTIES
   IMPORTED_LINK_INTERFACE_LIBRARIES_RELEASE "hcc::hccrt;hcc::hc_am"
   IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/lib/libhip_hcc.so"
@@ -38,7 +38,6 @@ set_target_properties(hip::hip_hcc PROPERTIES
   )
 else()
 set_target_properties(hip::hip_hcc PROPERTIES
-  IMPORTED_LINK_INTERFACE_LIBRARIES_RELEASE "hcc::hccrt"
   IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/lib/libhip_hcc.so"
   IMPORTED_SONAME_RELEASE "libhip_hcc.so"
   )
--- a/packaging/hip-targets.cmake
+++ b/packaging/hip-targets.cmake
@@ -86,7 +86,7 @@ set_target_properties(hip::host PROPERTIES
 # Create imported target hip::device
 add_library(hip::device INTERFACE IMPORTED)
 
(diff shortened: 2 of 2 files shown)


packaging/hip-targets-release.cmake
```diff
@@ -30,15 +30,14 @@ list(APPEND _IMPORT_CHECK_FILES_FOR_hip::hip_hcc_static "${_IMPORT_PREFIX}/lib/l
...
   IMPORTED_LINK_INTERFACE_LIBRARIES_RELEASE "hcc::hccrt;hcc::hc_am"
   IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/lib/libhip_hcc.so"
   IMPORTED_SONAME_RELEASE "libhip_hcc.so"
   )
 else()
 set_target_properties(hip::hip_hcc PROPERTIES
-  IMPORTED_LINK_INTERFACE_LIBRARIES_RELEASE "hcc::hccrt"
   IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/lib/libhip_hcc.so"
   IMPORTED_SONAME_RELEASE "libhip_hcc.so"
   )
```
emankov, line 43: The following `hip::hip_hcc` properties are unconditional and might be brought out of `if() ... else()`:
```cmake  
IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/lib/libhip_hcc.so"
IMPORTED_SONAME_RELEASE "libhip_hcc.so"
```

packaging/hip-targets-release.cmake
```diff
@@ -12,7 +12,7 @@ get_filename_component(_IMPORT_PREFIX "${_DIR}/../../../" REALPATH)
 
 # Import target "hip::hip_hcc_static" for configuration "Release"
 set_property(TARGET hip::hip_hcc_static APPEND PROPERTY IMPORTED_CONFIGURATIONS RELEASE)
-if(HIP_COMPILER STREQUAL "hcc" OR HIP_COMPILER STREQUAL "clang")
+if(CMAKE_CXX_COMPILER MATCHES "(.*)hcc" OR CMAKE_CXX_COMPILER MATCHES "(.*)clang(.*)")
 set_target_properties(hip::hip_hcc_static PROPERTIES
   IMPORTED_LINK_INTERFACE_LANGUAGES_RELEASE "CXX"
   IMPORTED_LINK_INTERFACE_LIBRARIES_RELEASE "hc_am"
```
emankov, line 18: The following `hip::hip_hcc_static` properties are unconditional and might be brought out of `if() ... else()`; `else()` is exhaustive:
```cmake
IMPORTED_LINK_INTERFACE_LANGUAGES_RELEASE "CXX"
IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/lib/libhip_hcc_static.a"
```

packaging/hip-targets-release.cmake
```diff
@@ -30,15 +30,14 @@ list(APPEND _IMPORT_CHECK_FILES_FOR_hip::hip_hcc_static "${_IMPORT_PREFIX}/lib/l
 
 # Import target "hip::hip_hcc" for configuration "Release"
 set_property(TARGET hip::hip_hcc APPEND PROPERTY IMPORTED_CONFIGURATIONS RELEASE)
-if(HIP_COMPILER STREQUAL "hcc" OR HIP_COMPILER STREQUAL "clang")
+if(CMAKE_CXX_COMPILER MATCHES "(.*)hcc" OR CMAKE_CXX_COMPILER MATCHES "(.*)clang(.*)")
```
yxsamliu, line 33: This should be

if (HIP_RUNTIME STREQUAL 'HCC')

whether the target requires these libs depend on whether the runtime is HIP/HCC, not depending on the compiler.

packaging/hip-targets.cmake
```diff
@@ -86,7 +86,7 @@ set_target_properties(hip::host PROPERTIES
 # Create imported target hip::device
 add_library(hip::device INTERFACE IMPORTED)
 
-if(HIP_COMPILER STREQUAL "hcc")
+if(CMAKE_CXX_COMPILER MATCHES "(.*)hcc")
```
yxsamliu, line 89: This should be

if (HIP_RUNTIME STREQUAL 'HCC')

mangupta: Agreed. For hip::device target, it should be fine to only test for HIP_RUNTIME. Device compilation cannot be done using standard c/c++ compilers.

packaging/hip-targets-release.cmake
mangupta: @yxsamliu The problem I believe is that adding hcc::hccrt also implies -hc flag for compilation. However there are multiple math libraries which use HIP APIs in host code but compile using standard c/c++ compilers. So to handle this we need to only link libhip_hcc.so, but not add the -hc flag. We can only do this by testing which compiler is being used. However given that this is a shared file, we should guard this with a ```AND HIP_RUNTIME STREQUAL 'HCC'```

packaging/hip-targets-release.cmake
pfultz2: Actually, `IMPORTED_LINK_INTERFACE_LIBRARIES_RELEASE "hcc::hccrt;hcc::hc_am"` shouldnt be there either, so the whole `if` statement can be removed.

pfultz2:
> So to handle this we need to only link libhip_hcc.so, but not add the -hc flag.

That is what the `hip::host` target is for. The `hip::hip_hcc` target should not add `hccrt` as a dependency. That is incorrect and inconsistent with `make install`(which does do it corrrectly). 

The PR#1860 should fix that, and there should be no need to check for the compiler name.

packaging/hip-targets-release.cmake
yxsamliu: If hcc::hccrt implies -hc flag, then only HCC compiler can handle it, since HIP-Clang does not support -hc flag.

However, don't math libs built with standard C/C++ compilers need to link to HCC runtime libraries when they link with HIP/HCC runtime library?

mangupta:
Closing since PR #2010 has been merged.

PR #2008, opened 2020-04-08
Disable hip_memory.h

yxsamliu:
Currently device side malloc is not working and takes excessive
device memory.

Disable it for now until a working malloc is implemented.

Change-Id: I1ad908c1c53a83752383b4be96688a848642c699


gargrahul:
Should this be combined with #1907 ?

yxsamliu:
better to use that one

PR #2009, opened 2020-04-08
Disable device side malloc

yxsamliu:
Currently device side malloc is not working and takes excessive
device memory.

Disable it for now until a working malloc is implemented.

Change-Id: I1ad908c1c53a83752383b4be96688a848642c699


yxsamliu:
Since this is urgent, and the existing PR has issues. I will just open a new one.

include/hip/hcc_detail/hip_runtime.h
```diff
@@ -305,11 +310,15 @@ static constexpr Coordinates<hip_impl::WorkitemId> threadIdx{};
...
 #if __HCC_OR_HIP_CLANG__
+#if __HIP_ENABLE_MALLOC__
 extern "C" __device__ void* __hip_malloc(size_t);
 extern "C" __device__ void* __hip_free(void* ptr);
-
 static inline __device__ void* malloc(size_t size) { return __hip_malloc(size); }
 static inline __device__ void* free(void* ptr) { return __hip_free(ptr); }
+#else
+static inline __device__ void* malloc(size_t size) { return nullptr; }
+static inline __device__ void* free(void* ptr) { return nullptr; }
```
scchan, line 320: Perhaps having a compilation error would be better?

yxsamliu: There may be cases they need these functions to compile.
They will fail at runtime anyway.

scchan: It's much easier to fix a build error than to debug a runtime failure.

yxsamliu: ok. removed

yxsamliu:
> Should we call this **HIP_ENABLE_DEVICE_MALLOC** ?

good point. fixed.

Issue #2011, opened 2020-04-09
hiprtc library used but not declared of compiler is not hcc


malcolmroberts:
In line 307 of CMakeLists.txt, we add the hiprtc library.  Later, we set target properties even if the library has not been added.  When using hip-clange, this produces a cmake failure.

Epliz:
Indeed, it is not possible to compile since #1971 from @scchan . @srinivamd seems to have seen this issue.  Would be great if someone was taking action.

Given that there is a Jenkinsfile, they must have some CI/CD. But it seems that there is nothing testing hip-clang in there. Would be great if testing of hip-clang was added in there to avoid this kind of issue in the future.

Epliz:
@malcolmroberts already opened the pull request #2012 to fix the issue. @scchan or @srinivamd can you have a look at it / add a reviewer to it?

malcolmroberts:
Thanks for looking at this.

My proposed patch isn't based on a deep analysis of the cmake code, but it might be enough.

I found this issue when following the hip-clang installation instructions.

srinivamd:
PR https://github.com/ROCm-Developer-Tools/HIP/pull/2005 was submitted for the same. Your fix looks good to me.

ppanchad-amd:
@malcolmroberts Please test with latest ROCm 6.0.2 (HIP 6.0.32831).  Please close ticket if not reproducible. Thanks!

PR #2012, opened 2020-04-09
Only set properties for hiprtc library if it has been added.

malcolmroberts:

PR #2014, opened 2020-04-10
fix CMakeLists.txt for hip-clang

Epliz:
Fixes the issue #2011 .


Epliz:
seems like there was already a pull request : #2012

PR #2015, opened 2020-04-10
Adding gcnArch char array for hipInfo

kjayapra-amd:


kjayapra-amd:
labels:pr:needs_staging

mangupta:
#2045 is a more complete change for the same.

PR #2023, opened 2020-04-17
Define hipExtEnableOutOfOrderLaunch

gargrahul:


include/hip/hip_ext.h
```diff
@@ -97,7 +99,7 @@ hipError_t hipExtModuleLaunchKernel(hipFunction_t f, uint32_t globalWorkSizeX,
                                     hipStream_t hStream, void** kernelParams, void** extra,
                                     hipEvent_t startEvent = nullptr,
                                     hipEvent_t stopEvent = nullptr,
-                                    uint32_t flags = 0);
+                                    uint32_t flags = hipExtInOrderLaunch);
```
jlgreathouse, line 102: The flags variable is still not defined in the doxygen comment block up above.

include/hip/hip_ext.h
```diff
@@ -60,6 +60,8 @@ hipError_t hipHccGetAcceleratorView(hipStream_t stream, hc::accelerator_view** a
 
 #endif  // #ifdef __HCC__
 
+#define hipExtInOrderLaunch 0
```
jlgreathouse, line 63: I believe you're using the flag as a bitfield (based on how later only check the bitwise & of this variable). So three things:
1) the "0" state should be a default, not just for InOrderLaunch. In the future, we may want to make bit 2 do something (like maybe force the stream to have a non-shared HSA queue, who knows). then the "0" state would be "InOrderLaunchAndNotSharedQueue" or something like that.
2) The EnableOutOfOrderLaunch would be more cleanly defined as either "1 << 0" or 0x1 instead of 1. That way, if we ever add more flags, they would be "1 << 1", "1 << 2" or "0x2" "0x4". This is defensive coding to prevent someone from accidentally trying to define the third flag as "3", which would (because of your bitwise check below) set the OutOfOrderLaunch flag.
3) You might want to change the name to hipExtFlagEnableOutOfOrderLaunch or something like that, so it's easy to figure out this is a #define for a flag.

Based on #1, I don't think you need to define a default "hipExtInOrderLaunch". flag=0 is just a normal launch without any of the extension flags. But then you will need to document the flag variable in the kernel launch APIs documentation below, so that folks know the default of "flags = 0" means "we are not using any of the hipExtFlagEnable flags defined in hip_ext.h".

include/hip/hip_ext.h
```diff
@@ -142,7 +144,7 @@ inline
 void hipExtLaunchKernelGGL(F kernel, const dim3& numBlocks,
                            const dim3& dimBlocks, std::uint32_t sharedMemBytes,
                            hipStream_t stream, hipEvent_t startEvent,
-                           hipEvent_t stopEvent, std::uint32_t flags,
+                           hipEvent_t stopEvent, std::uint32_t flags = hipExtInOrderLaunch,
```
jlgreathouse, line 147: Should we also have a doxygen block here defining this, so that users know how to use this function? Especially how to call the flags variable.

mangupta:

PR #2025, opened 2020-04-17
add constexpr constructor for vector types

scchan:
Change-Id: I45bb0537d6a24ee50b548c2fd8b4f20518764813

### DIFF
--- a/include/hip/hcc_detail/hip_vector_types.h
+++ b/include/hip/hcc_detail/hip_vector_types.h
@@ -313,6 +313,22 @@ THE SOFTWARE.
         using value_type = T;
 
         __host__ __device__
+        constexpr
+        HIP_vector_base() noexcept : data{} {};
+        __host__ __device__
+        explicit
+        constexpr
+        HIP_vector_base(T x) noexcept : data{x} {}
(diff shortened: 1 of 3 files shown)


include/hip/hcc_detail/hip_vector_types.h
```diff
@@ -312,6 +312,22 @@ THE SOFTWARE.
 
         using value_type = T;
 
+        __host__ __device__
+        constexpr
+        HIP_vector_base() noexcept : data{} {};
```
yxsamliu, line 317: Can we remove the default ctors ? Since cuda-clang does not allow shared variables to have non-trivial ctors since shared variables cannot be initialized. Adding this ctor will prevent vectors from being used as shared variables.

scchan: @AlexVlx  do we provide guarantee that it will get zero-initialized by default?

yxsamliu: nvcc does not initialize int2

https://godbolt.org/z/9HR2MQ

mangupta:

PR #2026, opened 2020-04-18
Let hipcc treats .cpp files as C++ files for HIP-Clang

yxsamliu:
This is to match nvcc behavior, speed up compilation of C++ programs, and fix some
compilation issue where C++ programs are compiled as HIP programs.

Currently it is controlled by an environment variable
HIP_COMPILE_CXX_AS_HIP. By default it is 1, where
hipcc treats .cpp files as HIP programs. If it is
set to 0, hipcc will treat .cpp files as C++ programs.
This is because some math libraries are still not
ready for the change, however rocBLAS and rocFFT
require this feature for OpenMP, therefore put it
under an environment variable so that rocBLAS
and rocFFT can use it.

Change-Id: I56a51e27079df850ee39d4217fb647c22d79f612

### DIFF
--- a/bin/hipcc
+++ b/bin/hipcc
@@ -85,6 +85,7 @@ $HIP_LIB_PATH=$ENV{'HIP_LIB_PATH'};
 $HIP_CLANG_PATH=$ENV{'HIP_CLANG_PATH'};
 $DEVICE_LIB_PATH=$ENV{'DEVICE_LIB_PATH'};
 $HIP_CLANG_HCC_COMPAT_MODE=$ENV{'HIP_CLANG_HCC_COMPAT_MODE'}; # HCC compatibility mode
+$HIP_COMPILE_CXX_AS_HIP=$ENV{'HIP_COMPILE_CXX_AS_HIP'} // "1";
 
 if (defined $HIP_VDI_HOME) {
     $HIP_INFO_PATH= "$HIP_VDI_HOME/lib/.hipInfo";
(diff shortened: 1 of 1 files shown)


yxsamliu:
This has been tested to work for rocBLAS for OpenMP issue.

bin/hipcc
```diff
@@ -762,20 +775,24 @@ if ($HIP_PLATFORM eq "clang") {
     }
     # Do not pass -mllvm on Windows since there is a clang bug causing duplicate -mllvm options in clang -cc1 on Windows.
     # ToDo : remove restriction for Windows after clang bug is fixed.
-    if (!$funcSupp and $optArg ne "-O0" and not $isWindows) {
+    if (!$funcSupp and $optArg ne "-O0" and not $isWindows and $hasHIP) {
         $HIPCXXFLAGS .= " -mllvm -amdgpu-early-inline-all=true -mllvm -amdgpu-function-calls=false";
         if ($needLDFLAGS and not $needCXXFLAGS) {
             $HIPLDFLAGS .= " -mllvm -amdgpu-early-inline-all=true -mllvm -amdgpu-function-calls=false";
```
srinivamd, line 781: Are these LDFLAGS needed based on this comment: https://github.com/ROCm-Developer-Tools/HIP/pull/2042#pullrequestreview-399934463 ?
Does hipcc on an object file as input with no source files work? (That was the attempted fix of #2042 )

yxsamliu: Yes. There is condition $hasHIP, which is true only if the command has .hip file or .cpp/.cc/.cxx files together with -x hip.

If there is only .o files, these options are not added.

srinivamd: Ok. The flags are getting trickier for me to track. Maybe refactoring could help?
Should LDFLAGS be prefixed by `-Wl` to be passed to the linker if that's what they are meant for as the name might suggest?

aaronenyeshi:
This is merged into amd-master-next

bin/hipcc
yxsamliu: in LD flags, the options are intended for clang, not for lld. When -fgpu-rdc is passed, .o files are handled by clang to do device linking and codegen.
We need some minor fixes to check for -fgpu-rdc for LD flags. However the current priority is converge master and amd-master-branch, so we will fix this later.

PR #2028, opened 2020-04-19
Enable template max and min for HIP-Clang

yxsamliu:
It was for HCC only. HIP-Clang also needs it for __fp16 since AMDMIGraphX uses it.

Change-Id: Id49322b7b89ef799accdf6b47627a6fce51d1ab5


pfultz2:
Where are the tests? There should be a test case added for this.

yxsamliu:
> Where are the tests? There should be a test case added for this.

Test added.

yxsamliu:
This one is already in amd-master-next branch. This is a backport to master branch.

PR #2031, opened 2020-04-20
[dtests] Fixing module texture test

cjatin:
It would pass true regardless of the mismatch.
Now instead of break, return false on single mismatch


mangupta:

PR #2032, opened 2020-04-20
Merge in the rocclr based hip runtime

mangupta:


saleelk:
> Can you tell me whether these headers have any local changes compared to original OpenCL headers?
> Khronos does not seem to have history for their headers bofore they've released the unified version:
> https://github.com/KhronosGroup/OpenCL-Headers
> 
> Can the current khronos' OpenCL headers be used instead of the headers under amdocl?

I had promoted the latest headers to amd-master-next branch on gerrit for HIP. So the merge should have it. If they aren't then it should be checked.

PR #2034, opened 2020-04-20
extractkernel obj dump args to use double dashes

aaronenyeshi:
More recent llvm-objdump has changed disassemble and mcpu arguments to require double dashes.


aaronenyeshi:
Hi Maneesh, looks like when we moved this over, we missed this patch: https://github.com/RadeonOpenCompute/hcc/pull/1440

PR #2036, opened 2020-04-22
[HIPTEST] Added IPC Event Tests

satyanveshd:


tests/unit/IPC/hipIpcEventHandle.cpp
```diff
@@ -0,0 +1,201 @@
...
+         
+        hipEvent_t ipc_event;
+        hipStream_t stream;
+        HIPCHECK(hipStreamCreate(&stream));
+  
+        HIPCHECK(hipIpcOpenEventHandle(&ipc_event, shrd_mem->eventHandle));
+        
+        HIPCHECK(hipEventRecord(ipc_event, NULL));
+
+        HIPCHECK(hipEventRecord(ipc_event, stream))
```
SarbojitAMD, line 120: Not sure what are we testing here call record on same event overrides previous state.

tests/unit/IPC/hipIpcEventHandle.cpp
```diff
@@ -0,0 +1,201 @@
...
+    }
+    
+    pid_t pid;
+
+    pid = fork();
+    
+    if(pid != 0) {
+    
+        int dev0 = 0;
+        int dev1 = 1;
```
SarbojitAMD, line 84: May I know why are we restricting the test only between devID =0 & devId=1? There can be a situation where even though setup is multi device but Dev0 and Dev1 do not have peer access but Dev0 and Dev3 has. I feel better approach would be to have a query function and prepare peer access list and then go for test.

tests/unit/IPC/hipIpcEventHandle.cpp
```diff
@@ -0,0 +1,201 @@
...
+  
+        HIPCHECK(hipIpcOpenEventHandle(&ipc_event, shrd_mem->eventHandle));
+        
+        HIPCHECK(hipEventRecord(ipc_event, NULL));
+
+        HIPCHECK(hipEventRecord(ipc_event, stream))
+        
+        HIPCHECK(hipStreamWaitEvent(stream, ipc_event, 0));
+        
+        HIPCHECK(hipEventSynchronize(ipc_event));
```
SarbojitAMD, line 124: May I know why do we need two synchronizations (hipStreamWaitEvent & hipEventSynchronize)?

tests/unit/IPC/hipIpcEventHandleNegative.cpp
```diff
@@ -0,0 +1,175 @@
...
+    HIPCHECK(hipDeviceCanAccessPeer(&canAccessPeer10, dev1, dev0));
+    if (!canAccessPeer01 || !canAccessPeer10) {
+        return -1;
+    }
+
+    HIPCHECK(hipSetDevice(dev0));
+    HIPCHECK(hipDeviceEnablePeerAccess(dev1, 0 /*flags*/));
+    HIPCHECK(hipSetDevice(dev1));
+    HIPCHECK(hipDeviceEnablePeerAccess(dev0, 0 /*flags*/));
+    HIPCHECK(hipSetDevice(dev0));
```
SarbojitAMD, line 48: I don't think we should have HIPCHECK(hipSetDevice(dev0)); here. It is very easy to miss this line. We should call it explicitly in the caller.

tests/unit/IPC/hipIpcEventHandle.cpp
satyanveshd: Addressed the review comments.

tests/unit/IPC/hipIpcEventHandle.cpp
satyanveshd: Modified the function to query for all the devices and find a combination of two devices with peer access on multi gpu setup.

tests/unit/IPC/hipIpcEventHandle.cpp
satyanveshd: Added the description. This only verifies the validity of the ipc event object with various event APIs.

mangupta:

PR #2042, opened 2020-04-23
fix unused argument warning when input source files

srinivamd:
Fix for SWDEV-230876 Warning messages observed while compiling tests 
Tests for source files on command-line before setting flags
Also added test for -fuse-ld=lld to set ldflags


srinivamd:
Adding @mangupta @yxsamliu for review

srinivamd:
Ok @yxsamliu didn't check #2026 first.

mangupta:
master-next is a read only mirror of the internal branch. Please submit the change there instead.

PR #2045, opened 2020-04-24
Added gcnArchName as string

satyanveshd:
gcnArchName is added to be a string instead of integer. Addresses the issue in SWDEV-227909.

Adds to and includes the changes in #2015 .

### DIFF
--- a/include/hip/hip_runtime_api.h
+++ b/include/hip/hip_runtime_api.h
@@ -114,7 +114,7 @@ typedef struct hipDeviceProp_t {
     int isMultiGpuBoard;                      ///< 1 if device is on a multi-GPU board, 0 if not.
     int canMapHostMemory;                     ///< Check whether HIP can map host memory
     int gcnArch;                              ///< AMD GCN Arch Value. Eg: 803, 701
-    char gcnArchName[256];                    ///< AMD GCN Arch Name (Arch Value as String).
+    char gcnArchName[16];                     ///< AMD GCN Arch Name (Arch Value as String).
     int integrated;            ///< APU vs dGPU
     int cooperativeLaunch;            ///< HIP device supports cooperative launch
     int cooperativeMultiDeviceLaunch; ///< HIP device supports cooperative launch on multiple devices
--- a/include/hip/hip_runtime_api.h
+++ b/include/hip/hip_runtime_api.h
@@ -114,6 +114,7 @@ typedef struct hipDeviceProp_t {
     int isMultiGpuBoard;                      ///< 1 if device is on a multi-GPU board, 0 if not.
     int canMapHostMemory;                     ///< Check whether HIP can map host memory
     int gcnArch;                              ///< AMD GCN Arch Value. Eg: 803, 701
+    char gcnArchName[256];                    ///< AMD GCN Arch Name (Arch Value as String).
     int integrated;            ///< APU vs dGPU
     int cooperativeLaunch;            ///< HIP device supports cooperative launch
     int cooperativeMultiDeviceLaunch; ///< HIP device supports cooperative launch on multiple devices
--- a/include/hip/hip_runtime_api.h
+++ b/include/hip/hip_runtime_api.h
@@ -113,8 +113,8 @@ typedef struct hipDeviceProp_t {
     size_t maxSharedMemoryPerMultiProcessor;  ///< Maximum Shared Memory Per Multiprocessor.
     int isMultiGpuBoard;                      ///< 1 if device is on a multi-GPU board, 0 if not.
     int canMapHostMemory;                     ///< Check whether HIP can map host memory
-    int gcnArch;                              ///< AMD GCN Arch Value. Eg: 803, 701
-    char gcnArchName[16];                     ///< AMD GCN Arch Name (Arch Value as String).
+    int gcnArch;                              ///< AMD GCN Arch Value (Major Number Only). Eg: 800, 1000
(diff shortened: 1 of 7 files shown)


aaronenyeshi:
+ @saadrahim @eidenyoshida - math libs may need some changes to support gcnArch as a char array instead of an int. I know there are some comparisons going on, could those be fixed, or does it need some API in HIP to support comparisons?

include/hip/hip_runtime_api.h
```diff
@@ -114,6 +114,7 @@ typedef struct hipDeviceProp_t {
     int isMultiGpuBoard;                      ///< 1 if device is on a multi-GPU board, 0 if not.
     int canMapHostMemory;                     ///< Check whether HIP can map host memory
     int gcnArch;                              ///< AMD GCN Arch Value. Eg: 803, 701
+    char gcnArchName[256];                    ///< AMD GCN Arch Name (Arch Value as String).
```
aaronenyeshi, line 117: does it need to be 256 char long? How about 8 or 16 char?

src/hip_hcc.cpp
```diff
@@ -791,6 +791,7 @@ hipError_t ihipDevice_t::initProperties(hipDeviceProp_t* prop) {
     err = hsa_agent_get_info(_hsaAgent, HSA_AGENT_INFO_NAME, &archName);
 
     prop->gcnArch = atoi(archName + 3);
+    strcpy(prop->gcnArchName, archName + 3);
```
aaronenyeshi, line 794: Should the gcnArchName include the full name gfx900, or gfx1010, rather than just 900 or 1010? It might make sense for users to expect the correct name with gfx* prefix, as in AMDGPU guide. https://llvm.org/docs/AMDGPUUsage.html#processors

saadrahim:
> 
> 
>     * @saadrahim @eidenyoshida - math libs may need some changes to support gcnArch as a char array instead of an int. I know there are some comparisons going on, could those be fixed, or does it need some API in HIP to support comparisons?

@aaronenyeshi give us a day to analyze the impact. I believe now have a macro available to have compiler type distinction between the architectures so the correct code can be generated without needing runtime comparisons.

src/hip_hcc.cpp
```diff
@@ -791,6 +791,7 @@ hipError_t ihipDevice_t::initProperties(hipDeviceProp_t* prop) {
     err = hsa_agent_get_info(_hsaAgent, HSA_AGENT_INFO_NAME, &archName);
 
     prop->gcnArch = atoi(archName + 3);
```
aaronenyeshi, line 793: Arch may be hexidecimals, so atoi will not work.

src/hip_hcc.cpp
satyanveshd: Addressed the comments. gcnArchName is changed to include the full name with gfx prefix.

src/hip_hcc.cpp
satyanveshd: gcnArch now represents a series of arch's with the help of major number. For all gfx 9 series, gcnArch will return 900 etc. For getting the exact ISA, one has to use gcnArchName as it represents full name. eg: gfx906, gfx90a etc.

include/hip/hip_runtime_api.h
```diff
@@ -114,6 +114,7 @@ typedef struct hipDeviceProp_t {
     int isMultiGpuBoard;                      ///< 1 if device is on a multi-GPU board, 0 if not.
     int canMapHostMemory;                     ///< Check whether HIP can map host memory
     int gcnArch;                              ///< AMD GCN Arch Value. Eg: 803, 701
```
aaronenyeshi, line 116: Please update the comment and all README that this is returning only major numbering only.

samples/1_Utils/hipInfo/hipInfo.cpp
```diff
@@ -138,7 +138,7 @@ void printDeviceProp(int deviceId) {
     cout << setw(w1) << "arch.hasSurfaceFuncs: " << props.arch.hasSurfaceFuncs << endl;
     cout << setw(w1) << "arch.has3dGrid: " << props.arch.has3dGrid << endl;
     cout << setw(w1) << "arch.hasDynamicParallelism: " << props.arch.hasDynamicParallelism << endl;
-    cout << setw(w1) << "gcnArch: " << props.gcnArch << endl;
+    cout << setw(w1) << "gcnArchName: " << props.gcnArchName << endl;
```
aaronenyeshi, line 141: Should we also add a line to print gcnArch?

src/hip_hcc.cpp
```diff
@@ -789,10 +789,14 @@ hipError_t ihipDevice_t::initProperties(hipDeviceProp_t* prop) {
...
-    prop->gcnArch = atoi(archName + 3);
+ 
+    std::string name = archName + 3; 
+    name = name.erase(name.size() - 2);
+    prop->gcnArch = stoi(name) * 100; 
+     
+    strcpy(prop->gcnArchName, archName);
     DeviceErrorCheck(err);
-
+
```
aaronenyeshi, line 799: white spaces

include/hip/hip_runtime_api.h
satyanveshd: Updated the comment and addressed other review comments.

leekillough:
Related: https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/120

mangupta:

PR #2049, opened 2020-04-27
[HIP-Clang] Updating the RUNPATH for clang case

pruthvistony:
- Adding the RUNPATH for all the lib/ELF generated using hipcc for
  hip-clang case.

Change-Id: I0270281d10452750c3589d9e7f4666554937d27d

### DIFF
--- a/bin/hipcc
+++ b/bin/hipcc
@@ -214,6 +214,7 @@ if ($HIP_PLATFORM eq "clang") {
     if (not $isWindows) {
       $HIPLDFLAGS .= " -Wl,--rpath-link=$HIP_LIB_PATH";
       $HIPLDFLAGS .= " -lhip_hcc";
+      $HIPLDFLAGS .= " -Wl,--enable-new-dtags -Wl,--rpath=$ROCM_PATH/lib:$ROCM_PATH/lib64:$ROCM_PATH/hip/lib ";
     } else {
       $HIPLDFLAGS .= " -lamdhip64";
     }


bin/hipcc
```diff
@@ -214,6 +214,7 @@ if ($HIP_PLATFORM eq "clang") {
     if (not $isWindows) {
       $HIPLDFLAGS .= " -Wl,--rpath-link=$HIP_LIB_PATH";
       $HIPLDFLAGS .= " -lhip_hcc";
+      $HIPLDFLAGS .= " -Wl,--enable-new-dtags -Wl,--rpath=$ROCM_PATH/lib:$ROCM_PATH/lib64:$ROCM_PATH/hip/lib ";
```
yxsamliu, line 217: What if users specified their own rpaths? Will they be overridden by the ones specified here?

pruthvistony: The user specified path should get appended to the these paths. The above paths helps in finding ROCM_PATH libraries. 
Also to note, we are adding RUNPATH so this can be overwritten by LD_LIBRARY_PATH if required by the user.

pruthvistony:
There is change in and-master-next branch as shown below
    if (not $isWindows  and not $compileOnly) {
      if ($linkType eq 0) {
        $toolArgs .= " -L$HIP_LIB_PATH -lamdhip64_static -L$ROCM_PATH/lib -lhsa-runtime64 -ldl ";
      } else {
        $toolArgs .= " -Wl,--enable-new-dtags -Wl,--rpath=$HIP_LIB_PATH:$ROCM_PATH/lib -lhip_hcc ";
      }
    }

is sufficient change. I understand it will reflect in master branch after the migration/upgrade.
So it seems this PR is not required.

PR #2061, opened 2020-05-01
SWDEV-221166 - Detect support for large bar access through HIP runtim…

AlexXAmd:
…e API

Change-Id: Iaa9756c1b5e40c1ab5afb38e44a6699fa5f6c13f

### DIFF
--- a/include/hip/hip_runtime_api.h
+++ b/include/hip/hip_runtime_api.h
@@ -136,7 +136,7 @@ typedef struct hipDeviceProp_t {
                                                     ///devices with unmatched block dimensions
     int cooperativeMultiDeviceUnmatchedSharedMem;   ///< HIP device supports cooperative launch on multiple
                                                     ///devices with unmatched shared memories
-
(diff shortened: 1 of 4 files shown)


include/hip/hip_runtime_api.h
```diff
@@ -136,7 +136,7 @@ typedef struct hipDeviceProp_t {
                                                     ///devices with unmatched block dimensions
     int cooperativeMultiDeviceUnmatchedSharedMem;   ///< HIP device supports cooperative launch on multiple
                                                     ///devices with unmatched shared memories
-
+    int isLargeBar;                  ///< 1: if it is a large PCI bar device, else 0
```
SarbojitAMD, line 139: Wondering why it can't be 'bool' when only two valid values allowed.

AlexXAmd: I followed the examples in this same struct. They use int to express bool.

AlexXAmd: Hi,
I appreciate all your comments. I will close the request now. I may address your comment in future.
Thanks,
Alex

PR #2070, opened 2020-05-05
Do not define assert when NDEBUG is defined

### LINKED ISSUE #2071: Device-side assert macro is defined even when NDEBUG is defined
Assertion arguments should only be expanded/evaluated when `NDEBUG` is not defined. See #2070 for a possible fix.

fthaler:
Fixes #2071


mangupta:

PR #2080, opened 2020-05-12
add support for cooperative group feature - thread_block

hsmahesha:


hsmahesha:
Hello @jlgreathouse, @yxsamliu and @b-sumner

The Jira ticket SWDEV-235824 is created by asking the support for cooperative group type - thread_block". This patch implements the same. Please review it.

b-sumner:
Is that really all of the operations supported on the thread_block group?  Also, what about tests?

jlgreathouse:
Yep, these are all the operations supported on the thread_block type. The `thread_group` type only offers `sync()` (which is equivalent to thread-block-level `__syncthreads`, `size()` (which is the number of threads in the block), `thread_rank()` (which is the 1D flattened thread index within the block), and `is_valid()`.

An important thing to note is that `thread_block` is a derived class of the base `thread_group` class, which I believe is accurately covered in this patch, so that's good.

There are a few more classes derived from `thread_group` that aren't yet supported, and have more complexity than this patch:

- `thread_group` generated with the `tiled_partition()` constructor.
  - This allows you to break apart a `thread_group` to create a sub-thread-block group, up to a maximum of machine-warp (32, 64) threads wide. The group must be power-of-two in width. 
  - I suspect we will need to chat about how this is handled if the input `thread_block` is from the derived `grid_group` or `multi_grid_group` class. I believe NV's CG implementation does not derive `grid_group` or `multi_grid_group` from `thread_group`, though this seems like a good idea to me (and we currently do it). So I would just want to make sure we handle that case properly.
- `thread_block_tile<int>` templated type
  - This templated type allows you to create a wavefront-or-smaller sized group that adds a bunch of extra wavefront-wide functionality like sub-wavefront shuffle, ballot, and match operations.
  - The idea is that you can generate an optimized inter-lane shuffle operations at compile time since you can (through the template) statically know how many neighboring threads are participating. e.g. if you only have 2 threads in the group, a ballot op only needs to talk to one neighbor.
- `coalesced_group`
  - This allows you to build up a `thread_group` from whatever lanes are active when the constructor is called. At that point, things like `thread_rank()` return a new reordering so that you are lane X within whatever lanes were enabled when this group was created, even if you were not lane X in the actual wavefront.
  - Note that you can also pass this coalesced group into a `tiled_partition()` constructor.

@hsmahesha it might be useful to chat, because I think that adding the basic `tiled_partition()` constructor into the `thread_group` class may be useful at this point. Then we would at least have full support for the base class (`thread_group`), the new `thread_block` you just added, `grid_group`, and `multi_grid_group` classes. We could then add the tiled and coalesced groups later when we want to think through the corner cases.

hsmahesha:
> Is that really all of the operations supported on the thread_block group? Also, what about tests?

Hello @b-sumner, as @jlgreathouse mentioned below, these are the operations supported as of now. Regarding test, I only informally and manually testing the features. Let me try, if I can add any tests to HIP test-suite.

hsmahesha:
> Yep, these are all the operations supported on the thread_block type. The `thread_group` type only offers `sync()` (which is equivalent to thread-block-level `__syncthreads`, `size()` (which is the number of threads in the block), `thread_rank()` (which is the 1D flattened thread index within the block), and `is_valid()`.
> 
> An important thing to note is that `thread_block` is a derived class of the base `thread_group` class, which I believe is accurately covered in this patch, so that's good.
> 
> There are a few more classes derived from `thread_group` that aren't yet supported, and have more complexity than this patch:
> 
> * `thread_group` generated with the `tiled_partition()` constructor.
>   
>   * This allows you to break apart a `thread_group` to create a sub-thread-block group, up to a maximum of machine-warp (32, 64) threads wide. The group must be power-of-two in width.
>   * I suspect we will need to chat about how this is handled if the input `thread_block` is from the derived `grid_group` or `multi_grid_group` class. I believe NV's CG implementation does not derive `grid_group` or `multi_grid_group` from `thread_group`, though this seems like a good idea to me (and we currently do it). So I would just want to make sure we handle that case properly.
> * `thread_block_tile<int>` templated type
>   
>   * This templated type allows you to create a wavefront-or-smaller sized group that adds a bunch of extra wavefront-wide functionality like sub-wavefront shuffle, ballot, and match operations.
>   * The idea is that you can generate an optimized inter-lane shuffle operations at compile time since you can (through the template) statically know how many neighboring threads are participating. e.g. if you only have 2 threads in the group, a ballot op only needs to talk to one neighbor.
> * `coalesced_group`
>   
>   * This allows you to build up a `thread_group` from whatever lanes are active when the constructor is called. At that point, things like `thread_rank()` return a new reordering so that you are lane X within whatever lanes were enabled when this group was created, even if you were not lane X in the actual wavefront.
>   * Note that you can also pass this coalesced group into a `tiled_partition()` constructor.
> 
> @hsmahesha it might be useful to chat, because I think that adding the basic `tiled_partition()` constructor into the `thread_group` class may be useful at this point. Then we would at least have full support for the base class (`thread_group`), the new `thread_block` you just added, `grid_group`, and `multi_grid_group` classes. We could then add the tiled and coalesced groups later when we want to think through the corner cases.

Thanks @jlgreathouse  for the comments. Let me go through the `tiled_partition()` related functionalities. Then, I will ping you for further discussion. At present, I am also bit busy with few P1 tickets. Can we close this PR and separately work on  `tiled_partition()` afterwards?

b-sumner:
This LGTM.  I don't see any reason to include tiled_partition or additional features here.

hsmahesha:
Hello @mangupta 

Please merge this PR to HIP master. It is got approved (by @b-sumner)

mangupta:
@hsmahesha Changes need to go to the internal repo.

mangupta:

PR #2083, opened 2020-05-15
Overload ihipCtx_t::getDevice() by removing const qualifier and repla…

zhaozhangjian:
…ce ihipCtx_t::getWriteableDevice() with this new ihipCtx_t::getDevice().


src/hip_hcc_internal.h
```diff
@@ -959,10 +959,9 @@ class ihipCtx_t {
     ihipCtxCritical_t& criticalData() { return _criticalData; };
 
     const ihipDevice_t* getDevice() const { return _device; };
-    int getDeviceNum() const { return _device->_deviceId; };
+    ihipDevice_t* getDevice() { return _device; };
 
-    // TODO - review uses of getWriteableDevice(), can these be converted to getDevice()
-    ihipDevice_t* getWriteableDevice() const { return _device; };
```
SarbojitAMD, line 965: Why 'const' qualifier is removed?

zhaozhangjian: By removing the 'const' qualifier, we can overload this getDevice() function instead of previously using another function - getWriteableDevice(). After this change:
'const ihipCtx_t *ptr->getDevice()' will call 'const ihipDevice_t* getDevice() const { return _device; };'
'ihipCtx_t *ptr->getDevice()' will call 'ihipDevice_t* getDevice() { return _device; };'

mangupta:

PR #2098, opened 2020-06-04
Fix include path and wrapper header

yxsamliu:
Currently std::complex and some other std functions require uses to
include hip_runtime.h before any other headers to work, which is not
reliable.

changes are made in clang to fix this issue:
https://reviews.llvm.org/D81176

which requires hipcc and HIP headers to make corresponding changes.

This patch will make sure the clang change will not break
HIP/ROCclr during this transition.

After the transition is done, we can remove explicitly setting
include path for HIP-Clang and HIP header in hipcc and hip config
cmake files and rely on clang driver to set it automatically.

Change-Id: I5d226861c2560ffa6c5ab17343a43cc378048061

### DIFF
--- a/include/hip/hcc_detail/hip_fp16_math_fwd.h
+++ b/include/hip/hcc_detail/hip_fp16_math_fwd.h
@@ -27,7 +27,7 @@ THE SOFTWARE.
 // */
 
 #include "host_defines.h"
-#if !__CLANG_HIP_RUNTIME_WRPPER_INCLUDED__
+#if !__CLANG_HIP_RUNTIME_WRAPPER_INCLUDED__
 extern "C"
 {
     __device__ __attribute__((const)) _Float16 __ocml_ceil_f16(_Float16);
@@ -80,4 +80,4 @@ extern "C"
     __device__ __attribute__((const)) __2f16 __ocml_sqrt_2f16(__2f16);
     __device__ __attribute__((const)) __2f16 __ocml_trunc_2f16(__2f16);
 }
-#endif // __CLANG_HIP_RUNTIME_WRPPER_INCLUDED__
+#endif // __CLANG_HIP_RUNTIME_WRAPPER_INCLUDED__
--- a/include/hip/hcc_detail/hip_fp16_math_fwd.h
+++ b/include/hip/hcc_detail/hip_fp16_math_fwd.h
@@ -27,7 +27,7 @@ THE SOFTWARE.
 // */
 
 #include "host_defines.h"
-
+#if !__CLANG_HIP_RUNTIME_WRPPER_INCLUDED__
 extern "C"
 {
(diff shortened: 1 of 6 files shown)


include/hip/hcc_detail/hip_fp16_math_fwd.h
```diff
@@ -80,3 +80,4 @@ extern "C"
     __device__ __attribute__((const)) __2f16 __ocml_sqrt_2f16(__2f16);
     __device__ __attribute__((const)) __2f16 __ocml_trunc_2f16(__2f16);
 }
+#endif // __CLANG_HIP_RUNTIME_WRPPER_INCLUDED__
```
emankov, line 83: `// !__CLANG_HIP_RUNTIME_WRPPER_INCLUDED__`

yxsamliu:
> Btw, where are you going to define `__CLANG_HIP_RUNTIME_WRPPER_INCLUDED__`?

It is defined in clang header file.

include/hip/hcc_detail/hip_fp16_math_fwd.h
```diff
@@ -27,7 +27,7 @@ THE SOFTWARE.
 // */
 
 #include "host_defines.h"
-
+#if !__CLANG_HIP_RUNTIME_WRPPER_INCLUDED__
```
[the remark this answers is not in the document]
yxsamliu: fixed

include/hip/hcc_detail/hip_fp16_math_fwd.h
yxsamliu: fixed

emankov: Here I meant adding `!` in all "#endif" comments to match the corresponding `#if`.

yxsamliu: sorry. this time really fixed

yxsamliu:
> LGTM, thanks.
> 
> Btw, now with https://reviews.llvm.org/D81176, I anticipate the same packaging issues with a correct `__clang_hip_runtime_wrapper.h` inclusion as we've already had with `hipify-clang` packaging -
> like [ROCm-Developer-Tools/HIPIFY#92](https://github.com/ROCm-Developer-Tools/HIPIFY/issues/92) about "incorrect" including `__clang_cuda_runtime_wrapper.h` and other clang system includes, which was fixed the following way: [ROCm-Developer-Tools/HIPIFY#126](https://github.com/ROCm-Developer-Tools/HIPIFY/issues/126).

clang only include these paths for .hip or -x hip, so probably it is not subject to that issue.

emankov:
>clang only include these paths for .hip or -x hip, so probably it is not subject to that issue.

As well as clang includes those paths for `.cu` or `-x cuda`. I think the issue might occur with using clang distributives (aka HIP artifacts/packages). But let's see it.

mangupta:

PR #2110, opened 2020-06-19
Add __attribute__((const)) to grid related functions declarations

dfukalov:
Make declarations consistent with [devicelibs' ockl.h](https://github.com/RadeonOpenCompute/ROCm-Device-Libs/blob/amd-stg-open/ockl/inc/ockl.h)
Without the attribute these functions don't have "readnone" LLVM IR attribute. And so some optimizations fails, e.g. Loop Invariant Code Motion doesn't hoist these calls out of a loop.

### DIFF
--- a/include/hip/hcc_detail/hip_runtime.h
+++ b/include/hip/hcc_detail/hip_runtime.h
@@ -402,22 +402,22 @@ typedef int hipLaunchParm;
 #pragma push_macro("__DEVICE__")
 #define __DEVICE__ static __device__ __forceinline__
 
-extern "C" __device__ size_t __ockl_get_local_id(uint);
+extern "C" __device__ __attribute__((const)) size_t __ockl_get_local_id(uint);
 __DEVICE__ uint __hip_get_thread_idx_x() { return __ockl_get_local_id(0); }
 __DEVICE__ uint __hip_get_thread_idx_y() { return __ockl_get_local_id(1); }
 __DEVICE__ uint __hip_get_thread_idx_z() { return __ockl_get_local_id(2); }
 
-extern "C" __device__ size_t __ockl_get_group_id(uint);
+extern "C" __device__ __attribute__((const)) size_t __ockl_get_group_id(uint);
 __DEVICE__ uint __hip_get_block_idx_x() { return __ockl_get_group_id(0); }
 __DEVICE__ uint __hip_get_block_idx_y() { return __ockl_get_group_id(1); }
 __DEVICE__ uint __hip_get_block_idx_z() { return __ockl_get_group_id(2); }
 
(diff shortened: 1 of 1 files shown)


yxsamliu:
I suspect `__attribute__((const))` can be used with these functions.

My understanding is that this attribute assumes the return value depends only on argument, i.e. compiler can assume foo(0) always return the same value. However, obviously the local id will return different value for different work item. I am not sure how this plays out in a divergent control flow.

My concern is that whether divergence analysis will be confused by this attribute. For example, `if (local_id(0)==0)` will be treated as a convergent control flow due to this attribute.

b-sumner:
LGTM.  I'm not aware of this attribute implying anything about convergence.

dfukalov:
> My understanding is that this attribute assumes the return value depends only on argument, i.e. compiler can assume foo(0) always return the same value....

Actually they say "return value is not affected by changes to the observable state of the program" that is true for these functions. Perhaps it's more clear if we think that a program runs once on all workitems and operates with vectors.
Other meaning is that these return values are not changed during whole program run time, for all workitems and don't depend on the state of a program.

> My concern is that whether divergence analysis will be confused by this attribute. For example, `if (local_id(0)==0)` will be treated as a convergent control flow due to this attribute.

I mean that `if (local_id(0)==0)` can be transformed to 
```
tmp = local_id(0)
<some code changing state of program>
if (tmp)
```
A compiler cannot do this without the attribute. In our case it's more complicated: we have declaration in hip header without the attribute and device library (check link in PL' description) that _has_ the attribute.
So some passes that performed before linking with devicelib are not allowed to make these transforms, but latter passes are.

yxsamliu:
> > My understanding is that this attribute assumes the return value depends only on argument, i.e. compiler can assume foo(0) always return the same value....
> 
> Actually they say "return value is not affected by changes to the observable state of the program" that is true for these functions. Perhaps it's more clear if we think that a program runs once on all workitems and operates with vectors.
> Other meaning is that these return values are not changed during whole program run time, for all workitems and don't depend on the state of a program.
> 
> > My concern is that whether divergence analysis will be confused by this attribute. For example, `if (local_id(0)==0)` will be treated as a convergent control flow due to this attribute.
> 
> I mean that `if (local_id(0)==0)` can be transformed to
> 
> ```
> tmp = local_id(0)
> <some code changing state of program>
> if (tmp)
> ```
> 
> A compiler cannot do this without the attribute. In our case it's more complicated: we have declaration in hip header without the attribute and device library (check link in PL' description) that _has_ the attribute.
> So some passes that performed before linking with devicelib are not allowed to make these transforms, but latter passes are.

I did not notice the attribute is already in device lib, then LGTM. Thanks.

dfukalov:
I have no rights to merge the change so please proceed the PL.

PR #2121, opened 2020-06-29
add extract_codeobject

david-salinas:
Preliminary extract_codeobject

### DIFF
--- a/bin/extract_codeobject
+++ b/bin/extract_codeobject
@@ -1,0 +1,426 @@
+#!/usr/bin/perl
+use strict;
+use File::Copy;
+use File::Spec;
+use File::Basename;
+use File::Which;
+use Cwd 'realpath';
+use Getopt::Std;
+use List::Util qw(max);
+
+sub usage {
+  print("Usage: $0 [OPTION]... -i <input>\n");
+  print("Extract the device kernels from an hcc executable.\n\n");
+  print("-h                                \tshow this help message\n");
+  print("-i <input>                        \t[DEPRECATED] input file\n");
+  print("-E <kernel-name>[:<kernel-name>]  \tExtract code object by URI name. Multiple code object names may be specified using \':\' delimiter.\n");
+  print("-L                                \tList code object bundles, by entry name, and their embedded kernel code objects, by URI, available in source input file (-i option).\n");
+  print("-b                                \tList only bundle entries.\n");
+  print("-c <bundle-entry-name>            \tList only embedded code objects for specific code object bundle <bundle-entry-name>\n");
+  print("-o                                \tExtract only binary code objects.  Do not generate disassembled ISA\n");
+  print("-v                                \tVerbose mode.\n");
+  exit;
+}
+
+# my $debug = 1;
+
+my $input_file;
+my @avail_bundles;
+my %bundle_kernels;
+my $extract_range_specifier;
+my $extract_process_id;
+my $extract_file;
+my $extract_offset;
+my $extract_size;
+
+sub print_bundles {
+  for my $bund ( keys %bundle_kernels ) {
+     print("Bundle Entry Name: $bund\n");
+  }
+}
+
+sub list_codeobject {
+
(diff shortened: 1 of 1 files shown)


bin/extract_codeobject
```diff
@@ -0,0 +1,426 @@
...
+use Cwd 'realpath';
+use Getopt::Std;
+use List::Util qw(max);
+
+sub usage {
+  print("Usage: $0 [OPTION]... -i <input>\n");
+  print("Extract the device kernels from an hcc executable.\n\n");
+  print("-h                                \tshow this help message\n");
+  print("-i <input>                        \t[DEPRECATED] input file\n");
+  print("-E <kernel-name>[:<kernel-name>]  \tExtract code object by URI name. Multiple code object names may be specified using \':\' delimiter.\n");
```
t-tye, line 16: kernel-name -> code-object-uri

I am not keen on the : sparator. Would it be better to just allow multiple -E arguments?

Would -u be a better option name as this is specifying a Uri?

bin/extract_codeobject
```diff
@@ -0,0 +1,426 @@
...
+  print("-L                                \tList code object bundles, by entry name, and their embedded kernel code objects, by URI, available in source input file (-i option).\n");
```
t-tye, line 17: List the bundled code objects, and the code objects they contain, that are embedded in the host executable specified by -e.

Would -l be a better name as easier to type and fits other commands that list things?

bin/extract_codeobject
```diff
@@ -0,0 +1,426 @@
...
+use File::Basename;
+use File::Which;
+use Cwd 'realpath';
+use Getopt::Std;
+use List::Util qw(max);
+
+sub usage {
+  print("Usage: $0 [OPTION]... -i <input>\n");
+  print("Extract the device kernels from an hcc executable.\n\n");
+  print("-h                                \tshow this help message\n");
```
t-tye, line 14: show->Show
Add trailing .

bin/extract_codeobject
```diff
@@ -0,0 +1,426 @@
...
+  print("-i <input>                        \t[DEPRECATED] input file\n");
```
t-tye, line 15: Why is this deprecated? Seems it is needed for the -l option.

Would -e be a better name for the host executable?

-e <host-executable> \tSpecify host executable file.

bin/extract_codeobject
```diff
@@ -0,0 +1,426 @@
...
+  print("-E <kernel-name>[:<kernel-name>]  \tExtract code object by URI name. Multiple code object names may be specified using \':\' delimiter.\n");
+  print("-L                                \tList code object bundles, by entry name, and their embedded kernel code objects, by URI, available in source input file (-i option).\n");
+  print("-b                                \tList only bundle entries.\n");
+  print("-c <bundle-entry-name>            \tList only embedded code objects for specific code object bundle <bundle-entry-name>\n");
```
t-tye, line 19: Not sure on the meaning of these. What name is used to identify embedded bundled code objects? The section name they are in?

Does -c list code objects in embedded bundled code objects that match the provided name? Can wild cards be used? Is the main use of this to get the URI of a code object so it can be extracted using -u?

What file names are used for extracted code objects?

It just feels like this is really 2 tools. One that simply extracts code objects by a [list of] URI; and one that takes a [list of] host executabes and displays the embedded code objects. Perhaps this should be two suitably named tools with natural options? It that is done may want to consider an option to output to stdout so can pipe to other tools such as disassembly.

I think having some simple tools would fit better in the Unix style of composable tools.

bin/extract_codeobject
```diff
@@ -0,0 +1,426 @@
...
+  print("-o                                \tExtract only binary code objects.  Do not generate disassembled ISA\n");
```
t-tye, line 20: I think the default should be to only extract the code objects. The user can always use llvm-obdump to get a disassembly. I think it is better to keep this tool only for extracting.

mangupta:

PR #2138, opened 2020-08-07
Introduce the ROCMINFO_PATH variable to override hard-coded rocm_agent_enumerator location

haampie:
In the spack package manager we do not provide a `bin/` or `lib/` folder in the prefix. The problem is that `hipcc` expects the `bin/` folder to exist. This PR introduces a variable `ROCMINFO_PATH` which can be set to the rocminfo package root folder, such that `rocm_agent_enumerator` can still be located.

### DIFF
--- a/bin/hipcc
+++ b/bin/hipcc
@@ -684,7 +684,8 @@ if($HIP_PLATFORM eq "hcc" or $HIP_PLATFORM eq "clang"){
             $targetsStr = $ENV{HCC_AMDGPU_TARGET};
         } else {
             # Else try using rocm_agent_enumerator
-            $ROCM_AGENT_ENUM = "${ROCM_PATH}/bin/rocm_agent_enumerator";
+            $ROCMINFO_PATH = $ENV{'ROCMINFO_PATH'} // $ROCM_PATH;
+            $ROCM_AGENT_ENUM = "${ROCMINFO_PATH}/bin/rocm_agent_enumerator";
             $targetsStr = `${ROCM_AGENT_ENUM} -t GPU`;
             $targetsStr =~ s/\n/,/g;
         }


pfultz2:
Why not have hipcc find `rocm_agent_enumerator` in the `PATH`? Spack will add that package to the `PATH` if its set as a runtime dependency. This seems better than creating another custom env variable.

bin/hipcc
```diff
@@ -684,7 +684,8 @@ if($HIP_PLATFORM eq "hcc" or $HIP_PLATFORM eq "clang"){
             $targetsStr = $ENV{HCC_AMDGPU_TARGET};
         } else {
             # Else try using rocm_agent_enumerator
-            $ROCM_AGENT_ENUM = "${ROCM_PATH}/bin/rocm_agent_enumerator";
+            $ROCMINFO_PATH = $ENV{'ROCMINFO_PATH'} // $ROCM_PATH;
+            $ROCM_AGENT_ENUM = "${ROCMINFO_PATH}/bin/rocm_agent_enumerator";
```
haampie, line 688: @pfultz2  do you mean to change it to

```suggestion
            $ROCM_AGENT_ENUM = "rocm_agent_enumerator";
```
? Or would that break hipcc for people? Otherwise we might have to run `rocm_agent_enumerator` once to see if it works, and if not resort back to `${ROCM_PATH]/bin/rocm_agent_enumerator`?

pfultz2: I am not familiar with perl, but I would think you need to use something like `$(which rocm_agent_enumerator)` but I believe the path will need to be appended as well with `$ROCM_PATH/bin` doing something like `export PATH="$PATH:$ROCM_PATH/bin`.

haampie:
@pfultz2 can we get a change like this in? In spack there's been a patch since 3.5.0 to fix this issue, would be nice to drop it at some point.

mangupta:

PR #2149, opened 2020-08-25
Minor compilation fixes for hipDispatchLatency

baryluk:
```
hipDispatchEnqueueRateMT.cpp:136:10: error: no template named 'vector' in namespace 'std'
    std::vector<std::future<void>> threads;
    ~~~~~^
hipDispatchEnqueueRateMT.cpp:129:18: warning: braces around scalar initializer [-Wbraced-scalar-init]
        shared = {0};
                 ^~~

```

### DIFF
--- a/samples/1_Utils/hipDispatchLatency/hipDispatchEnqueueRateMT.cpp
+++ b/samples/1_Utils/hipDispatchLatency/hipDispatchEnqueueRateMT.cpp
@@ -29,6 +29,7 @@ THE SOFTWARE.
 #include <thread>
 #include <future>
 #include <functional>
+#include <vector>
 
 #define NUM_GROUPS 1
 #define GROUP_SIZE 1
@@ -126,7 +127,7 @@ struct thread_pool {
             thread.get();
         }
         threads.clear();
-        shared = {0};
+        shared.store(0, std::memory_order_seq_cst);
     }
     ~thread_pool() {
         finish();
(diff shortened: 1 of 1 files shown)


samples/1_Utils/hipDispatchLatency/hipDispatchEnqueueRateMT.cpp
```diff
@@ -126,15 +127,15 @@ struct thread_pool {
             thread.get();
         }
         threads.clear();
-        shared = {0};
+        shared.store(0, std::memory_order_seq_cst);
```
emankov, line 130: Could you please explain why sequentially-consistent ordering was chosen here?

emankov (changes requested):

samples/1_Utils/hipDispatchLatency/hipDispatchEnqueueRateMT.cpp
baryluk: It doesn't actually matter here. When this line is executed all other threads that could possibly touch this memory location are already gone. So no synchronization is needed. Any new threads that might possible start in the future will be see any store to `shared` due to implicit synchronization by creation of threads.

However, I just used seq_cst, just for safety and documentation. It is not performance critical, and it is a sample after all anyhow.

The seq_cst is actually a default for `std::atomic_int::store` method.

The shared is only really used by the worker threads to implement a thread id and a barrier, it is not actually really used for any synchronization in the `thread_pool` itself. But it must be reset, to allow new calls to `start`.

emankov (approved):

baryluk:
How I can retrigger CI?

saadrahim:
CI is linking to a server that has been decommissioned. There is no way to trigger CI currently. Can a maintainer of this project update the CI requirement?

mangupta:

PR #2159, opened 2020-09-03
--rpath -> -rpath

haampie:
`-Wl,--rpath` works, but it seems to be an unofficial API.

See https://linux.die.net/man/1/ld for reference.


mangupta:
Closing all existing pull requests that have been in this state for over 6 months. In case this pull request is still valid, please rebase your changes against develop branch and resubmit.

Issue #631, opened 2018-08-15
HIP linker script fails due to empty HCC_HOME in FindHIP.cmake


tdd11235813:
In cmake I use `hip_add_executable` which is defined by [FindHIP.cmake](https://github.com/ROCm-Developer-Tools/HIP/blob/master/cmake/FindHIP.cmake). I only have nvcc platform at the moment. However, a recent change to FindHIP.cmake leads to a wrong call of `hipcc_cmake_linker_helper`.
The `HCC_HOME` is empty, so `hipcc_cmake_linker_helper` mistakenly gets the next argument as the HCC_PATH, which is actually the object files, so nvcc does not get the object files anymore and fails:

```bash
# /usr/bin/cmake -E cmake_link_script CMakeFiles/bufferCopy.dir/link.txt --verbose=1 /libraries/hip/git-nvcc9.1/bin/hipcc_cmake_linker_helper     CMakeFiles/bufferCopy.dir src/bufferCopy_generated_bufferCopy.cpp.o -o bufferCopy -lcudart -lpthread -lrt 
nvcc fatal   : No input files specified; use option --help for more information
```

To make it working again I had to define HCC_HOME outside and before the hip_add_executable macro in FindHIP.cmake:
```cmake
if(NOT DEFINED(HCC_HOME))                
  set(HCC_HOME "${HIP_ROOT_DIR}/bin/hcc")
endif()
###
set(CMAKE_HIP_LINK_EXECUTABLE "${HIP_HIPCC_CMAKE_LINKER_HELPER} ${HCC_HOME} <FLAGS> <CMAKE_CXX_LINK_FLAGS> <LINK_FLAGS> <OBJECTS> -o <TARGET> <LINK_LIBRARIES>")
```
Edit: added last line to code (must set `CMAKE_HIP_LINK_EXECUTABLE` with the new `HCC_HOME`).

mangupta:
That is definitely odd. We already have something similar already in place in FindHIP.cmake. See https://github.com/ROCm-Developer-Tools/HIP/blob/master/cmake/FindHIP.cmake#L557-L559. This should have worked fine in your case. Do you have a modified or an older version of FindHIP.cmake?

tdd11235813:
any news on this? We had to change our cmake code to workaround this issue with HIP, see [cross-ref](https://github.com/ComputationalRadiationPhysics/alpaka/pull/661)

MrBurmark:
I ran into the same issue with rocm 3.5.1.
[...]

PR #2194, opened 2020-11-24
Detect and report empty compiler paths in cmake_linker_helper

joshessman-llnl:
This PR is intended to detect and report empty `HCC_HOME` conditions like the one discussed in #631 and similar conditions where `HIP_CLANG_PATH` is empty.

The added check only attempts to confirm that the first command-line argument is a directory and not that the directory is actually the root of a compiler installation.


mangupta:
Closing all existing pull requests that have been in this state for over 6 months. In case this pull request is still valid, please rebase your changes against develop branch and resubmit.

PR #2213, opened 2021-01-08
Fix memory leak caused by undestroyed hip::Device and hip::Stream.

zhaozhangjian:
1. introduce hip::tearDown() to deallocate hip::Device instances in
g_devices and host_device, which should be issued inside
amd::Runtime::tearDown() in ROCclr;
2. add std::list<Stream*> streams_ to class hip::Device() to record
created hip::Stream instances and deallocate them in the destructor
of hip::Device;
3. define hip::Device::null_stream_ as Stream* type instread of Stream
type to prevent double free in the destructor of hip::Devcie.

### DIFF
--- a/rocclr/hip_context.cpp
+++ b/rocclr/hip_context.cpp
@@ -70,6 +70,15 @@ void init() {
   PlatformState::instance().init();
 }
 
+void tearDown() {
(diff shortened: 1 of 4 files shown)


mangupta:

PR #2218, opened 2021-01-11
Fix patch version number when building from tarball

haampie:
I have git installed on my system, and I'm using spack to build hip, which downloads tarballs without the git repo.

When building, it somehow ends up with incorrect patch version numbers (tried it on 3.9.0 ... 4.0.0).

This patch ensures that `find_package(Git)` will only ever be run when there is in fact a `.git` folder in the project source directory.

Before this patch, I ended up with shared libs like this:

```
$ ls /home/harmen/spack/opt/spack/linux-ubuntu20.04-zen2/gcc-9.3.0/hip-3.9.0-bluiqfy2yt6jobgtn4qkqmdvvrqkghx6/lib/
cmake  hip_prof_str.h  libamdhip64.so  libamdhip64.so.3  libamdhip64.so.3.9.21014-
```

After this patch, things look much better:

```
$ ls /home/harmen/spack/opt/spack/linux-ubuntu20.04-zen2/gcc-9.3.0/hip-3.9.0-4qli74twugr4mbbx7ee5niudq3oiph7x/lib/
cmake  hip_prof_str.h  libamdhip64.so  libamdhip64.so.3  libamdhip64.so.3.9.0
```

In fact previously I coulnd't even use the `hip::amdhip64` target because it would link against `/abs/path/to/libamdhip64.so.3.9.21014-`, which resulted in a linking error in GCC because of the trailing dash.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,7 +52,16 @@ string(REPLACE "." ";" VERSION_LIST ${HIP_BASE_VERSION})
 list(GET VERSION_LIST 0 HIP_VERSION_MAJOR)
 list(GET VERSION_LIST 1 HIP_VERSION_MINOR)
 
-find_package(Git)
+# only look for git when we have a git repo
+if (IS_DIRECTORY "${PROJECT_SOURCE_DIR}/.git")
+  find_package(Git)
+endif()
+
+# FIXME: Two different version strings used.
+
+set(HIP_PACKAGING_VERSION_PATCH "0")
+set(HIP_VERSION_GITDATE "0")
+set(HIP_VERSION_PATCH "0")
 
 # FIXME: Two different version strings used.
 if(GIT_FOUND)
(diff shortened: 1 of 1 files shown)


haampie:
Any chance this gets reviewed / merged? @pfultz2

haampie:
Bump @pfultz2

haampie:
Any new here @pfultz2 / @yxsamliu ?

mangupta:

haampie:
Sorry @mangupta but this is just rude... nobody made an attempt to even look at this PR, and the only response I get is a copy & pasted message that this PR is closed.

cgmb:
Thanks for the report, @haampie. I ran into the same problem when building HIP with some extra profiling flags and I was very confused. Your messages here and on the spack slack helped me understand what was wrong.

I've submitted a fix for this and a couple related issues to hipamd. You can expect the fix to arrive in a future ROCm release. I'm not sure exactly which one, but probably ROCm 5.3.

Issue #2220, opened 2021-01-13
RelWithDebInfo builds are prevented


yalue:
Currently, HIP can only be built using `Debug` or `Release` cmake build types.  This restriction appears to be arbitrary, resulting only from a check in `CMakeLists.txt` in order to set a different default installation path for each build type.  For those of us experimenting with HIP modifications, a RelWithDebInfo build is quite useful for quickly obtaining stack trace information should a HIP application crash, while otherwise allowing the experimental build to replace the pre-built HIP installed to `/opt/rocm*`.

Modifying CMakeLists.txt to set the `RelWithDebInfo` install path to the same location as `Release` builds allows cmake to complete and HIP to install with no apparent problems.

satyanveshd:
HIP can be built using RelWithDebInfo cmake build type and can be installed properly with the recent builds.  This can be verified with Rel 4.1 onwards.

PR #2221, opened 2021-01-13
Enable the RelWithDebInfo cmake build type

yalue:
This change attempts to address issue #2220.  Please let me know if there are other complications reasons for avoiding a RelWithDebInfo build of HIP, as I have been using such a build in my own work and, to my knowledge, have not encountered any problems with it.

As for what is actually changed in this PR:

It modifies a check in CMakeLists.txt, which appeared to be the only location preventing a "RelWithDebInfo" build type.  I tested this change by building HIP using `-DCMAKE_BUILD_TYPE=RelWithDebInfo`, which compiled and installed successfully.  Running this build of HIP appears to cause no problems, either.


haampie:
RelWithDebInfo would be a great default for all rocm packages, would be nice to get stack traces

satyanveshd:
Closing this as the RelWithDebInfo cmake build type is enabled with the recent HIP versions and it can be verified from 4.1 onwards.

PR #2226, opened 2021-01-21
Clarifying that hipExtLaunch timing does not include cache flush time

gilbertlee-amd:
- Providing some clarity on what is omitted from being timed by hipExtLaunch
- Please double-check for correct wording

This arose from misinterpreting that timing differences were due to lower overhead, due to a combined call (instead of sandwiching the kernel launch between hipRecordEvent()),  instead of not actually timing cache flushes.


doctorcolinsmith:
Is there a good person on the HIP team to review this item?  Maybe @AlexVlx or @emankov?

AlexVlx:
@doctorcolinsmith you probably want @bensander or @hthangirala to clarify here, since they were more involved with this particular interface.

mangupta:

PR #2242, opened 2021-03-18
Fix perl and python shebang lines

haampie:
- Do not expect perl to be bin /usr/bin/perl, but use perl in $PATH
- Don't expect python to be in /usr/bin/python but use what is in $PATH

### DIFF
--- a/bin/hipify-perl
+++ b/bin/hipify-perl
@@ -1,4 +1,4 @@
-#!/usr/bin/perl -w
+#!/usr/bin/env perl
 
 ##
 # Copyright (c) 2015-present Advanced Micro Devices, Inc. All rights reserved.
@@ -26,6 +26,7 @@
 
 #usage hipify-perl [OPTIONS] INPUT_FILE
 
+use warnings;
 use Getopt::Long;
 my $whitelist = "";
 my $fileName = "";
--- a/bin/hipify-perl
+++ b/bin/hipify-perl
@@ -1,4 +1,4 @@
-#!/usr/bin/env perl
+#!/usr/bin/perl -w
 
 ##
 # Copyright (c) 2015-present Advanced Micro Devices, Inc. All rights reserved.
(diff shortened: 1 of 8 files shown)


haampie:
Can we still get this into 4.1.0? In spack we build perl as a build dependency, and we have to replace the shebangs by hand as a workaround.

bin/hipify-perl
```diff
@@ -26,6 +26,7 @@
 
 #usage hipify-perl [OPTIONS] INPUT_FILE
 
+use warnings;
```
emankov, line 29: `hipify-perl` is autogenerated script. The changes should be done in hipify-clang first: https://github.com/ROCm-Developer-Tools/HIPIFY. There is the comment above:
`#IMPORTANT: Do not change this file manually: it is generated by hipify-clang --perl`

haampie: thanks!

haampie: In my defense, I did not change it manually :p

emankov: That is why there is excessive `use warnings;` in it?

haampie: I've added those lines `use warnings;` because `/usr/bin/env [command]` only accepts a single executable, not flags :(. Do you see a place where `use warnings;` is repeated / should be dropped?

emankov: Now I understand: `use warnings;` is added to support `-w`, right?

haampie: Yes!

haampie: Ok, I've left hipify-perl as is

haampie:
Should I include the hipify-perl script from https://github.com/ROCm-Developer-Tools/HIPIFY/pull/302 in this PR now that it is merged? Or is that automated?

mangupta:

PR #2292, opened 2021-06-14
Fix race in hip prof api.h

kikimych:
There is racing sequence of calls: T1-api_callbacks_spawner_t(), is_enabled() returns false, T2-api_callbacks_table_t::set_callback(), T2-api_callbacks_table_t::set_activity(),T1-api_callbacks_spawner_t::call(),api_callbacks_spawner_t::~api_callbacks_spawner_t() leading to different return values of is_enabled() function in constructor and destructor. So, hip_cb_table_entry_t.sem for that instance of  api_callbacks_spawner_t is not released. Stored original value of is_enabled() call in instance.


atamazov (changes requested):
This PR consists of 135 commits. Something is terribly wrong. I think that this PR should be re-created using valid baseline. Or re-targeted to different destination branch.

mangupta:

PR #2202, opened 2020-12-07
Remove -lnuma in hipcc

haampie:

PR #2306, opened 2021-08-05
Don't link to numa unconditionally

haampie:
Replaces #2202

### DIFF
--- a/bin/hipcc
+++ b/bin/hipcc
@@ -702,7 +702,7 @@ if ($HIP_PLATFORM eq "amd") {
 
     if (not $isWindows  and not $compileOnly) {
       if ($linkType eq 0) {
-        $toolArgs = " -L$HIP_LIB_PATH -lamdhip64 -L$ROCM_PATH/lib -lhsa-runtime64 -ldl -lnuma " . ${toolArgs};
+        $toolArgs = " -L$HIP_LIB_PATH -lamdhip64 -L$ROCM_PATH/lib -lhsa-runtime64 -ldl " . ${toolArgs};
       } else {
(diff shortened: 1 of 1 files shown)


mangupta:
HIP uses [ROCclr](https://github.com/ROCm-Developer-Tools/ROCclr/) which in turn has a dependency on NUMA (See https://github.com/ROCm-Developer-Tools/ROCclr/blob/main/device/rocm/rocdevice.cpp#L53-L55). HIP packages released by AMD are always built with NUMA support enabled. So hipcc defaults to linking to numa on AMD platform.

haampie:
Ah okay, you mean `libamdrocclr` is always a static library and therefore you have to link in the transient dependency on libnuma?

haampie:
@mangupta I think hipcc never links `libamdrocclr` directly?

As far as I can see from the hip build logs, `libamdhip64.so` links to `libamdrocclr.a` and of course `libnuma.so`:

```
g++ ... -shared -Wl,-soname,libamdhip64.so.4 -o ../lib/libamdhip64.so.4.2.0 ... /path/to/libamdrocclr_static.a -ldl /path/to/libhsa-runtime64.so.1.3.0 -pthread /path/to/libamd_comgr.so.2.0 /usr/lib/x86_64-linux-gnu/librt.so /path/to/lib/libnuma.so 
```

And I don't see hipcc adding `-lamdrocclr_static` flags... so I believe there is no need to add `-lnuma` to every invocation of hipcc.

So I think this PR still makes sense?

Edit: -lnuma would only ever be necessary if you build HIP with `-DBUILD_SHARED_LIBRARIES=off`, but then I suppose it's broken anyways since you're not explicitly linking `-lamdrocclr_static` either...

haampie:
Ah yes, I think I see it now... it's only applied when using the static lib. Sorry for the noise.

PR #2381, opened 2021-10-15
SWDEV-303204 - Disable texture/image Apis on some devices.

TomSang:
Modify texture tests to pass on devices that don't support
texture.
Add hipDeviceAttributeImageSupport to check image support
in runtime
Change-Id: Ia89c494e651a6198a24448b59a91e046a9ebea38

### DIFF
--- a/docs/markdown/hip_kernel_language.md
+++ b/docs/markdown/hip_kernel_language.md
@@ -494,7 +494,11 @@ Following is the list of supported floating-point intrinsics. Note that intrinsi
 | double __dsqrt_rn ( double  x ) <br><sub>Compute `√x` in round-to-nearest-even mode.</sub> |
 
 ## Texture Functions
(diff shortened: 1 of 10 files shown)


mangupta (commented):
@tomsang: Can you please cross-check if the changes to the test source is applicable to the ones in https://github.com/ROCm-Developer-Tools/HIP/tree/develop/tests/catch/unit/texture as well? If so, can you update them as well?

TomSang:
Hi @mangupta   Yes, the changes to the test source is applicable to the ones in Catch2.   But we'd better to update Catch2 test in the later time.    This is because  I have other two patches that depend on this ,  where more existing test cases are updated and a few new test  cases are added.   This is why I hope this PR can be  merged ASAP.

TomSang:
!verify

gargrahul (approved):
LGTM

PR #2396, opened 2021-10-25
SWDEV-296926 - hipcc/hipconfig refactoring from perl to C++ executables

agunashe:
This change updates hipcc and hipconfig  from perl to C++ executables. These exes should be built while building HIP back-end.
Change-Id: I75281253f893d653a99a16b61c12cb56ac0cfd1b


src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+                                "gfx801", "gfx802", "gfx803", "gfx805", "gfx810",
+                                "gfx900", "gfx902", "gfx904", "gfx906", "gfx908", "gfx909", "gfx90a",
+                                "gfx1010", "gfx1011", "gfx1012", "gfx1030", "gfx1031", "gfx1032" };
+
+
+// Known Features
+vector<string> knownFeatures = { "sramecc - " , "sramecc + ", "xnack - ", "xnack + " };
+
+
+enum PlatformType {
```
Kerilk, line 59: We may want to rename this to spirv rather than intel.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+  fs::path full_path(getSelfPath());
+  if (envVariables_.hipPathEnv_.empty())
+    variables_.hipPathEnv_ = (full_path.parent_path()).string();
+  else
+    variables_.hipPathEnv_ = envVariables_.hipPathEnv_;
+}
+
+
+// constructs the ROCM path
+void HipBinBase::constructRoccmPath() {
```
Kerilk, line 333: This seems strangely AMD/roccm specific in a class that should be vendor agnostic. The hipPath part looks fine.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+}
+
+// returns the Roccm path
+const string& HipBinBase::getRoccmPath() const {
+  return variables_.roccmPathEnv_;
+}
+
+
+// returns the Rocclr Home path
+void HipBinBase::constructRocclrHomePath() {
```
Kerilk: This seems strangely AMD/rocclr specific in a class that should be vendor agnostic.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+
+
+
+// returns the Rocclr Home path
+const string& HipBinBase::getRocclrHomePath() const {
+  return variables_.hipRocclrPathEnv_;
+}
+
+// construct hsa Path
+void HipBinBase::constructHsaPath() {
```
Kerilk: This seems strangely AMD/hsa specific in a class that should be vendor agnostic.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+  }
+  catch (...) {
+    value = defaultValue;
+  }
+  return value;
+}
+
+
+// reads the Hip Version
+void HipBinBase::readHipVersion() {
```
Kerilk: Should the version be only dependent on the HIP common repository? Or should it depend on the platform used? Especially if a platform does not support the latest HIP specification?

Kerilk (commented):
My comments are mostly architectural ones due to time constraints on reviewing the changes.
I don't think any of those are blocking issues.

src/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+
+project (hipconfig)
+add_executable(hipconfig hipBin.cpp)
+if (NOT WIN32)   #c++17 does not require the std lib linking
+  target_link_libraries(hipconfig ${LINK_LIBS} ) # for hipconfig
+endif()
+
+set(HIP_VERSION_MAJOR 4 PARENT_SCOPE)
+set(HIP_VERSION_MINOR 4 PARENT_SCOPE)
+set(HIP_VERSION_PATCH 4 PARENT_SCOPE)
```
linehill: Missing newline at the end of the file.

src/hipBin.cpp
```diff
@@ -0,0 +1,267 @@
...
+class HipBinBase;
+class HipBinAmd;
+class HipBinNvidia;
+class HipBin;
+
+
+class HipBin {
+ private:
+  vector<HipBinBase*> hipBinBasePtrs_;
+  vector<PlatformInfo> platformVec_;
```
linehill, line 40: Can't we query platform information via hipBinBasePtrs_?

src/hipBin.cpp
```diff
@@ -0,0 +1,267 @@
...
+HipBin::~HipBin() {
+  delete hipBinNVPtr_;
+  delete hipBinAMDPtr_;
+  hipBinBasePtrs_.clear();  // clearing the vector so no one accesses the pointers
+  platformVec_.clear();     // clearing the platform vector as the pointers are deleted
+}
+
+
+// subtring is present in string
+bool HipBin::substringPresent(string fullString, string subString) const {
```
linehill: Move this out of HipBin class to own utility module or class.

src/hipBin.cpp
```diff
@@ -0,0 +1,267 @@
...
+    for (int i = 1; i < argc; ++i) {
+      HipBinCommand cmd;
+      cmd = gethipconfigCmd(argv[i]);
+      switch (cmd) {
+      case help: platformPtrs.at(j)->printUsage();
+        break;
+      case path: {
+        string hipPath;
+        hipPath = platformPtrs.at(j)->getHipPath();
+        cout << hipPath;
```
linehill:
```suggestion
        cout << platformPtrs[j]->getHipPath();
```
Similar change for the other cases in the below. No need to use at() as j is known to stay within the bounds of container.

src/hipBin.cpp
```diff
@@ -0,0 +1,267 @@
...
+      default:
+        platformPtrs.at(j)->printUsage();
+        break;
+      }
+    }
+  }
+}
+
+
+bool HipBin::checkCmd(const vector<string>& commands, const string& argument) {
```
linehill: Move this out of HipBin class to own utility module or class.

src/hipBin.cpp
```diff
@@ -0,0 +1,267 @@
...
+}
+
+
+// subtring is present in string
+bool HipBin::substringPresent(string fullString, string subString) const {
+  bool found = false;
+  if (fullString.find(subString) != std::string::npos) {
+    found = true;
+  }
+  return found;
```
linehill:
```suggestion
  return fullString.find(subString) != std::string::npos;
```

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+#ifndef SRC_HIPBIN_BASE_H_
+#define SRC_HIPBIN_BASE_H_
+// We haven't checked which filesystem to include yet
+#ifndef INCLUDE_STD_FILESYSTEM_EXPERIMENTAL
```
linehill: This detection logic could be moved to cmake.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+
+// Check for feature test macro for <filesystem>
+#   if defined(__cpp_lib_filesystem)
```
linehill: Thumbs up for the indentation of the preprocessor directives. I like to see this style being used.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+#include <sstream>
+#include <string>
+#include <map>
+#include <fstream>
+#include <regex>
+#include <algorithm>
+#include <vector>
+
+#if defined(_WIN32) || defined(_WIN64)
+  #include "tchar.h"
```
linehill:
```suggestion
#  include "tchar.h"
```
For consistent code style. I prefer this style so IDE's auto-indentation or code-format tools (e.g. clang-format) won't butcher the indentation.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+# define HIPCC_VERBOSE                  "HIPCC_VERBOSE"
+# define HCC_AMDGPU_TARGET              "HCC_AMDGPU_TARGET"
+
+# define HIP_BASE_VERSION_MAJOR     "4"
+# define HIP_BASE_VERSION_MINOR     "4"
+# define HIP_BASE_VERSION_PATCH     "0"
+# define HIP_BASE_VERSION_GITHASH   "0"
+
+// Known HIP target names.
+vector<string> knownTargets = { "gfx700", "gfx701", "gfx702", "gfx703", "gfx704", "gfx705",
```
linehill: Move this to hipBin_amd.h.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+                                "gfx801", "gfx802", "gfx803", "gfx805", "gfx810",
+                                "gfx900", "gfx902", "gfx904", "gfx906", "gfx908", "gfx909", "gfx90a",
+                                "gfx1010", "gfx1011", "gfx1012", "gfx1030", "gfx1031", "gfx1032" };
+
+
+// Known Features
+vector<string> knownFeatures = { "sramecc - " , "sramecc + ", "xnack - ", "xnack + " };
```
linehill: Move this to hipBin_amd.h.

src/hipBin_base.h
linehill, line 59: Is the enumeration necessary? I only see it being converted to std::string. Could we have a virtual HipBinBase::getPlatformName() method instead? This would move the HIP-common towards the platform/backend agnostic direction. I have similar though on the CompilerType and RuntimeType enumerations too.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+  void constructHipPath();
+  void constructRoccmPath();
+  void constructHsaPath();
+  void readHipVersion();
+  const string& getHipPath() const;
+  const string& getRoccmPath() const;
+  const string& getHsaPath() const;
+  const string& getHipVersion() const;
+  void constructRocclrHomePath();
+  const string& getRocclrHomePath() const;
```
linehill: Relocate these to the HipBinAmd class.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+  const EnvVariables& getEnvVariables() const;
+  void readOSInfo();
+  const OsType& getOSInfo() const;
+  void constructHipPath();
+  void constructRoccmPath();
+  void constructHsaPath();
+  void readHipVersion();
+  const string& getHipPath() const;
+  const string& getRoccmPath() const;
+  const string& getHsaPath() const;
```
linehill: Relocate these to the HipBinAmd class.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+  string chomp(string str) const;
+  void getSystemInfo() const;
+  void printEnvirnomentVariables() const;
+  void readEnvVariables();
+  const EnvVariables& getEnvVariables() const;
+  void readOSInfo();
+  const OsType& getOSInfo() const;
+  void constructHipPath();
+  void constructRoccmPath();
+  void constructHsaPath();
```
linehill: Move these to HipBinAmd class.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+  const string& getRoccmPath() const;
+  const string& getHsaPath() const;
+  const string& getHipVersion() const;
+  void constructRocclrHomePath();
+  const string& getRocclrHomePath() const;
+  string readConfigMap(map<string, string> hipVersionMap, string keyName, string defaultValue) const;
+  void printUsage() const;
+  bool canRun(string exeName, string& cmdOut);
+  map<string, string> parseConfigFile(fs::path configPath) const;
+  bool substringPresent(string fullString, string subString) const;
```
linehill: Duplication of HipBin::substringPresent. Also, move this to own utility module/class.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+struct SystemCmdOut {
+  string out;
+  int exitCode = 0;
+};
+
+
+class HipBinBase {
+ public:
+  HipBinBase();
+  virtual ~HipBinBase() {}
```
linehill: Call deleteTempFiles() here if the lifetime of the temp files do not need to be extended beyond the HipBinBase instance.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+void HipBinBase::deleteTempFiles() {
+  try {
+    // Deleting temp files vs the temp directory
+    for (unsigned int i = 0; i < tmpFiles_.size(); i++) {
+      fs::remove(tmpFiles_.at(i));
+    }
+  }
+  catch(...) {
+    for (unsigned int i = 0; i < tmpFiles_.size(); i++) {
+      cout << "temp tmpFiles_ name: "<< tmpFiles_.at(i) <<endl;
```
linehill: Remove debug printing.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+    isFile.close();
+  }
+  return configMap;
+}
+
+// Delete all created temporary files
+void HipBinBase::deleteTempFiles() {
+  try {
+    // Deleting temp files vs the temp directory
+    for (unsigned int i = 0; i < tmpFiles_.size(); i++) {
```
linehill: The for-loop could be hoisted out the try-catch clause so we keep trying to delete the other temp files when one deletion attempt fails. Now it does not matter as the temp files are created within the same directory(?).

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+class HipBinBase {
+ public:
+  HipBinBase();
+  virtual ~HipBinBase() {}
+  // Common helper functions
+  std::string getSelfPath() const;
+  vector<string> splitStr(string fullStr, char delimiter) const;
+  string replaceStr(const string& s, const string& toReplace, const string& replaceWith) const;
+  string replaceRegex(const string& s, std::regex toReplace, string replaceWith) const;
+  SystemCmdOut exec(const char* cmd, bool printConsole) const;
```
linehill: Relocate these to own utility module or class. Do the same for all the other non-virtual methods which do not depend on the state of this class.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+    }
+  }
+}
+
+// Create a new temporary directory and return it
+string HipBinBase::getTempDir() {
+  // mkdtemp is only applicable for unix and not windows.
+  // Using filesystem becasuse of windows limitation
+  string tmpdir = fs::temp_directory_path().string();
+  tmpDirs_.push_back(tmpdir);
```
linehill: I'm not seeing the tmpDirs_ being referenced anywhere else. Perhaps something was forgotten.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+    #else
+      FILE* pipe = popen(cmd, "r");
+    #endif
+    if (!pipe) throw std::runtime_error("popen() failed!");
+    try {
+      while (fgets(buffer, sizeof buffer, pipe) != NULL) {
+        result += buffer;
+      }
+    } catch (...) {
+      cout << "Error while executing the command: " <<cmd <<endl;
```
linehill:
```suggestion
      cout << "Error while executing the command: " << cmd << endl;
```

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+  }
+  catch(...) {
+    sysOut.exitCode = -1;
+  }
+  return sysOut;
+}
+
+
+// compiler canRun or not
+bool HipBinBase::canRun(string exeName, string& cmdOut) {
```
linehill: Rename to something that reflects the comment above - e.g. canRunCompiler.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+    executable = false;
+  } else {
+    string myline;
+    std::ifstream fp;
+    fp.open(tmpFileName);
+    if (fp.is_open()) {
+      while (fp) {
+        std::getline(fp, myline);
+        cmdOut += myline;
+      }
```
linehill:
```suggestion
      while (std::getline(fp, myline)) {
        cmdOut += myline;
      }
```

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+      fs::path exePath(path);
+      path = exePath.parent_path().string();
+    }
+  #endif
+  return path;
+}
+
+
+// removes the empty spaces and end lines
+string HipBinBase::chomp(string str) const {
```
linehill: I'd prefer this method being named as 'trim'.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+}
+
+// matches the pattern in the string
+bool HipBinBase::stringRegexMatch(string fullString, string pattern) const {
+  bool found = false;
+  std::regex e(pattern);
+  if (std::regex_match(fullString, e)) {
+    found = true;
+  }
+  return found;
```
linehill:
```suggestion
  return std::regex_match(fullString, std::regex(pattern));
```

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+  string intermediate;
+  while (getline(check1, intermediate, delimiter)) {
+    tokens.push_back(intermediate);
+  }
+  return tokens;
+}
+
+
+// replaces the toReplace string with replaceWith string. Returns the new string
+string HipBinBase::replaceStr(const string& s, const string& toReplace, const string& replaceWith) const {
```
linehill: Overextended lines? Check ROCm developer's style guide.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+// returns the value of the key from the Map passed
+string HipBinBase::readConfigMap(map<string, string> hipVersionMap, string keyName, string defaultValue) const {
+  string value;
+  try {
+    value = hipVersionMap.at(keyName);
+  }
+  catch (...) {
+    value = defaultValue;
+  }
+  return value;
```
linehill:
```suggestion
  auto it = hipVersionMap.find(keyName);
  if (it != hipVersionMap.end()) {
    return it->second;
  }
  return defaultValue;
```

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+};
+
+HipBinBase::HipBinBase() {
+  readOSInfo();                 // detects if windows or linux
+  readEnvVariables();           // reads the envirnoment variables
+  constructHipPath();           // constructs HIP Path
+  constructRoccmPath();         // constructs Roccm Path
+  constructRocclrHomePath();    // constructs RocclrHomePath
+  constructHsaPath();           // constructs hsa path
+  readHipVersion();             // stores the hip version
```
linehill: Are these methods only meant to be called by the constructor? If so, I'd put them in private scope.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+  string ldLibraryPathEnv_ = "";
+  string verboseEnv_ = "";
+  string hipccCompileFlagsAppendEnv_ = "";
+  string hipccLinkFlagsAppendEnv_ = "";
+  string hipLibPathEnv_ = "";
+  string deviceLibPathEnv_ = "";
+  string hipClangHccCompactModeEnv_ = "";
+  string hipCompileCxxAsHipEnv_ = "";
+  string hccAmdGpuTargetEnv_ = "";
+  void operator =(const EnvVariables& var) {
```
linehill: If I'm not missing anything, then this method is same as the default assignment operator. This can be deleted.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+    TSTR wide = TSTR(buffer).substr(0, pos);
+    path = std::string(wide.begin(), wide.end());
+  #else
+    char buff[MAX_PATH_CHAR];
+    ssize_t len = ::readlink("/proc/self/exe", buff, sizeof(buff) - 1);
+    if (len != -1) {
+      buff[len] = '\0';
+      path = std::string(buff);
+      fs::path exePath(path);
+      path = exePath.parent_path().string();
```
linehill: This does not take account that the result from the readlink() is potentially truncated (see: https://linux.die.net/man/2/readlink). The truncation case has to be handled in some way (e.g. throw an error or retry with larger buffer).

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+    out = std::regex_replace(out, toReplace, replaceWith);
+  }
+  return out;
+}
+
+// prints the envirnoment variables
+void HipBinBase::printEnvirnomentVariables() const {
+  const OsType& os = getOSInfo();
+  if (os == windows) {
+    cout << "PATH=" << envVariables_.path_ << "\n" << endl;
```
linehill: Looks like a debug printing. Remove it if so.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+  vector<string> splitStr(string fullStr, char delimiter) const;
+  string replaceStr(const string& s, const string& toReplace, const string& replaceWith) const;
+  string replaceRegex(const string& s, std::regex toReplace, string replaceWith) const;
+  SystemCmdOut exec(const char* cmd, bool printConsole) const;
+  string getTempDir();
+  void deleteTempFiles();
+  string mktempFile(string name);
+  string chomp(string str) const;
+  void getSystemInfo() const;
+  void printEnvirnomentVariables() const;
```
linehill:
```suggestion
  void printEnvironmentVariables() const;
```
Typo here.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+  fileName = _mktemp(&name[0]);
+# else
+  fileName = mktemp(&name[0]);
+# endif
+  tmpFiles_.push_back(fileName);
+  return fileName;
+}
+
+// gets the path of the executable name
+std::string HipBinBase::getSelfPath() const {
```
linehill: Style inconsistency: Some methods has `std::` while others omit it. Pick either style.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+// prints system information
+void HipBinBase::getSystemInfo() const {
+  const OsType& os = getOSInfo();
+  if (os == windows) {
+    cout << endl << "== Windows Display Drivers" << endl;
+    cout << "Hostname      :";
+    system("hostname");
+    system("wmic path win32_VideoController get AdapterCompatibility,"
+    "InstalledDisplayDrivers,Name | findstr /B /C:\"Advanced Micro Devices\"");
+  } else {
```
linehill:
```suggestion
  } else {
    assert(os == linux);
```
Let's add this if this branch is meant to work in linux (compatible) OS.

src/hipBin_base.h
```diff
@@ -0,0 +1,869 @@
...
+string HipBinBase::replaceRegex(const string& s, std::regex toReplace, string replaceWith) const {
+  std::string out = s;
+  while (std::regex_search(out, toReplace)) {
+    out = std::regex_replace(out, toReplace, replaceWith);
+  }
+  return out;
+}
+
+// prints the envirnoment variables
+void HipBinBase::printEnvirnomentVariables() const {
```
linehill: This method only prints the contents of the PATH variable. I'd have expected this to print the other variables listed in EnvVariables too.

linehill (commented):
I’d prefer to see a better way to incorporate the HIP-Common into the backend/platform implementations without needing to patch the HIP-Common. FWIW, the HIP-Common would be configurable. For example, a backend could be hooked into the HIP-Common as followed:
```
HipBin::HipBin() {
  hipBinPtr_ = @HIPBACKEND@;
  if (hipBinPtr_->detectPlatform()) {
    ...
}
```
Where cmake substitutes the @HIPBACKEND@ to a HipBinBase instance provided by the backend.

I also prefer to have the vendor specific details in this patch to be moved to their respective backend/platform repositories (e.g. relocate hipBin_nvidia.h and hipBin_amd.h files to HIPAMD).

More comments and suggestions are inlined.

Note, that I skipped review on the hipBin_nvidia.h and the hipBin_amd.h files as, IMO, they belong to the backend side.

src/hipBin_base.h
agunashe: I have removed the intel platform from the enum and added a comment to add new platforms.

src/hipBin_base.h
agunashe: Currently roccmPath is used in nvidia hipconfig too, so we still have some dependency. This needs more discussion.

src/hipBin_base.h
agunashe: rocclrHomePath is now moved to AMD specific class.

src/hipBin_base.h
agunashe: HsaPath is now moved to AMD specific class

src/hipBin_base.h
agunashe: Since we currently go through hip for all platforms, we would need the hipVersion information for all.

src/CMakeLists.txt
agunashe: Fixed in the latest update.

src/hipBin.cpp
agunashe: hipBinBasePtrs_ is a vector which gets populated with the valid platform pointers.  The platform information can be queried once the vector is populated

src/hipBin.cpp
agunashe: fixed in the latest update

src/hipBin.cpp
agunashe: I prefer to use at as its more safe during a out of range indexing - see https://thispointer.com/c-how-to-get-element-by-index-in-vector-at-vs-operator/

src/hipBin.cpp
agunashe: fixed in the latest update

src/hipBin.cpp
agunashe: fixed in the latest update

src/hipBin_base.h
agunashe: Preferred to leave this in the header file as the experimental filesystem/ filesystem is only used in this binary.

src/hipBin_base.h
agunashe: unfortunately this style of indentation does not go along with other hip code. So fixed this to match the indentation style of other hip code

src/hipBin_base.h
agunashe: fixed to match with other hip code

src/hipBin_base.h
agunashe: fixed in the latest update

src/hipBin_base.h
agunashe: fixed in the latest update

src/hipBin_base.h
agunashe: I prefer to leave this enum logic as all the information related to the platform are in one place bundled together in struct PlatformInfo. When we have more platforms we can think of moving it to the appropriate classes

src/hipBin_base.h
agunashe: fixed in the latest update

src/hipBin_base.h
agunashe: fixed in the latest update

src/hipBin_base.h
agunashe: fixed in the latest update

src/hipBin_base.h
agunashe: fixed in the latest update. Created a new class HipBinUtil which will hold all the utility functions for all clients.

src/hipBin_base.h
agunashe: fixed in the latest  update. Moved the deleteTempFiles to the destructor of HipBinUtil class

src/hipBin_base.h
agunashe: This is more an error message to point out issue while removing.

src/hipBin_base.h
agunashe: Fixed in the latest update. Moved to the for loop outside the try catch so that the deletion continues

src/hipBin_base.h
agunashe: Moved all the utility functions to a new class HipBinUtil

src/hipBin_base.h
agunashe: tmpDirs_ was not used anywhere. removed in the latest update

src/hipBin_base.h
agunashe: fixed in the latest update

src/hipBin_base.h
agunashe: Fixed in the latest update

src/hipBin_base.h
agunashe: Fixed in the latest update

src/hipBin_base.h
agunashe: Fixed in the latest update

src/hipBin_base.h
agunashe: thanks. Fixed in the latest update

src/hipBin_base.h
agunashe: Fixed in the latest update

src/hipBin_base.h
agunashe: Thanks. Fixed in the latest update

src/hipBin_base.h
agunashe: Moved to private -fixed in the latest update

src/hipBin_base.h
agunashe: thanks, removed the assignment overloading since its the same as default

src/hipBin_base.h
agunashe: throwing an exception with a message incase of issues

src/hipBin_base.h
agunashe: this is part of hipconfig all function call, not a debug printing

src/hipBin_base.h
agunashe: Fixed in the latest update

src/hipBin_base.h
agunashe: fixed in the latest update

src/hipBin_base.h
agunashe: added the assert statement in the latest update

src/hipBin_base.h
agunashe: this prints all hip related variables too and not just PATH variable

agunashe:
> ore comments and suggestions are inlined.
> 
> Note, that I skipped review on the hipBin_nvidia.h and the hipBin_amd.h files as, IMO, they belong to the backend side.



> I’d prefer to see a better way to incorporate the HIP-Common into the backend/platform implementations without needing to patch the HIP-Common. FWIW, the HIP-Common would be configurable. For example, a backend could be hooked into the HIP-Common as followed:
> 
> ```
> HipBin::HipBin() {
>   hipBinPtr_ = @HIPBACKEND@;
>   if (hipBinPtr_->detectPlatform()) {
>     ...
> }
> ```
> 
> Where cmake substitutes the @HIPBACKEND@ to a HipBinBase instance provided by the backend.
> 
> I also prefer to have the vendor specific details in this patch to be moved to their respective backend/platform repositories (e.g. relocate hipBin_nvidia.h and hipBin_amd.h files to HIPAMD).
> 
> More comments and suggestions are inlined.
> 
> Note, that I skipped review on the hipBin_nvidia.h and the hipBin_amd.h files as, IMO, they belong to the backend side.


This needs more discussion internally. For now, this will reside in the common class.  In the future we can move it to cmake file as desired.

agunashe:
This has been moved to a new project - https://github.com/ROCm-Developer-Tools/HIPCC.
Closing this PR

PR #2432, opened 2021-12-09
SWDEV-306204 - Update HIP doxygen and documents

jujiang-del:
Change-Id: I796eda5cb14dfd00d2a806c57c0550356fbbcdb2

### DIFF
--- a/docs/markdown/hip_faq.md
+++ b/docs/markdown/hip_faq.md
@@ -42,7 +42,7 @@ HIP provides the following:
 - Memory management (hipMalloc(), hipMemcpy(), hipFree(), etc.)
 - Streams (hipStreamCreate(),hipStreamSynchronize(), hipStreamWaitEvent(),  etc.)
 - Events (hipEventRecord(), hipEventElapsedTime(), etc.)
-- Kernel launching (hipLaunchKernel is a standard C/C++ function that replaces <<< >>>)
+- Kernel launching (hipLaunchKernel/hipLaunchKernelGGL is the preferred way of launching kernels. hipLaunchKernelGGL is a standard C/C++ macro that can serve as an alternative way to launch kernels, replacing the CUDA triple-chevron (<<< >>>) syntax).
 - HIP Module API to control when adn how code is loaded.
 - CUDA-style kernel coordinate functions (threadIdx, blockIdx, blockDim, gridDim)
 - Cross-lane instructions including shfl, ballot, any, all
--- a/docs/markdown/hip_kernel_language.md
+++ b/docs/markdown/hip_kernel_language.md
@@ -31,7 +31,6 @@
 - [Surface Functions](#surface-functions)
 - [Timer Functions](#timer-functions)
 - [Atomic Functions](#atomic-functions)
-  * [Caveats and Features Under-Development:](#caveats-and-features-under-development)
 - [Warp Cross-Lane Functions](#warp-cross-lane-functions)
   * [Warp Vote and Ballot Functions](#warp-vote-and-ballot-functions)
   * [Warp Shuffle Functions](#warp-shuffle-functions)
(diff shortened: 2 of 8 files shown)


gargrahul (dismissed):

docs/markdown/hip_faq.md
```diff
@@ -42,7 +42,7 @@ HIP provides the following:
 - Memory management (hipMalloc(), hipMemcpy(), hipFree(), etc.)
 - Streams (hipStreamCreate(),hipStreamSynchronize(), hipStreamWaitEvent(),  etc.)
 - Events (hipEventRecord(), hipEventElapsedTime(), etc.)
-- Kernel launching (hipLaunchKernel is a standard C/C++ function that replaces <<< >>>)
+- Kernel launching (hipLaunchKernel/hipLaunchKernelGGL is the right API to call. hipLaunchKernelGGL is a standard C/C++ micro that can serve as an alternative way to launch kernek, and it can replace <<< >>>, if the user so desires).
```
Maetveis:
```suggestion
- Kernel launching (hipLaunchKernel/hipLaunchKernelGGL is the preferred way of launching kernels. hipLaunchKernelGGL is a standard C/C++ macro that can serve as an alternative way to launch kernels, replacing the CUDA triple-chevron (`<<< >>>`) syntax).
```
typo fixes and small a tweak in wording

docs/markdown/hip_kernel_language.md
```diff
@@ -708,12 +686,25 @@ The Cuda `__prof_trigger()` instruction is not supported.
 
 ## Assert
 
-The assert function is under development.
-HIP does support an "abort" call which will terminate the process execution from inside the kernel.
+The assert function is supported in HIP.
+There are two kinds of implementations for assert functions depending on the use sceneries,
+- One is only for the host version of assert, which is defined in assert.h
+- Another is the device version of assert, which is implemented in hip/amd_detail/hip_runtime.h.
+In HIP application, users should include the header file(s) accordingly while make assert function calls.
```
Maetveis: This is simply not true: hip/amd_detail/hip_runtime.h does not define `assert()` it defines `__assert_fail` which is called by the C library provided `assert` macro when the condition is false. It is valid to include `assert.h` in "device" code.

```cuda
#include <hip/hip_runtime.h>
#include <assert.h>

__global__ void f(int* ptr){
   assert(ptr);
   *ptr = 0;
}
```

docs/markdown/hip_programming_guide.md
```diff
@@ -3,17 +3,19 @@
...
 
 ### Memory allocation flags
-hipHostMalloc always sets the hipHostMallocPortable and hipHostMallocMapped flags. Both usage models described above use the same allocation flags, and the difference is in how the surrounding code uses the host memory.
-
-hipHostMallocNumaUser is the flag to allow host memory allocation to follow numa policy set by user.
+There are flags parameter which can specify options how to allocate the memory, for example,
+hipHostMallocPortable, the memory is considered allocated by all contexts, not just the one on which the allocation is made.
+hipHostMallocMapped, will map the allocation into the address space for the current device, and the device pointer can be obtained with the API hipHostGetDevicePointer().
+hipHostMallocNumaUser is the flag to allow host memory allocation to follow numa policy by user.
+All allocation flags are independent, and can be used with combination without restriction, for instance, hipHostMalloc can be called with both hipHostMallocPortable and hipHostMallocMapped flags set. Both usage models described above use the same allocation flags, and the difference is in how the surrounding code uses the host memory.
```
Maetveis:
```suggestion
All allocation flags are independent, and can be used in any combination without restriction, for instance, hipHostMalloc can be called with both hipHostMallocPortable and hipHostMallocMapped flags set. Both usage models described above use the same allocation flags, and the difference is in how the surrounding code uses the host memory.
```

Maetveis (commented):
Sorry for being nosy. I was just browsing around and found a few typos and inaccuracies.

docs/markdown/hip_kernel_language.md
yxsamliu: You are right. the document should be updated as something like:

Users need to include assert.h to use assert. For assert to work in both device and host functions, users need to include "hip/hip_runtime.h".

docs/markdown/hip_faq.md
jujiang-del: Thanks for the review. Updated.

docs/markdown/hip_kernel_language.md
jujiang-del: Thanks. Updated.

docs/markdown/hip_programming_guide.md
jujiang-del: Thanks. Updated.

PR #2446, opened 2021-12-21
SWDEV-314867 - Fix clang missing issue in cuda docker

TomSang:
Cuda dockers doesn't contain clang. The dependence on clang
will fail catch2 building. Catch2 building actually doesn't
need clang. Remove clang dependence from catch2 cmake file.

Change-Id: I8fd76e99d27ce4f5769baffae0608dd1cf0f4188


TomSang:
!psdb

TomSang:
!hip

TomSang:
@mangupta @gargrahul 
This has been pending for a very long time.
Is it still necessary ?
Doesn't any other developer meet the same issue ?

gargrahul (approved):

PR #2473, opened 2022-02-05
SWDEV-321082 - Windows passing GPU arch list from command line

agunashe:
Currently HIP tests depends on HCC_AMDGPU_TARGET for the arch list in
Windows. This change adds the capability to build HIP tests for any arch

Change-Id: Ifb10a9c343cf36c1db22433119addef1a358baab

### DIFF
--- a/tests/catch/CMakeLists.txt
+++ b/tests/catch/CMakeLists.txt
@@ -116,6 +116,17 @@ if(NOT DEFINED OFFLOAD_ARCH_STR AND EXISTS "${ROCM_PATH}/bin/rocm_agent_enumerat
     endif()
 endif()
 
+# In windows get the gpu targets list from cmd line
+if (HIP_PLATFORM STREQUAL "amd" AND NOT UNIX)
+    if (DEFINED GPU_ARCH_LIST)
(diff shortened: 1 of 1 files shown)


tests/catch/CMakeLists.txt
```diff
@@ -116,6 +116,17 @@ if(NOT DEFINED OFFLOAD_ARCH_STR AND EXISTS "${ROCM_PATH}/bin/rocm_agent_enumerat
     endif()
 endif()
 
+# In windows get the gpu targets list from cmd line
+if (HIP_PLATFORM STREQUAL "amd" AND NOT UNIX)
```
mangupta: Why limit this to only windows? We could make use of this infrastructure on linux as well.

mangupta (changes requested):

tests/catch/CMakeLists.txt
agunashe: removed NOT UNIX check. Using the command line GPU_ARCH_LIST for all OS

agunashe:
This PR is not required , architecture can be passed to hip tests from cmd line  like -DOFFLOAD_ARCH_STR="--offload-arch=gfx803 --offload-arch=gfx908"

PR #2489, opened 2022-02-15
SWDEV-294375 - Adding coalesced_groups HIP tests

sourabhuday:
Change-Id: I000a402e32d6057bab11829af2b78931fcd67fb5

HIP unit test cases for ROCm cooperative groups sub-feature, coalesced_groups.


mangupta (changes requested):
@sourabhuday : Can we also get the catch2 equivalent of these tests? Or do you plan to handle that as a separate PR?
CC: @gargrahul

sourabhuday:
@mangupta I will handle it in a separate PR.

PR #2517, opened 2022-02-22
Updated negative tests for hipStreamGetPriority

FMarno:
I've also added a macro to hip_test_common since I plan to use it in other tests.

### DIFF
--- a/tests/catch/unit/stream/hipStreamGetPriority.cc
+++ b/tests/catch/unit/stream/hipStreamGetPriority.cc
@@ -16,24 +16,18 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 */
-/**
+
+/*
 Testcase Scenarios :
 1) Negative tests for hipStreamGetPriority api.
 2) Create stream and check default priority of stream is within range.
 3) Create stream with high or low priority and check priority is set as expected.
-4) Create stream with higher priority or lower priority for the priority range returned.
+4) Create stream with higher priority or lower priority for the priority range returned, the stream
+priority should be clamped to the priority range.
 5) Create stream with CUMask and check priority is returned as expected.
 */
 
 #include <hip_test_common.hh>
-
-/**
- * Check the error returned when a null stream is used.
- */
-TEST_CASE("Unit_hipStreamGetPriority_NullStream") {
-  int priority = 0;
-  HIP_CHECK_ERROR(hipStreamGetPriority(nullptr, &priority), hipErrorInvalidResourceHandle);
-}
 
 /**
  * Check the error returned when an invalid pointer to a priority is used.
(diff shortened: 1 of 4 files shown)


mangupta:
@FMarno : Could you rebase this on top of latest develop. After merging in #2560, this PR has merge conflicts.

tests/catch/unit/stream/hipStreamGetPriority.cc
```diff
@@ -28,46 +28,62 @@ Testcase Scenarios :
...
+    REQUIRE(priority_low >= priority);
+    REQUIRE(priority >= priority_high);
+  }
+  SECTION("High Priority") {
+    HIP_CHECK(hipStreamCreateWithPriority(&stream, hipStreamDefault, priority_high));
+    HIP_CHECK(hipStreamGetPriority(stream, &priority));
+    REQUIRE(priority == priority_high);
+  }
+  SECTION("Higher Priority") {
+    HIP_CHECK(hipStreamCreateWithPriority(&stream, hipStreamNonBlocking, priority_high - 1));
```
chrispaquot: check that priority_high isn't equal to priority_low

tests/catch/unit/stream/hipStreamGetPriority.cc
```diff
@@ -28,46 +28,62 @@ Testcase Scenarios :
...
+    HIP_CHECK(hipStreamGetPriority(stream, &priority));
+    REQUIRE(priority == priority_high);
+  }
+  SECTION("Low Priority") {
+    HIP_CHECK(hipStreamCreateWithPriority(&stream, hipStreamDefault, priority_low));
+    HIP_CHECK(hipStreamGetPriority(stream, &priority));
+    REQUIRE(priority_low == priority);
+  }
+  SECTION("Lower Priority") {
+    HIP_CHECK(hipStreamCreateWithPriority(&stream, hipStreamNonBlocking, priority_low + 1));
```
chrispaquot: same here

tests/catch/unit/stream/hipStreamGetPriority.cc
FMarno: I don't understand why this is relevant? This test checks that priority to clamped to the range of priorities.
If priorities aren't supported then priority_high will equal priority_low which will equal 0

tests/catch/unit/stream/hipStreamGetPriority.cc
```diff
@@ -56,35 +50,43 @@ TEST_CASE("Unit_hipStreamGetPriority_happy") {
...
+      REQUIRE(priority_low >= priority);
+      REQUIRE(priority >= priority_high);
+    }
+    SECTION("High Priority") {
+      HIP_CHECK(hipStreamCreateWithPriority(&stream, hipStreamDefault, priority_high));
+      HIP_CHECK(hipStreamGetPriority(stream, &priority));
+      REQUIRE(priority == priority_high);
+    }
+    SECTION("Higher Priority") {
+      HIP_CHECK(hipStreamCreateWithPriority(&stream, hipStreamNonBlocking, priority_high - 1));
```
mangupta, line 74: Not sure what is the expected behavior if highest priority is 0? Then priority_high - 1 would be equal to -1. I think this test might fail.

FMarno: the expected behavior is for the value to be clamped to the priority range, so the resulting stream will have a priority of 0 in your example. The nvidia machine i tested had `priority_high` of -5 and the amd one had `priority_high` of -1, so I think this should be fine.

tests/catch/unit/stream/hipStreamGetPriority.cc
chrispaquot: Which doesn't work because then you'll -1 to the API expecting it to be "fine"

tests/catch/unit/stream/hipStreamGetPriority.cc
chrispaquot: Now it works that way but I don't think priorities are defined as such. Someone can choose low=0=high so expecting -1 to be fine is not really fine.

FMarno: Sorry, I'm not sure I understand the problem.

For CUDA, specifying a priority outside the range of `priority_high` to `priority_low` will result in the priority of the created stream to be clamped to that range. This is documented [here](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1gfdb79818f7c0ee7bc585648c91770275)  in the description for `cudaDeviceGetStreamPriorityRange` and [here](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__STREAM.html#group__CUDART__STREAM_1ge2be9e9858849bf62ba4a8b66d1c3540) for `cudaStreamCreateWithPriority`.

I can find no documentation mentioning priority values should be greater than or equal to zero. All the apis seems to use a signed int value for priority and, as I mentioned, I have found the devices I have access to have a `priority_high` of less than 0.

The only issue I can foresee is if `priority_high` is equal to `std::numeric_limits<int>::min`, leading to underflow.

chrispaquot: OK I missed the clamping part. Thanks!

FMarno:
@chrispaquot is anything blocking this now? I see that one of the checks have failed but I'm not sure of the details on that.

chrispaquot (approved):

PR #2523, opened 2022-02-27
Remove reference to nonexistent device property

sethrj:
The `hipDeviceProps` attribute `executionUnitsPerMultiprocessor` would be useful as documented, but it does not exist nor does it even have a record in the Git history.


sethrj:
Looks like the HIP documentation might've moved into a private repo (see https://github.com/RadeonOpenCompute/ROCm/discussions/2169). @saadrahim for what it's worth, the [new AMD documentation  page](https://docs.amd.com/projects/HIP/en/latest/reference/kernel_language.html#porting-from-cuda-launch-bounds) still references this nonexistent attribute: it would be great if you could fix this.

PR #2548, opened 2022-03-14
EXSWCPHIPT-139 - Added tests for hipStreamDestroy

FMarno:
CMakeLists.txt was updated so any changes to config_amd_*.json copied when the project was rebuilt.

Unit_hipStreamDestroy_DoubleDestroy fails, but this could just be undefined behaviour.

Unit_hipStreamDestroy_WithPendingWork was found to hang, I think this is due to hipStreamDestroy not synchronizing before destroying the stream, but I'm not sure what the intended functionality is.

Depends on #2517

### DIFF
--- a/tests/catch/unit/stream/hipStreamDestroy.cc
+++ b/tests/catch/unit/stream/hipStreamDestroy.cc
@@ -1,0 +1,97 @@
+/*
+Copyright (c) 2022 Advanced Micro Devices, Inc. All rights reserved.
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+#include <chrono>
+#include <hip_test_common.hh>
+
+namespace hipStreamDestroyTests {
+
+TEST_CASE("Unit_hipStreamDestroy_Default") {
+  hipStream_t stream{};
+  HIP_CHECK(hipStreamCreate(&stream));
+  HIP_CHECK(hipStreamDestroy(stream));
+}
+
+TEST_CASE("Unit_hipStreamDestroy_Negative_DoubleDestroy") {
+  hipStream_t stream{};
+  HIP_CHECK(hipStreamCreate(&stream));
+  HIP_CHECK(hipStreamDestroy(stream));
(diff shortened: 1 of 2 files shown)


tests/catch/unit/stream/hipStreamDestroy.cc
```diff
@@ -0,0 +1,131 @@
...
+__global__ void waiting_kernel(int* semaphore) {
+  size_t tid = blockIdx.x * blockDim.x + threadIdx.x;
+  if (tid == 0) {
+    while (atomicCAS(semaphore, 1, 2) == 0) {
+    }
+  }
+}
+
+// NOTE possible integration test
+TEST_CASE("Unit_hipStreamDestroy_WithPendingWork") {
```
chrispaquot, line 74: I don't like using timers like that as the system could be busy doing something else and all the assumed delays are off.
I don't feel like you need to add the thread to delay the completion of the kernel. You can prob do a spin loop inside the kernel or do the signaling from the host using hipHostAlloc memory instead of hipMalloc or something.

tests/catch/hipTestMain/config/config_amd_linux.json
```diff
@@ -1,6 +1,10 @@
 {
     "DisabledTests":
 	[
+	   "# Skipped due to EXSWCPHIPT-19",
```
FMarno: convert to HIP_SKIP_TEST

tests/catch/hipTestMain/config/config_amd_windows.json
```diff
@@ -29,7 +29,11 @@
 	   "Unit_hipTextureMipmapObj2D_Check",
 	   "Unit_hipStreamPerThread_DeviceReset_1",
 	   "Unit_hipStreamPerThread_DeviceReset_2",
-	   "Unit_hipManagedKeyword_MultiGpu"	   
+	   "Unit_hipManagedKeyword_MultiGpu",
+	   "# Skipped due to EXSWCPHIPT-19",
```
FMarno: convert to HIP_SKIP_TEST

tests/catch/unit/stream/hipStreamDestroy.cc
```diff
@@ -0,0 +1,131 @@
...
+ *
+ * @param semaphore the semaphore to wait for.
+ */
+__global__ void waiting_kernel(int* semaphore) {
+  size_t tid = blockIdx.x * blockDim.x + threadIdx.x;
+  if (tid == 0) {
+    while (atomicCAS(semaphore, 1, 2) == 0) {
+    }
+  }
+}
```
FMarno: find a different solution to this. No guarantee that both kernels will run concurrently so this could lead to deadlock.

FMarno (commented):
Also need to rework Unit_hipStreamDestroy_WithPendingWork

chrispaquot:
Any update on this?

FMarno:
Hi @chrispaquot. @jatinx looking at a solution using managed memory but hit a bit of a snag with the atomics, more details are in ticket EXSWCPHIPT-78.

tests/catch/include/hip_test_common.hh
```diff
@@ -329,11 +329,89 @@ struct MemTraits<MemcpyAsync> {
   }
 };
 
-}  // namespace HipTest
 
+namespace {
```
fabiomestre: The changes to this file will have to be removed since the same code will be commited in a separate PR

tests/catch/unit/stream/hipStreamDestroy.cc
```diff
@@ -0,0 +1,97 @@
...
+  HIP_CHECK(hipFree(deviceData));
+}
+
+// hipStreamDestroy should return immediately then clean up the resources when the stream is empty
+// of work
+#if HT_AMD /* Disabled because frequency based wait is timing out on nvidia platforms */
+TEST_CASE("Unit_hipStreamDestroy_WithPendingWork") {
+#if HT_AMD
+  HipTest::HIP_SKIP_TEST(
+      "EXSWCPHIPT-44 - expected hipStreamDestroy to return immediately then release the resources "
```
fabiomestre, line 77: Need further clarification on why EXSWCPHIPT-44 was rejected. And maybe update the skip test message accordingly

FMarno:
@chrispaquot This is ready for review now. Unfortunately we've not been able to find a cross-platform reliable solution to replace the kernel that runs for a long time, but the current solution is at least reused between multiple tests. Due to that reuse, this PR is block by PR #2750.

FMarno:
@chrispaquot @mangupta This is now unblocked and ready to merge

PR #2549, opened 2022-03-15
EXSWCPHIPT-140 - Added hipStreamCreateWithFlags tests

FMarno:
Moved existing hipStreamCreateWithFlags tests to a dedicated file and added some extra tests.

depends on #2517

### DIFF
--- a/tests/catch/unit/stream/hipStreamCreateWithFlags.cc
+++ b/tests/catch/unit/stream/hipStreamCreateWithFlags.cc
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2021 Advanced Micro Devices, Inc. All rights reserved.
+Copyright (c) 2022 Advanced Micro Devices, Inc. All rights reserved.
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
 in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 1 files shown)


tests/catch/unit/stream/hipStreamCreateWithFlags.cc
```diff
@@ -17,23 +17,146 @@ OUT OF OR INN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
+  HIP_CHECK(hipMalloc(&device_data, data_size));
+  HIP_CHECK(hipMemset(device_data, 0, data_size));
+
+  hipStream_t comparingStream{};
+  HIP_CHECK(hipStreamCreateWithFlags(&comparingStream, hipStreamNonBlocking));
+
+  // kernel will wait until the second kernel sets some memory
+  hipLaunchKernelGGL(comparing_kernel, dim3(1), dim3(1), 0, comparingStream, device_data, 1);
+  // kernel will set some memory, allowing the first kernel to complete
+  hipLaunchKernelGGL(setting_kernel, dim3(1), dim3(1), 0, defaultStream, device_data, 1);
```
chrispaquot: Assuming overlapping in such a way might not always work in case the streams are on the same HW queue and that queue is full from extra work.
We usually measure concurrency by measuring time of each kernels then in parallel and t1+t2 >= combined.

chrispaquot: I guess since it's just one thread, it might just be fine.

chrispaquot (dismissed):

FMarno:
This PR is blocked by the same issue as #2548.

chrispaquot (dismissed):

tests/catch/include/hip_test_common.hh
```diff
@@ -329,11 +329,89 @@ struct MemTraits<MemcpyAsync> {
   }
 };
 
-}  // namespace HipTest
 
+namespace {
+static __global__ void waitKernel(clock_t offset) {
```
fabiomestre: Changes to this file need to be removed before removing the draft attribute

FMarno: Done

FMarno:
This is waiting on changes to hip_test_common.hh in #2750

FMarno:
@mangupta @chrispaquot This PR is now unblocked and ready to merge

PR #2558, opened 2022-03-17
Add test for hipStreamAttachMemAsync

jatinx:


gandryey (dismissed):
hipStreamAttachMemAsync is a nop under the current HMM implementation. All allocations have global access all time.

tests/catch/unit/stream/hipStreamAttachMemAsync.cc
```diff
@@ -0,0 +1,136 @@
...
+  if (streamType == StreamAttachTestType::StreamPerThread) {
+    stream = hipStreamPerThread;
+  } else if (streamType == StreamAttachTestType::CreatedStream) {
+    HIP_CHECK(hipStreamCreate(&stream));
+    REQUIRE(stream != nullptr);
+  }
+
+  SECTION("Invalid Resource Handle") {
+    int definitelyNotAManagedVariable = 0;
+    HIP_CHECK_ERROR(hipStreamAttachMemAsync(stream, &definitelyNotAManagedVariable, sizeof(int),
```
mangupta: dev_ptr needs to be type void* and not int*. Causes compilation failures. Similar issue is seen in the rest of the file as well.

mangupta (changes requested):

tests/catch/unit/stream/hipStreamAttachMemAsync.cc
jatinx: done

chrispaquot (dismissed):

tests/catch/unit/stream/CMakeLists.txt
```diff
@@ -25,9 +25,13 @@ set(TEST_SRC
     hipStreamCreateWithFlags.cc
     hipStreamCreateWithPriority.cc
     hipAPIStreamDisable.cc
+    # hipStreamAttachMemAsync.cc   # Disabling it on nvidia due to issue in function definition of hipStreamAttachMemAsync
```
chrispaquot, line 28: @gandryey Any comment on this?

gandryey: hipStreamAttachMemAsync is a nop for AMD. I'm not entirely sure what should be done there. We always have global attachment and transfer data per a GPU page fault. It seems nvidia had some kind of SW management before and that call could optimize data transfers per stream.

gandryey (approved):

PR #2559, opened 2022-03-17
Add tests for hipStreamWaitEvent

jatinx:


tests/catch/unit/stream/hipStreamWaitEvent.cc
```diff
@@ -0,0 +1,131 @@
...
+
+  int *d_a{nullptr}, *waitData{nullptr};
+  HIP_CHECK(hipMalloc(&d_a, sizeof(int)));
+  HIP_CHECK(hipMalloc(&waitData, sizeof(int)));
+
+  HIP_CHECK(hipMemset(d_a, 0, sizeof(int)));
+  HIP_CHECK(hipMemset(waitData, 0, sizeof(int)));
+
+  launchFunctor<<<1, 1, 0, blockedStreamA>>>(setValLambda, d_a, 1);  // d_a = 1
+  launchFunctor<<<1, 1, 0, blockedStreamA>>>(waitLambda,
```
chrispaquot: Instead of using a different kernel to unblock the other one, use host allocation to do that. This way you cannot run into a deadlock in case resources are scarce or if context switch isn't enabled.

chrispaquot: any update on this?

jatinx: Working on a different way to test wait events.

chrispaquot (dismissed):

PR #2609, opened 2022-04-04
Extending tests for hipHostRegister

dylanangus:
PR is dependent on previously uploaded PRs


tests/catch/unit/memory/hipHostRegister.cc
```diff
@@ -156,5 +152,56 @@ TEMPLATE_TEST_CASE("Unit_hipHostRegister_Memcpy", "",
...
+  TestType* hostPtr = nullptr;
+
+  size_t sizeBytes = 1 * sizeof(TestType);
+  SECTION("hipHostRegister Negative Test - nullptr") {
+    HIP_CHECK_ERROR(hipHostRegister(hostPtr, 1, 0), hipErrorInvalidValue);
+  }
+
+// Different error codes are reported for AMD and Nvidia devices
+#if HT_AMD
+  auto errorCode = hipErrorOutOfMemory;
```
chrispaquot: I think we have a fix for this coming soon.

fabiomestre: The test is still failing as of today with the most recent changes from hipamd. I will check again next week.

chrispaquot: yea I don't see it either. @gargrahul Do you know if we got a mirroring going last Friday?

fabiomestre: I updated the PR to change all the tests that have been fixed. Unfortunately, I'm unable to test the change because the most recent hipamd build is not building.

tests/catch/unit/memory/hipHostRegister.cc
```diff
@@ -156,5 +152,49 @@ TEMPLATE_TEST_CASE("Unit_hipHostRegister_Memcpy", "",
...
+  SECTION("hipHostRegister Negative Test - nullptr") {
+    HIP_CHECK_ERROR(hipHostRegister(hostPtr, 1, 0), hipErrorInvalidValue);
+  }
+
+  hostPtr = reinterpret_cast<TestType*>(malloc(sizeBytes));
+  SECTION("hipHostRegister Negative Test - zero size") {
+    HIP_CHECK_ERROR(hipHostRegister(hostPtr, 0, 0), hipErrorInvalidValue);
+  }
+
+#if HT_NVIDIA
```
chrispaquot, line 176: Did you file a ticket? We should return an error too

fabiomestre: I think the assumption here was that flags are completely ignored on AMD implementation and as such it didn't make sense to write negative testing for it. 

If that's not the case, then there is more errors that could be returned to match Nvidia implementation. Only 0x00 | 0x04 | 0x02 | 0x01 | 0x08 should be considered valid values for flags. 

We can add that extra negative testing and open a ticket to return error if the flags don't match those values. Would that make sense?

chrispaquot: sure

chrispaquot (approved):

PR #2613, opened 2022-04-05
Testing HipMalloc3D / HipMallocPitch / HipMemAllocPitch

fabiomestre:
This pull request has a dependency on:
https://github.com/ROCm-Developer-Tools/HIP/pull/2517


chrispaquot (approved):

tests/catch/unit/memory/hipMallocPitch.cc
```diff
@@ -0,0 +1,296 @@
...
+}
+
+TEST_CASE("Unit_hipMemAllocPitch_ValidatePitch") {
+  size_t pitch;
+  hipDeviceptr_t ptr;
+  hipExtent validExtent{generateExtent(AllocationApi::hipMemAllocPitch)};
+  MemoryInfo memBeforeAllocation{createMemoryInfo()};
+  unsigned int elementSizeBytes = GENERATE(4, 8, 16);
+
+#if HT_NVIDIA /* EXSWCPHIPT-46 */
```
chrispaquot: Actually we have a fix internally coming for this.

chrispaquot (changes requested):
We should probably change the EXSWCPHIPT-46 cases

tests/catch/unit/memory/hipMallocPitch.cc
fabiomestre: I can update this PR once the fix is on github. 
Or feel free to merge it in its current state (it will not fail) and I can then create a new PR to remove the ifdefs once the fix is merged to hipamd.

mangupta: I believe https://github.com/rocm-developer-tools/hipamd/commit/ffafa224d583ebf4acb4a0a36a6606b05ccab5e9 addresses these scenarios. PR can be updated now.

fabiomestre: I updated the PR to change all the tests that have been fixed. Unfortunately, I'm unable to test the change because the most recent hipamd build is not building. 

There is also a test that already exists in the Repo but is failing on Nvidia platforms: "Max size_t all dimensions". I modified it to check for the correct error code. I have created EXSWCPHIPT-117 for this issue.

Issue #2256, opened 2021-04-06
HIP_ROCCLR_HOME potentially used incorrectly in hipcc


haampie:
The 4.1.0 release changed how env / config variables are set, which is related to #2224.

Currently `HIP_ROCCLR_HOME` is used in `hipcc` to compute:

- `DEVICE_LIBPATH` -- this is overridden a few lines later with `$ROCM_PATH/amdgcn/bitcode`
- `HIP_INCLUDE_PATH` -- this is overridden with `$HIP_PATH/include`
- `HIP_LIB_PATH` -- this is *not* overridden and defaults `$HIP_ROCCLR_HOME/lib`

`HIP_LIB_PATH` is used to set the `-L ...` flag, which is used so that linking `-lamdhip64` works by default.

The problem is that `HIP_LIB_PATH` currently points to `$HIP_ROCCLR_HOME/lib` if not set as an environment variable:

```perl
if (!defined $HIP_LIB_PATH) {
    $HIP_LIB_PATH = "$HIP_ROCCLR_HOME/lib";
}
```

which is incorrect, because that folder does not have `libamdhip64.so`:

```console
path/to/hip-rocclr-4.1.0-u6jg27pd6e2foenlqjte3julsuex2r3a/lib $ ls
cmake  libamdrocclr_static.a
```

It should point to `$HIP_PATH/lib`

```console
path/to/hip-4.1.0-upqzah3svncjsrexw766rzklimzl7uj2/lib $ ls
cmake  hip_prof_str.h  libamdhip64.so  libamdhip64.so.4  libamdhip64.so.4.1.0
```

Of course I can work around this by setting the `HIP_LIB_PATH` variable... but this begs the question what the point of all Perl scripts is if I ultimately have to set all environment variables myself?

Could you potentially make a spack install part of your new release testing? It's a good way to figure out if the ROCm packages work when all subprojects are installed into separate directories instead of sharing a particular prefix `/opt/rocm-x.y.z`, and it would be a great way to test whether paths are correctly "computed" or whether they rely on everything being installed in the same prefix folder.

haampie:
Hm, and it seems `HIP_ROCCLR_HOME` is used to locate `.hipInfo`, but that file is installed as part of this repository, not https://github.com/ROCm-Developer-Tools/ROCclr/search?q=hipInfo? At least that's the case on my rocm installation.

```
if (defined $HIP_ROCCLR_HOME) {
[...]

PR #2113, opened 2020-06-22
add hipInfo file for nvcc

acxz:
This allows HIP_PLATFORM to be defined as nvcc during runtime for hipconfig.

See related discussion at https://github.com/rocm-arch/rocm-arch/issues/266

PR #2623, opened 2022-04-08
fix logic for finding nvcc even if clang exists

### LINKED ISSUE #1650: Better way to detect hip_platform
Hi,

I have a problem with detecting the correct hip_platform on Nvidia machines. The problem lies in the "hipconfig" script below. Since I have installed rocm-dev package for cross-compilation reasons, the command "/opt/rocm/hcc/bin/hcc --version" always runs gracefully and the HIP_PLATFORM is set to "hcc". This causes a problem when the gpu card on the machine is Nvidia. I was wondering if there is a better way to detect platform such as actually probing the system for gpu device.
```
sub can_run {
    my ($exe) = @_;
    `$exe --version 2>&1`;
    if ($? == 0) {
        return 1;
    } else {
        return 0;
    }
}
```
and 
```
if (not defined $HIP_PLATFORM) {
    if (can_run("$HCC_HOME/bin/hcc") or can_run("hcc")) {
        $HIP_PLATFORM = "hcc";
    } ...
```

Note: I know "export HIP_PLATFORM=nvcc" solves this issues but ideally I don't want to require the end-user of my program to do this. 

Thanks

acxz:
Issue: if a system has both `nvcc` and `clang++` then the HIP-PLATFORM autodetects to `amd`, instead of `nvidia`.

this PR fixes it by checking for nvcc first and then defaulting to amd if nvcc is not found.

Follow up of https://github.com/ROCm-Developer-Tools/HIP/pull/2113

Fixes: https://github.com/ROCm-Developer-Tools/HIP/issues/1650

Fixes comment: https://github.com/ROCm-Developer-Tools/HIP/issues/2256#issuecomment-818801109

### DIFF
--- a/bin/hipvars.pm
+++ b/bin/hipvars.pm
@@ -121,9 +121,7 @@ if (defined $HIP_RUNTIME and $HIP_RUNTIME eq "rocclr" and !defined $HIP_ROCCLR_H
 }
 
 if (not defined $HIP_PLATFORM) {
-    if (can_run("$HIP_CLANG_PATH/clang++") or can_run("clang++")) {
-        $HIP_PLATFORM = "amd";
-    } elsif (can_run("$CUDA_PATH/bin/nvcc") or can_run("nvcc")) {
+    if (can_run("$CUDA_PATH/bin/nvcc") or can_run("nvcc")) {
         $HIP_PLATFORM = "nvidia";
         $HIP_COMPILER = "nvcc";
         $HIP_RUNTIME = "cuda";


acxz:
@SarbojitAMD or anyone else, can they take a look at this PR?

acxz:
@SarbojitAMD @mangupta another friendly ping

acxz:
@SarbojitAMD @mangupta this issue has existed for roughly 2 years. Can this PR get some eyes on it?

mangupta:
@acxz : Some target systems have development environment & GPUs for both platforms. The current implementation is designed to "lean" towards selecting AMD platform in these cases. Users who don't wish for that can always set the appropriate HIP_PLATFORM environment variable. The proposed patch would change this behavior to the opposite direction.

acxz:
Thank you for the review!

> The current implementation is designed to "lean" towards selecting AMD platform in these cases.

This makes sense. The specific case, this PR is trying to address is when the system has clang++ (not compiled with rocm) and nvcc. In such a case, the first if statement gets triggered for amd, even though the user does not have amd.

Would only checking for `can_run("$HIP_CLANG_PATH/clang++")` in the first if sttement be more apt? That will retain behavior to lean towards amd on dual systems, but prevent the incorrect assignment of amd on the above situation.

PR #2664, opened 2022-05-03
Add HIP_CHECK_THREAD and REQUIRE_THREAD macro for multi threaded HIP API tests

jatinx:
- Adds ```HIP_CHECK_THREAD```, which is similar to ```HIP_CHECK``` but is only to be called in multi threaded tests
- Adds ```REQUIRE_THREAD``` which is similar to ```REQUIRE``` but is to be called in multi threaded tests
- Adds documentation of usage of these macros
- ~~Fix current multi threaded tests to use this macro~~ - To be done in later PRs
- Fixes some breaking tests

### DIFF
--- a/tests/catch/include/hip_test_common.hh
+++ b/tests/catch/include/hip_test_common.hh
@@ -22,9 +22,14 @@ THE SOFTWARE.
 
 #pragma once
 #include "hip_test_context.hh"
-#include <hip_test_rtc.hh>
 #include <catch.hpp>
+#include <atomic>
+#include <chrono>
 #include <stdlib.h>
+#include <iostream>
+#include <iomanip>
+#include <mutex>
+#include <cstdlib>
 
 #define HIP_PRINT_STATUS(status) INFO(hipGetErrorName(status) << " at line: " << __LINE__);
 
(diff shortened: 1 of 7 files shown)


tests/catch/include/hip_test_common.hh
```diff
@@ -23,39 +23,79 @@ THE SOFTWARE.
...
+
 #define HIPRTC_CHECK(error)                                                                        \
   {                                                                                                \
     auto localError = error;                                                                       \
-    if (localError != HIPRTC_SUCCESS) {                                                            \
-      INFO("Error: " << hiprtcGetErrorString(localError) << " Code: " << localError << " Str: "    \
-                     << #error << " In File: " << __FILE__ << " At line: " << __LINE__);           \
-      REQUIRE(false);                                                                              \
-    }                                                                                              \
+    INFO("Matching Error to HIPRTC_SUCCESS: "                                                      \
```
chrispaquot: I'm not sure we want to always display info if localError == success

tests/catch/include/hip_test_common.hh
```diff
@@ -23,39 +23,79 @@ THE SOFTWARE.
...
 #define HIP_PRINT_STATUS(status) INFO(hipGetErrorName(status) << " at line: " << __LINE__);
 
 #define HIP_CHECK(error)                                                                           \
   {                                                                                                \
     hipError_t localError = error;                                                                 \
-    if ((localError != hipSuccess) && (localError != hipErrorPeerAccessAlreadyEnabled)) {          \
-      INFO("Error: " << hipGetErrorString(localError) << " Code: " << localError << " Str: "       \
-                     << #error << " In File: " << __FILE__ << " At line: " << __LINE__);           \
-      REQUIRE(false);                                                                              \
+    INFO("Matching Error to hipSuccess or hipErrorPeerAccessAlreadyEnabled: "                      \
```
chrispaquot: we shouldn't always display info

tests/catch/include/hip_test_common.hh
jatinx: INFO is scoped and only printed when the condition that follows it (in the scope) is false.

So If REQUIRE, which follows INFO fails, the message will be printed.

jatinx:
Reducing the size of this PR to decrease the review times.

1st step, we will merge the infra changes

tests/catch/include/hip_test_common.hh
chrispaquot: How is that even possible?

jatinx: This is a catch2 feature.

"Additional messages can be logged during a test case. Note that the messages logged with INFO are scoped and thus will not be reported if failure occurs in scope preceding the message declaration"

https://github.com/catchorg/Catch2/blob/devel/docs/logging.md

In our case that's what we want, if the subsequent check fails the INFO would be printed with all the information

tests/catch/include/hip_test_common.hh
chrispaquot: Oh, good to know!

chrispaquot (dismissed):

tests/catch/unit/memory/hipArrayCreate.cc
```diff
@@ -32,6 +32,15 @@ static constexpr auto NUM_H{4};
 static constexpr auto BIGNUM_H{100};
 static constexpr auto ARRAY_LOOP{100};
 
+#ifdef __HIP_PLATFORM_NVIDIA__
+#define HIP_TEX_REFERENCE hipTexRef
+#define HIP_ARRAY hiparray
+#else  // __HIP_PLATFORM_NVIDIA__
+#define ARRAY_DESTROY(array) HIPCHECK(hipFreeArray(array));
```
mangupta: ARRAY_DESTROY needs to be defined for nvidia platform as well. Otherwise it causes the test to fail in building on nvidia platform.

mangupta (changes requested):
Fails to build "tests/catch/unit/memory/hipArrayCreate.cc" on nvidia platform due to missing macro.

chrispaquot (dismissed):

mangupta:
@jatinx : Can you please rebase this on top of current develop. Quite a few PRs were merged recently leading to several merge conflicts.

mangupta (approved):
Reapproving

PR #2675, opened 2022-05-12
hipMallocArray tests for default array and surface load store

FMarno:
Tests hipMallocArray including some basic operations on textures and surfaces made from the array.

There are negative tests and a test for texture gather also but they are dependent on another MR so I'll put up a PR for them later.

### DIFF
--- a/tests/catch/hipTestMain/config/config_amd_linux.json
+++ b/tests/catch/hipTestMain/config/config_amd_linux.json
@@ -1,12 +1,7 @@
 {
-  "DisabledTests" : [
-    "Unit_hipStreamPerThread_DeviceReset_1", "# Following test fails due to EXSWCPHIPT-59",
-    "Unit_hipMallocArray_Negative_ChannelAfterZeroChannel",
-    "# Following test fails due to EXSWCPHIPT-59",
-    "Unit_hipMallocArray_Negative_DifferentChannelSizes",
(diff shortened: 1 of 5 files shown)


tests/catch/hipTestMain/config/config_amd_linux.json
```diff
@@ -1,7 +1,12 @@
 {
-    "DisabledTests":
-	[
-	   "Unit_hipStreamPerThread_DeviceReset_1"
-	]
-
+  "DisabledTests" : [
+    "Unit_hipStreamPerThread_DeviceReset_1", "# Following test fails due to EXSWCPHIPT-59",
+    "Unit_hipMallocArray_Negative_ChannelAfterZeroChannel",
+    "# Following test fails due to EXSWCPHIPT-59",
```
chrispaquot: Shouldn't those be comments?

FMarno: Unfortunately there are no comments available in JSON, so we thought this would be a good compromise. From experience it doesn't seem to affect catch2 or ctest.

chrispaquot: Can't we just put that info inside the test source code then?

FMarno: Sure, I've sorted that out.

chrispaquot (approved):

PR #2713, opened 2022-06-02
EXSWCPHIPT-112 - Unit test for hipMalloc3DArray for default and surface arrays

FMarno:
This PR is stacked on #2702.

PR #2714, opened 2022-06-02
EXSWCPHIPT-113 - Unit tests for hipArray3DCreate for default and surface arrays

FMarno:
PR is stacked on #2713

### DIFF
--- a/tests/catch/unit/memory/DriverContext.cc
+++ b/tests/catch/unit/memory/DriverContext.cc
@@ -24,17 +24,13 @@ THE SOFTWARE.
 #include <hip_test_common.hh>
 
 DriverContext::DriverContext() {
-#if HT_NVIDIA
   HIP_CHECK(hipInit(0));
   HIP_CHECK(hipDeviceGet(&device, 0));
   HIP_CHECK(hipDevicePrimaryCtxRetain(&ctx, device));
   HIP_CHECK(hipCtxPushCurrent(ctx));
-#endif
 }
 
 DriverContext::~DriverContext() {
-#if HT_NVIDIA
   HIP_CHECK(hipCtxPopCurrent(&ctx));
   HIP_CHECK(hipDevicePrimaryCtxRelease(device));
-#endif
 }
--- a/tests/catch/unit/memory/DriverContext.hh
+++ b/tests/catch/unit/memory/DriverContext.hh
@@ -26,10 +26,8 @@ THE SOFTWARE.
 
 class DriverContext {
  private:
-#if HT_NVIDIA
   hipCtx_t ctx;
   hipDevice_t device;
-#endif
 
  public:
   DriverContext();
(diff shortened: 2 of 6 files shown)


tests/catch/unit/memory/DriverContext.hh
```diff
@@ -0,0 +1,41 @@
...
+THE SOFTWARE.
+*/
+
+#pragma once
+
+#include <hip_test_context.hh>
+
+class DriverContext {
+ private:
+#if HT_NVIDIA
```
chrispaquot: You don't really have to protect this

tests/catch/unit/memory/DriverContext.cc
```diff
@@ -0,0 +1,40 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+#include "DriverContext.hh"
+#include <hip_test_common.hh>
+
+DriverContext::DriverContext() {
+#if HT_NVIDIA
```
chrispaquot: Do we really need to protect for HT_NVIDIA here?

FMarno: The tests pass either way. Would you prefer I remove it then?

tests/catch/unit/memory/DriverContext.hh
FMarno: If I remove the implementation code then the compiler complains about unused variables.

tests/catch/unit/memory/DriverContext.cc
chrispaquot: yes

tests/catch/unit/memory/DriverContext.hh
chrispaquot: umm... ok, this is fine then.

tests/catch/unit/memory/DriverContext.cc
FMarno: I've done that now.

tests/catch/unit/memory/DriverContext.hh
FMarno: Since I removed the other preprocessor macros this has also been removed now.

chrispaquot (dismissed):

tests/catch/unit/memory/hipMalloc3DArray.cc
```diff
@@ -176,20 +126,145 @@ TEST_CASE("Unit_hipMalloc3DArray_MultiThread") {
...
+  const auto avail = getFreeMem();
 
   if (pavail != avail) {
     WARN("Memory leak of hipMalloc3D API in multithreaded scenario");
     REQUIRE(false);
   }
 }
 
+namespace {
+void checkArrayIsExpected(hipArray_t array, const hipChannelFormatDesc& expected_desc,
```
chrispaquot, line 149: What's the point of this function?

FMarno: like #2713, I have updated it do actually do something for the NVIDIA backend.

chrispaquot (dismissed):

mangupta (approved):
Adding back approval

PR #2717, opened 2022-06-06
Add sleep time after tests are run

arjun-raj-kuppala:


keryell:
Is there a motivation for this?
If so, it should be explained somewhere with a comment around he `sleep`.

PR #2718, opened 2022-06-06
EXSWCPHIPT-98 - [catch2] Added testing for hipFree API's

ShanoToni:


tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,452 @@
...
+  lauchLongRunningKernel(50);
+  // make sure device is busy
+  HIP_CHECK_ERROR(hipStreamQuery(nullptr), hipErrorNotReady);
+  HIP_CHECK(workIsDoneCheck<TestType*>(hostPtr, HostFree));
+
+  if (streamType == CREATEDSTR) HIP_CHECK(hipStreamDestroy(stream));
+}
+
+TEMPLATE_TEST_CASE("Unit_hipFreeImplicitSyncArray", "", char, float, float2, float4) {
+  HipTest::HIP_SKIP_TEST("EXSWCPHIPT-79");
```
chrispaquot: Doers this work with any platform?

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,452 @@
...
+  SECTION("InvalidPtr") {
+    hostPtr = new (char);
+    HIP_CHECK_ERROR(freeStuff(hostPtr, HostFree), hipErrorInvalidValue);
+    free(hostPtr);
+  }
+}
+
+// Freeing a invalid pointer with array
+TEST_CASE("Unit_hipFreeNegativeArray") {
+  HipTest::HIP_SKIP_TEST("EXSWCPHIPT-80&81");
```
chrispaquot: same question here

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,452 @@
...
+  }
+
+  for (auto& t : threads) {
+    t.join();
+  }
+  HIP_CHECK_THREAD_FINALIZE();
+}
+
+TEMPLATE_TEST_CASE("Unit_hipFreeMultiTArray", "", char, int, float2, float4) {
+  HipTest::HIP_SKIP_TEST("EXSWCPHIPT-79&81");
```
chrispaquot: same here

tests/catch/unit/memory/hipFree.cc
ShanoToni: On AMD devices the test causes a segfault, while on Nvidia devices the behaviour is handled however we still get error code 709: (context is destroyed)

tests/catch/unit/memory/hipFree.cc
ShanoToni: For this test case we expect the invalid memory to return ```hipErrorInvalidValue``` for invalid pointers being past and ```hipSuccess``` for nullptr, the observed result on AMD devices is the opposite (invalid pointers return success, nullptr retures invalid value)
On Nvidia devices we get get error code 709: (context is destroyed) for arrayFree negative testing and 400 (invalid resource handle) for arrayDestroy which is not the expected behaviour.

tests/catch/unit/memory/hipFree.cc
ShanoToni: Observed errors causing this test to fail are the same as in ```Unit_hipFreeImplicitSyncArray```
(On AMD devices the test causes a segfault, while on Nvidia devices the behaviour is handled however we still get error code 709: (context is destroyed)

tests/catch/unit/memory/hipFree.cc
chrispaquot: that test is quite different on AMD and NV (arrayPtr vs cuArrayPtr) so I don't like it the way it's written right now.

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,452 @@
...
+    case HostFree:
+      return hipHostFree(ptr);
+      break;
+    default:
+      return hipErrorIllegalState;
+      break;
+  }
+}
+
+#if HT_NVIDIA
```
chrispaquot: Why even have templates here? Just type cast and #ifdef the destroy case.

tests/catch/unit/memory/hipFree.cc
chrispaquot: what returns that error on NV? Seems like they're just not crashing but still returning some garbage since this has nothing to do with context.

tests/catch/unit/memory/hipFree.cc
ShanoToni: The function has been refactored. Thank you.

tests/catch/unit/memory/hipFree.cc
ShanoToni: The errors have been removed by patch in the HIPAMD repo: 06f64e1a53ddc9b0ca02993ff5bff95bea1f8f7f

tests/catch/unit/memory/hipFree.cc
ShanoToni: The test cases have been refactored and have been separated into smaller cases.

chrispaquot (dismissed):

mangupta (dismissed):
Reapproving which was lost due to resolving merge conflicts

chrispaquot (dismissed):

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,557 @@
...
+
+enum FreeType { DevFree, ArrayFree, ArrayDestroy, HostFree };
+
+__global__ void waitKernel(clock_t offset) {
+  auto time = clock();
+  while (clock() - time < offset) {
+  }
+}
+
+static __global__ void clock_kernel(clock_t clock_count, size_t* clockOut, size_t* co = nullptr) {
```
FMarno: since the co arguement is always used, remove the default value.

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,557 @@
...
+  *clockOut = clock();
+  clock_t start_clock = clock();
```
FMarno:
```suggestion
  clock_t start_clock = clock();
  *clockOut = start_clock;
```
This is so they both have the same value.

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,557 @@
...
+  clock_t clock_offset = 0;
+  while (clock_offset < clock_count) {
+    clock_offset = clock() - start_clock;
+  }
+  if (co != nullptr) *co = clock_offset;
+}
+
+// number of clocks the device is running at (device frequency)
+static size_t ticks = 0;
```
FMarno:
```suggestion
static size_t ticksPerMillisecond = 0;
```

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,557 @@
...
+
+  constexpr float mseconds = 1000;
+  constexpr float error = 0.02 * mseconds;
+
+  clock_t devFreq = static_cast<clock_t>(prop.clockRate);  // in kHz
+  clock_t time = devFreq * mseconds;
+
+
+  while (1) {
+    auto start = std::chrono::high_resolution_clock::now();
```
FMarno: events should be used to time the length of a kernel accurately

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,557 @@
...
+  HIP_CHECK(hipGetDeviceProperties(&prop, device));
+
+  constexpr float mseconds = 1000;
+  constexpr float error = 0.02 * mseconds;
+
+  clock_t devFreq = static_cast<clock_t>(prop.clockRate);  // in kHz
+  clock_t time = devFreq * mseconds;
+
+
+  while (1) {
```
FMarno: Can we get a count down to failure? like if it takes 10 attempts then maybe it's time to give up.

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,557 @@
...
+  hipDeviceProp_t prop;
+  int device;
+  size_t *clockFromKernel, *clockOffset;
+  HIP_CHECK(hipMalloc(&clockFromKernel, sizeof(size_t)));
+  HIP_CHECK(hipMalloc(&clockOffset, sizeof(size_t)));
+  HIP_CHECK(hipGetDevice(&device));
+  HIP_CHECK(hipGetDeviceProperties(&prop, device));
+
+  constexpr float mseconds = 1000;
+  constexpr float error = 0.02 * mseconds;
```
FMarno:
```suggestion
  constexpr float tolerance = 0.02 * mseconds;
```
Just seems a bit more appropriate

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,557 @@
...
+static size_t findTicks() {
+  hipDeviceProp_t prop;
+  int device;
+  size_t *clockFromKernel, *clockOffset;
+  HIP_CHECK(hipMalloc(&clockFromKernel, sizeof(size_t)));
+  HIP_CHECK(hipMalloc(&clockOffset, sizeof(size_t)));
+  HIP_CHECK(hipGetDevice(&device));
+  HIP_CHECK(hipGetDeviceProperties(&prop, device));
+
+  constexpr float mseconds = 1000;
```
FMarno:
```suggestion
  constexpr float milliseconds = 1000;
```
to remove confusion since microseconds could also be applicable here

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,557 @@
...
+      if (result >= mseconds) {
+        time -= (time * off);
+      } else {
+        time += (time * off);
+      }
+    }
+  }
+}
+
+static void runKernelForMs(size_t ms, hipStream_t stream = nullptr) {
```
FMarno:
```suggestion
static void runKernelForMs(size_t millis, hipStream_t stream = nullptr) {
```

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,557 @@
...
+  constexpr float mseconds = 1000;
+  constexpr float error = 0.02 * mseconds;
+
+  clock_t devFreq = static_cast<clock_t>(prop.clockRate);  // in kHz
+  clock_t time = devFreq * mseconds;
+
+
+  while (1) {
+    auto start = std::chrono::high_resolution_clock::now();
+    hipLaunchKernelGGL(clock_kernel, dim3(1), dim3(1), 0, 0, time, clockFromKernel, clockOffset);
```
FMarno: clockFromKernel isn't read so delete it

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,557 @@
...
+        time -= (time * off);
+      } else {
+        time += (time * off);
+      }
+    }
+  }
+}
+
+static void runKernelForMs(size_t ms, hipStream_t stream = nullptr) {
+  if (ticks == 0) {
```
FMarno: This all assumes you will always be using the same device and only a single thread (setting ticks could occur on multiple devices at once). This may be fine for this test, but might get a little dodgy if you try to use this in more places. 
If its fine for now then I suppose an appropriate response is just to detail these restrictions on the function.

tests/catch/unit/memory/hipFree.cc
FMarno: and the null check seems a bit unnecessary after you remove the default.

ShanoToni: Good point, removed clockOut

tests/catch/unit/memory/hipFree.cc
ShanoToni: Change made

tests/catch/unit/memory/hipFree.cc
ShanoToni: Change made

tests/catch/unit/memory/hipFree.cc
ShanoToni: Did not consider this, Events added to time kernel. Thank you!

tests/catch/unit/memory/hipFree.cc
ShanoToni: Counting down to approximation, failure was determined to be too extreme.

tests/catch/unit/memory/hipFree.cc
ShanoToni: Change made

tests/catch/unit/memory/hipFree.cc
ShanoToni: Amended

tests/catch/unit/memory/hipFree.cc
ShanoToni: Change made

tests/catch/unit/memory/hipFree.cc
ShanoToni: Deleted

tests/catch/unit/memory/hipFree.cc
ShanoToni: Added comments explaining lack of thread safety, as well as approach to take if changing devices.

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+  HIP_CHECK(hipGetLastError());
+}
+
+
+// helper functions to release memory
+template <typename T> hipError_t freeStuff(T ptr, FreeType type) {
+  switch (type) {
+    case DevFree:
+      return hipFree(ptr);
+      break;
```
FMarno: the break isn't needed after the return.

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+  // Warmup
+  hipLaunchKernelGGL(clock_kernel, dim3(1), dim3(1), 0, 0, time, clockOffset);
+  HIP_CHECK(hipGetLastError());
+  HIP_CHECK(hipDeviceSynchronize());
+
+  // try 10 times to find device frequency
+  int attempts = 6;
+  size_t co = 0;
+
+  while (attempts > 0) {
```
FMarno: use a for-loop here to keep the decrementing simpler

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+    float executionTime = 0;
+    HIP_CHECK(hipEventElapsedTime(&executionTime, start, stop));
+
+    HIP_CHECK(hipMemcpy(&co, clockOffset, sizeof(size_t), hipMemcpyDeviceToHost));
+    if (executionTime >= (milliseconds - tolerance) &&
+        executionTime <= (milliseconds + tolerance)) {
+      HIP_CHECK(hipFree(clockOffset));
+      HIP_CHECK(hipEventDestroy(start));
+      HIP_CHECK(hipEventDestroy(stop));
+      return co;
```
FMarno: maybe just break here instead of repeating the cleanup code.

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+  constexpr float tolerance = 0.00002 * milliseconds;
+
+  clock_t devFreq = static_cast<clock_t>(prop.clockRate);  // in kHz
+  clock_t time = devFreq * milliseconds;
+  // Warmup
+  hipLaunchKernelGGL(clock_kernel, dim3(1), dim3(1), 0, 0, time, clockOffset);
+  HIP_CHECK(hipGetLastError());
+  HIP_CHECK(hipDeviceSynchronize());
+
+  // try 10 times to find device frequency
```
FMarno, line 94:
```suggestion
  // try 10 times to find device frequency
  // after 10 attempts the result is likely good enough so just accept it
```

FMarno (commented):
the timer code looks pretty good, I just had some small comments

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+#include <hip_test_common.hh>
+#include <hip_test_checkers.hh>
+#include <hip_test_kernels.hh>
+#include "DriverContext.hh"
+
+/*
+ * This testcase verifies [ hipFree || hipFreeArray || hipArrayDestroy || hipHostFree with
+ * hipHostMalloc ]
+ * 1. Check that hipFree implicitly synchronises the device.
+ * 2.Perform multiple allocations and then call hipFree on each pointer concurrently (from unique
```
FMarno:
```suggestion
 * 2. Perform multiple allocations and then call hipFree on each pointer concurrently (from unique
```

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+TEST_CASE("Unit_hipFreeDoubleArrayDestroy") {
+#if HT_NVIDIA
+  HipTest::HIP_SKIP_TEST("EXSWCPHIPT-120");
+  return;
+#endif
+  size_t width = GENERATE(32, 512, 1024);
+  size_t height = GENERATE(0, 32, 512, 1024);
+  DriverContext ctx{};
+  // HipTest::HIP_SKIP_TEST("EXSWCPHIPT-81");
+  // return;
```
FMarno: remove dead code.

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+}
+
+#endif
+
+// Freeing a invalid pointer with on device
+TEST_CASE("Unit_hipFreeNegativeDev") {
+  char* devPtr;
+  SECTION("InvalidPtr") {
+    devPtr = new (char);
+    HIP_CHECK_ERROR(freeStuff(devPtr, DevFree), hipErrorInvalidValue);
```
FMarno: All the instance of freeStuff with a hardcoded value for type just seem like an unneeded indirection in my opinion.

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+ * 4. Pass an invalid ptr and check that hipErrorInvalidValue is returned.
+ * 5. Call hipFree twice on the same pointer and check that the implementation handles the second
+ * call correctly.
+ * 6. HipHostFree only:
+ *    Try to free memory that has been registered with hipHostRegister and check that
+ * hipErrorInvalidValue is returned.
+ */
+
+
+enum FreeType { DevFree, ArrayFree, ArrayDestroy, HostFree };
```
FMarno: I would prefer enum class.

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+#endif
+
+// Freeing a invalid pointer with on device
+TEST_CASE("Unit_hipFreeNegativeDev") {
+  char* devPtr;
+  SECTION("InvalidPtr") {
+    devPtr = new (char);
+    HIP_CHECK_ERROR(freeStuff(devPtr, DevFree), hipErrorInvalidValue);
+    delete devPtr;
+  }
```
FMarno:
```suggestion
  SECTION("InvalidPtr") {
    char value;
    HIP_CHECK_ERROR(hipFree(&value), hipErrorInvalidValue);
  }
```

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+
+  for (auto ptr : ptrs) {
+    SECTION("ArrayFree") {
+      threads.push_back(
+          std::thread([ptr]() { HIP_CHECK_THREAD(workIsDoneCheck<hipArray_t>(ptr, ArrayFree)); }));
+    }
+    SECTION("ArrayDestroy") {
+      threads.push_back(std::thread(
+          [ptr]() { HIP_CHECK_THREAD(workIsDoneCheck<hipArray_t>(ptr, ArrayDestroy)); }));
+    }
```
FMarno:
```suggestion
      threads.emplace_back(
          [ptr] { HIP_CHECK_THREAD(workIsDoneCheck<hipArray_t>(ptr, ArrayDestroy));});
    }
```

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+  for (auto& ptr : ptrs) {
+    HIP_CHECK(hipMallocArray(&ptr, &desc, extent.width, extent.height, hipArrayDefault));
+  }
+
+  std::vector<std::thread> threads;
+
+  for (auto ptr : ptrs) {
+    SECTION("ArrayFree") {
+      threads.push_back(
+          std::thread([ptr]() { HIP_CHECK_THREAD(workIsDoneCheck<hipArray_t>(ptr, ArrayFree)); }));
```
FMarno:
```suggestion
      threads.emplace_back(
          [ptr] { HIP_CHECK_THREAD(workIsDoneCheck<hipArray_t>(ptr, ArrayFree)); });
```

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+    }
+    SECTION("ArrayDestroy") {
```
FMarno: hipArrayDestroy should be used with hipArrayCreate

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+  DriverContext ctx;
+
+
+  SECTION("ArrayDestroy") {
+    std::vector<hiparray> ptrs(numAllocs);
+    HIP_ARRAY_DESCRIPTOR cuDesc;
+    cuDesc.Width = width;
+    cuDesc.Height = height;
+    cuDesc.Format = HIP_AD_FORMAT_UNSIGNED_INT8;
+    cuDesc.NumChannels = 2;
```
FMarno: you've hard coded the layout here. there is a helper class for this [here](https://github.com/ROCm-Developer-Tools/HIP/blob/1c4822f24a99448c14f375ac9dd54cde2c598a54/tests/catch/unit/memory/hipArrayCommon.hh#L37) which will give you the correct format and numchannel at compile time. example use [here](https://github.com/ROCm-Developer-Tools/HIP/blob/1c4822f24a99448c14f375ac9dd54cde2c598a54/tests/catch/unit/memory/hipArrayCreate.cc#L224)

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+  extent.width = width;
+  extent.height = height;
+  hipChannelFormatDesc desc = hipCreateChannelDesc<char>();
+
+  HIP_CHECK(hipMallocArray(&arrayPtr, &desc, extent.width, extent.height, hipArrayDefault));
+
+  HIP_CHECK(freeStuff(arrayPtr, ArrayFree));
+  HIP_CHECK_ERROR(freeStuff(arrayPtr, ArrayFree), hipErrorContextIsDestroyed);
+}
+#if HT_NVIDIA
```
FMarno: why is this test both removed with the preprocessor and skipped?

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+  }
+}
+#else
+
+// Freeing a invalid pointer with array
+TEST_CASE("Unit_hipFreeNegativeArray") {
+  hipArray_t arrayPtr{};
+
+  SECTION("InvalidPtr") {
+    arrayPtr = static_cast<hipArray*>(malloc(sizeof(char)));
```
FMarno: not freed

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+
+// Freeing a invalid pointer with on device
+TEST_CASE("Unit_hipFreeNegativeHost") {
+  char* hostPtr{nullptr};
+  SECTION("NullPtr") {
+    hostPtr = nullptr;
+    HIP_CHECK(freeStuff(hostPtr, HostFree));
+  }
+  SECTION("InvalidPtr") {
+    hostPtr = new (char);
```
FMarno: not freed

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+    HIP_CHECK_ERROR(freeStuff(hostPtr, HostFree), hipErrorInvalidValue);
+  }
+  SECTION("hipHostRegister") {
+    hostPtr = new char;
```
FMarno: not freed

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+    HIP_CHECK(freeStuff(devPtr, DevFree));
+  }
+}
+
+// Freeing a invalid pointer with on device
+TEST_CASE("Unit_hipFreeNegativeHost") {
+  char* hostPtr{nullptr};
+  SECTION("NullPtr") {
+    hostPtr = nullptr;
+    HIP_CHECK(freeStuff(hostPtr, HostFree));
```
FMarno:
```suggestion
    HIP_CHECK(hipHostFree(nullptr));
```

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+
+  if (streamType == CREATEDSTR) HIP_CHECK(hipStreamDestroy(stream));
+}
+#else  // AMD
+
+TEMPLATE_TEST_CASE("Unit_hipFreeImplicitSyncArray", "", char, float, float2, float4) {
+  enum StreamType { NULLSTR, CREATEDSTR };
+  auto streamType = GENERATE(NULLSTR, CREATEDSTR);
+  hipStream_t stream{nullptr};
+  if (streamType == CREATEDSTR) HIP_CHECK(hipStreamCreate(&stream));
```
FMarno: stream isn't used

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+TEMPLATE_TEST_CASE("Unit_hipFreeImplicitSyncArray", "", char, float, float2, float4) {
+  DriverContext ctx;
+
+  size_t width = GENERATE(32, 512, 1024);
+  size_t height = GENERATE(32, 512, 1024);
+
+  enum StreamType { NULLSTR, CREATEDSTR };
+  auto streamType = GENERATE(NULLSTR, CREATEDSTR);
+  hipStream_t stream{nullptr};
+  if (streamType == CREATEDSTR) HIP_CHECK(hipStreamCreate(&stream));
```
FMarno: stream isn't used

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+  HIP_CHECK_ERROR(hipStreamQuery(nullptr), hipErrorNotReady);
+  HIP_CHECK(workIsDoneCheck<TestType*>(devPtr, DevFree));
+  if (streamType == CREATEDSTR) HIP_CHECK(hipStreamDestroy(stream));
+}
+
+TEMPLATE_TEST_CASE("Unit_hipFreeImplicitSyncHost", "", char, float, float2, float4) {
+  enum StreamType { NULLSTR, CREATEDSTR };
+  auto streamType = GENERATE(NULLSTR, CREATEDSTR);
+  hipStream_t stream{nullptr};
+  if (streamType == CREATEDSTR) HIP_CHECK(hipStreamCreate(&stream));
```
FMarno: stream isn't used

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,580 @@
...
+  }
+  return error;
+}
+
+// DevFree, ArrayFree, ArrayDestroy, HostFree
+TEMPLATE_TEST_CASE("Unit_hipFreeImplicitSyncDev", "", char, float, float2, float4) {
+  enum StreamType { NULLSTR, CREATEDSTR };
+  auto streamType = GENERATE(NULLSTR, CREATEDSTR);
+  hipStream_t stream{nullptr};
+  if (streamType == CREATEDSTR) HIP_CHECK(hipStreamCreate(&stream));
```
FMarno: stream isn't used

tests/catch/unit/memory/hipFree.cc
ShanoToni: Comment added

tests/catch/unit/memory/hipFree.cc
ShanoToni: Memory release added.

tests/catch/unit/memory/hipFree.cc
ShanoToni: Memory release added.

tests/catch/unit/memory/hipFree.cc
ShanoToni: Memory release added.

tests/catch/unit/memory/hipFree.cc
ShanoToni: Some inconsistent failures are being looked into this test case, which might or might not fail, keeping test as skipped for now.

tests/catch/unit/memory/hipFree.cc
ShanoToni: Stream creation removed

tests/catch/unit/memory/hipFree.cc
ShanoToni: Stream creation removed

tests/catch/unit/memory/hipFree.cc
ShanoToni: Stream creation removed

tests/catch/unit/memory/hipFree.cc
ShanoToni: Stream creation removed

tests/catch/unit/memory/hipFree.cc
ShanoToni: Change added

tests/catch/unit/memory/hipFree.cc
ShanoToni: Added vec_info in testing

tests/catch/unit/memory/hipFree.cc
ShanoToni: Amended

tests/catch/unit/memory/hipFree.cc
ShanoToni: Changed

tests/catch/unit/memory/hipFree.cc
ShanoToni: Changed

tests/catch/unit/memory/hipFree.cc
ShanoToni: Test case simplified.

tests/catch/unit/memory/hipFree.cc
ShanoToni: Enum class it is

tests/catch/unit/memory/hipFree.cc
ShanoToni: Removed

tests/catch/unit/memory/hipFree.cc
ShanoToni: Space added

tests/catch/unit/memory/hipFree.cc
ShanoToni: Breaks removed.

tests/catch/unit/memory/hipFree.cc
ShanoToni: While replaced with for loop

tests/catch/unit/memory/hipFree.cc
ShanoToni: Changed

tests/catch/unit/memory/hipFree.cc
ShanoToni: Fair enough, removed

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,522 @@
...
+static void runKernelForMs(size_t millis, hipStream_t stream = nullptr) {
+  if (ticksPerMillisecond == 0) {
+    ticksPerMillisecond = findTicks();
+  }
+  hipLaunchKernelGGL(waitKernel, dim3(1), dim3(1), 0, stream, ticksPerMillisecond * millis / 1000);
+  HIP_CHECK(hipGetLastError());
+}
+
+
+// DevFree, ArrayFree, ArrayDestroy, HostFree
```
FMarno: what is this comment for?

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,522 @@
...
+TEMPLATE_TEST_CASE("Unit_hipFreeImplicitSyncDev", "", char, float, float2, float4) {
+  TestType* devPtr{};
+  size_t size_mult = GENERATE(1, 32, 64, 128, 256);
+  HIP_CHECK(hipMalloc(&devPtr, sizeof(TestType) * size_mult));
+
+  runKernelForMs(50);
```
FMarno: I would consider raising the value here. The contention may be quite high for a given CI system so a bit of extra padding is not a bad thing.

FMarno: also it looks like all instances of runKernelForMs use the same timeout, it's probably worth while refactoring this 50 into a named variable.

tests/catch/unit/memory/hipFree.cc
FMarno: When the nullptr section runs you are now calling delete on a pointer that hasn't been allocated.
Do you need a heap alloced value or is stack alloced sufficient? 
if it needs to be heap alloced then maybe use a unique_ptr

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,522 @@
...
+    free(arrayPtr);
+    free(cuArrayPtr);
+  }
+  SECTION("NullPtr") {
+    arrayPtr = nullptr;
+    cuArrayPtr = nullptr;
+    SECTION("ArrayFree") { HIP_CHECK(hipFreeArray(arrayPtr)); }
+    SECTION("ArrayDestroy") {
+      HIP_CHECK_ERROR(hipArrayDestroy(cuArrayPtr), hipErrorInvalidResourceHandle);
+    }
```
FMarno:
```suggestion
    SECTION("ArrayFree") { HIP_CHECK(hipFreeArray(nullptr)); }
    SECTION("ArrayDestroy") {
      HIP_CHECK_ERROR(hipArrayDestroy(nullptr), hipErrorInvalidResourceHandle);
    }
```

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,522 @@
...
+
+#if HT_NVIDIA
+TEST_CASE("Unit_hipFreeNegativeArray") {
+  DriverContext ctx;
+  hipArray_t arrayPtr{};
+  hiparray cuArrayPtr{};
+
+  SECTION("InvalidPtr") {
+    arrayPtr = static_cast<hipArray*>(malloc(sizeof(char)));
+    cuArrayPtr = static_cast<hiparray>(malloc(sizeof(char)));
```
FMarno: you use `malloc` here but `new` in other places. I would prefer a consistent strategy (and would suggest unique pointer as that strategy).

FMarno:
```suggestion
    arrayPtr = static_cast<hipArray*>(0xDEAD);
    cuArrayPtr = static_cast<hiparray>(0xBEEF);
```
if you just want so invalid values to tests as pointer.
Personally I think this test is unnecessary and too much like undefined behaviour. Also I think the behaviour is covered in the double free test.

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,522 @@
...
+    SECTION("ArrayFree") { HIP_CHECK_ERROR(hipFreeArray(arrayPtr), hipErrorContextIsDestroyed); }
+    SECTION("ArrayDestroy") {
+      HIP_CHECK_ERROR(hipArrayDestroy(arrayPtr), hipErrorContextIsDestroyed);
+    }
+  }
+
+  SECTION("NullPtr") {
+    arrayPtr = nullptr;
+    SECTION("ArrayFree") { HIP_CHECK_ERROR(hipFreeArray(arrayPtr), hipErrorInvalidValue); }
+    SECTION("ArrayDestroy") { HIP_CHECK_ERROR(hipArrayDestroy(arrayPtr), hipErrorInvalidValue); }
```
FMarno:
```suggestion
    SECTION("ArrayFree") { HIP_CHECK_ERROR(hipFreeArray(nullptr), hipErrorInvalidValue); }
    SECTION("ArrayDestroy") { HIP_CHECK_ERROR(hipArrayDestroy(nullptr), hipErrorInvalidValue); }
```

tests/catch/unit/memory/hipFree.cc
FMarno: same as above, I think this behaviour is covered by the double free test.

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,522 @@
...
+  cuDesc.Format = vec_info::format;
+  cuDesc.NumChannels = vec_info::size;
+  HIP_CHECK(hipArrayCreate(&ArrayPtr, &cuDesc));
+  HIP_CHECK(hipArrayDestroy(ArrayPtr));
+  HIP_CHECK_ERROR(hipArrayDestroy(ArrayPtr), hipErrorContextIsDestroyed);
+}
+#endif
+
+
+// DevFree, ArrayFree, ArrayDestroy, HostFree Mutithreaded
```
FMarno: I'm not sure what this comment means

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,522 @@
...
+  size_t allocSize = sizeof(TestType) * GENERATE(1, 32, 64, 128);
+
+  for (auto& ptr : ptrs) {
+    HIP_CHECK(hipMalloc(&ptr, allocSize));
+  }
+
+  std::vector<std::thread> threads;
+
+  for (auto ptr : ptrs) {
+    threads.emplace_back(std::thread([ptr] {
```
FMarno:
```suggestion
    threads.emplace_back([ptr] {
```
emplace_back should forward the argument to the constructor of the type.

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,522 @@
...
+  size_t allocSize = sizeof(TestType) * GENERATE(1, 32, 64, 128);
+
+  for (auto& ptr : ptrs) {
+    HIP_CHECK(hipHostMalloc(&ptr, allocSize));
+  }
+
+  std::vector<std::thread> threads;
+
+  for (auto ptr : ptrs) {
+    threads.emplace_back(std::thread([ptr] {
```
FMarno:
```suggestion
    threads.emplace_back([ptr] {
```

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,522 @@
...
+      threads.emplace_back(std::thread([ptr] {
+        HIP_CHECK_THREAD(hipArrayDestroy(ptr));
+        HIP_CHECK_THREAD(hipStreamQuery(nullptr));
+      }));
+    }
+    for (auto& t : threads) {
+      t.join();
+    }
+    HIP_CHECK_THREAD_FINALIZE();
+  };
```
FMarno:
```suggestion
  }
```

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,522 @@
...
+    cuDesc.Height = height;
+    cuDesc.Format = vec_info::format;
+    cuDesc.NumChannels = vec_info::size;
+    for (auto& ptr : ptrs) {
+      HIP_CHECK(hipArrayCreate(&ptr, &cuDesc));
+    }
+
+
+    for (auto& ptr : ptrs) {
+      threads.emplace_back(std::thread([ptr] {
```
FMarno:
```suggestion
      threads.emplace_back([ptr] {
```

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,522 @@
...
+      }));
+    }
+    for (auto& t : threads) {
+      t.join();
+    }
+    HIP_CHECK_THREAD_FINALIZE();
+  };
+
+  SECTION("ArrayFree") {
+    constexpr size_t numAllocs = 10;
```
FMarno: this is shadowing the same value

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,522 @@
...
+    extent.height = height;
+    hipChannelFormatDesc desc = hipCreateChannelDesc<TestType>();
+
+    for (auto& ptr : ptrs) {
+      HIP_CHECK(hipMallocArray(&ptr, &desc, extent.width, extent.height, hipArrayDefault));
+    }
+
+    for (auto ptr : ptrs) {
+      SECTION("ArrayFree") {
+        threads.emplace_back(std::thread([ptr] {
```
FMarno:
```suggestion
        threads.emplace_back([ptr] {
```

tests/catch/unit/memory/hipFree.cc
ShanoToni: Removed old comments

tests/catch/unit/memory/hipFree.cc
ShanoToni: Testing showed the API operations taking at most 2ms (for significantly bigger memory operations) I believe 50 ms to be a safe time.

Also moved the 50ms to a variable.

tests/catch/unit/memory/hipFree.cc
ShanoToni: Change made

tests/catch/unit/memory/hipFree.cc
ShanoToni: Considered it, and I agree double free captures the case well enough invalid ptr sections removed

tests/catch/unit/memory/hipFree.cc
ShanoToni: Change made

tests/catch/unit/memory/hipFree.cc
ShanoToni: Removed

tests/catch/unit/memory/hipFree.cc
ShanoToni: Amended

tests/catch/unit/memory/hipFree.cc
ShanoToni: Fixed

tests/catch/unit/memory/hipFree.cc
ShanoToni: Removed, No idea how that got there :)

tests/catch/unit/memory/hipFree.cc
ShanoToni: fixed

tests/catch/unit/memory/hipFree.cc
ShanoToni: Fixed

tests/catch/unit/memory/hipFree.cc
ShanoToni: Addressed

chrispaquot (dismissed):

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,522 @@
...
+static void runKernelForMs(size_t millis, hipStream_t stream = nullptr) {
+  if (ticksPerMillisecond == 0) {
+    ticksPerMillisecond = findTicks();
+  }
+  hipLaunchKernelGGL(waitKernel, dim3(1), dim3(1), 0, stream, ticksPerMillisecond * millis / 1000);
+  HIP_CHECK(hipGetLastError());
+}
+
+
+// DevFree, ArrayFree, ArrayDestroy, HostFree
```
FMarno:
```suggestion
```

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,522 @@
...
+// Freeing a invalid pointer with on device
+TEST_CASE("Unit_hipFreeNegativeDev") {
+  SECTION("InvalidPtr") {
+    char value;
+    HIP_CHECK_ERROR(hipFree(&value), hipErrorInvalidValue);
+  }
+  SECTION("NullPtr") { HIP_CHECK(hipFree(nullptr)); }
+}
+
+// Freeing a invalid pointer with on device
```
FMarno:
```suggestion
// Freeing a invalid pointer with on host
```

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,522 @@
...
+  DriverContext ctx;
+  hipArray_t arrayPtr{};
+  hiparray cuArrayPtr{};
+
+  SECTION("NullPtr") {
+    SECTION("ArrayFree") { HIP_CHECK(hipFreeArray(nullptr)); }
+    SECTION("ArrayDestroy") {
+      HIP_CHECK_ERROR(hipArrayDestroy(nullptr), hipErrorInvalidResourceHandle);
+    }
+  }
```
FMarno:
```suggestion
    SECTION("ArrayFree") { HIP_CHECK(hipFreeArray(nullptr)); }
    SECTION("ArrayDestroy") {
      HIP_CHECK_ERROR(hipArrayDestroy(nullptr), hipErrorInvalidResourceHandle);
    }
```

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,522 @@
...
+// Freeing a invalid pointer with array
+TEST_CASE("Unit_hipFreeNegativeArray") {
+  hipArray_t arrayPtr{};
+
+  SECTION("NullPtr") {
+    SECTION("ArrayFree") { HIP_CHECK_ERROR(hipFreeArray(nullptr), hipErrorInvalidValue); }
+    SECTION("ArrayDestroy") { HIP_CHECK_ERROR(hipArrayDestroy(nullptr), hipErrorInvalidValue); }
+  }
+
+  free(arrayPtr);
```
FMarno:
```suggestion

    SECTION("ArrayFree") { HIP_CHECK_ERROR(hipFreeArray(nullptr), hipErrorInvalidValue); }
    SECTION("ArrayDestroy") { HIP_CHECK_ERROR(hipArrayDestroy(nullptr), hipErrorInvalidValue); }
 
```

FMarno (commented):
other than the `free` this is just nitpicks, so fix that and I'll approve.

tests/catch/unit/memory/hipFree.cc
ShanoToni: Fixed

tests/catch/unit/memory/hipFree.cc
ShanoToni: Fixed

tests/catch/unit/memory/hipFree.cc
ShanoToni: Fixed

tests/catch/unit/memory/hipFree.cc
ShanoToni: Fixed

FMarno (dismissed):
LGTM

chrispaquot (dismissed):

mangupta (approved):

mangupta:
Closing since changes are a part of #2750

PR #2719, opened 2022-06-06
EXSWCPHIPT-42 - Add HIP RTC support to the test framework

fabiomestre:


chrispaquot (approved):

tests/README.md
```diff
@@ -147,6 +147,22 @@ For example,
...
 
+### RTC Testing
+
+To enable RTC testing, cmake needs to be passed the DRTC_TESTING=1 options.
+
+When this option is passed, all tests that support this functionality will be run using HIP RTC to compile and run.
+
+To enable HIP RTC support for a specific test:
+    1 - Move all its kernels to tests/catch/kernels (one file per kernel)
+    2 - Update tests/catch/kernels/CMakeLists.txt
```
alistair-low, line 158: Could you maybe say more about what we have to do in steps 2-4?

tests/catch/external/Catch2/cmake/Catch2/Catch.cmake
```diff
@@ -203,7 +203,16 @@ function(hip_add_exe_to_target)
     "${list_args}"
   )
   # Create shared lib of all tests
+  if(NOT RTC_TESTING)
+  add_executable(${_NAME} EXCLUDE_FROM_ALL ${_TEST_SRC} $<TARGET_OBJECTS:Main_Object> $<TARGET_OBJECTS:KERNELS>)
```
alistair-low, line 207: Looks like lines 206 and 209 should be indented

tests/catch/include/hip_test_common.hh
```diff
@@ -171,7 +172,59 @@ static inline void HIP_SKIP_TEST(char const* const reason) noexcept {
...
+ * @param numThreads
+ * @param memPerBlock
+ * @param stream
+ * @param packedArgs A list of kernel arguments to be forwarded.
+ */
+template <typename... Typenames, typename K, typename Dim, typename... Args>
+void launchKernel(K kernel, Dim numBlocks, Dim numThreads, std::uint32_t memPerBlock,
+                  hipStream_t stream, Args&&... packedArgs) {
+#ifndef RTC_TESTING
+    validateArguments(kernel, packedArgs...);
```
alistair-low, line 220: Looks like these two lines should be tab width = 2 to match the rest of the code

tests/catch/include/hip_test_rtc.hh
```diff
@@ -0,0 +1,279 @@
...
+                     std::uint32_t memPerBlock, hipStream_t stream, Args&&... packedArgs) {
+
+  printInfo();
+  TestContext& testContext = TestContext::get();
+  std::string kernelName = (*getKernelName)();
+
+  std::vector<std::string> kernelTypenames{std::string(HipTest::getTypeName<Typenames>())...};
+  std::string kernelExpression = reconstructExpression(kernelName, kernelTypenames);
+
+  if (testContext.getFunction(kernelExpression) == nullptr) {
```
alistair-low, line 229: You could maybe put this logic into compileRTC so that you call it with the source and it returns the compiled kernel, without you having to check in this function whether or not it's already been compiled. But that's maybe not worth the effort.

tests/catch/include/hip_test_rtc.hh
```diff
@@ -0,0 +1,279 @@
...
+    hiprtcProgram rtcProgram{compileRTC(kernelName, kernelExpression)};
+    std::vector<char> compiledCode{getKernelCode(rtcProgram)};
+
+    hipModule_t module;
+
+    REQUIRE(hipSuccess == hipModuleLoadData(&module, compiledCode.data()));
+
+      hipFunction_t kernelFunction;
```
alistair-low, line 237: Indentation looks wrong.

alistair-low (commented):
Looks good to me, I could understand the code and didn't see any major problems. I left a few minor comments.

tests/README.md
```diff
@@ -147,6 +147,22 @@ For example,
 Here "-C performance" indicate the "performance" configuration of ctest.
 ```
 
+### RTC Testing
+
+To enable RTC testing, cmake needs to be passed the DRTC_TESTING=1 options.
```
FMarno, line 152: -DRTC_TESTING=1 with a '-'

FMarno: "options" -> "option"

tests/README.md
```diff
@@ -147,6 +147,22 @@ For example,
...
+
+To enable HIP RTC support for a specific test:
+    1 - Move all its kernels to tests/catch/kernels (one file per kernel)
+    2 - Update tests/catch/kernels/CMakeLists.txt
+    3 - Update tests/catch/include/kernels.hh
+    4 - Update tests/catch/include/kernel_mapping.hh
+    5 - Include kernels.hh
+    6 - Call hipTest::launchKernel() function instead of hipLaunchKernelGGL()
+
+Note: HIP RTC does not do implicit casting of kernel parameters. This requires the test writer to explicitly do all the casting before running the kernel. The code will not compile otherwise.
```
FMarno, line 164: Does the code not compile in even when compiled with RTC_TESTING? I thought we discussed that it would compile with RTC_TESTING since you don't really know the type at compile time. If that is the case then it would be worth mentioning that runtime errors can occur.

tests/catch/include/hip_test_rtc.hh
FMarno: as discussed, this is a potential race condition that could lead to a memory leak.

tests/catch/unit/memory/hipHostMalloc.cc
```diff
@@ -185,6 +182,7 @@ TEST_CASE("Unit_hipHostMalloc_NonCoherent") {
                    SYNC_STREAM, ptrType);
   CheckHostPointer(numElements, A, hipEventReleaseToSystem,
                    SYNC_EVENT, ptrType);
+  TestContext::get().cleanContext();
```
FMarno, line 185: I think it may be a little tricky to remember to call this clean up function for every test that uses a kernel. Could we tie the state to an object local to each test so that the destructor is always called? or what do you think?

FMarno (commented):
Looks pretty good to me.
The only thing is the awkwardness of calling cleanContext, but that may be unavoidable.

PR #2721, opened 2022-06-07
EXSWCPHIPT-35 - Add hipEventQuery tests

jatinx:


tests/catch/unit/event/Unit_hipEventQuery.cc
```diff
@@ -0,0 +1,101 @@
...
+
+// Wait till *ptr is equal to old_val, replace it with new_val
+// Basically wrapper around CAS
+__global__ void blockKernel(int* ptr, int old_val, int new_val) {
+  while (atomicCAS_system(ptr, old_val, new_val) != new_val)
+    ;
+}
+
+TEST_CASE("Unit_hipEventQuery_DifferentDevice") {
+  if (HipTest::getDeviceCount() <= 1) {
```
fabiomestre: Ctest will not mark this test as skipped. It would be better to use HIP_SKIP_TEST macro.

tests/catch/unit/event/Unit_hipEventQuery.cc
```diff
@@ -0,0 +1,101 @@
...
+  HIP_CHECK(hipMallocManaged(&ptr, sizeof(int), hipMemAttachGlobal));
+  REQUIRE(ptr != nullptr);
+
+  // Set it to 0
+  __atomic_store_n(ptr, 0, __ATOMIC_SEQ_CST);
+
+  constexpr int old_val = 10, new_val = 20;
+
+  // Start kernel on 1st device
+  {
```
fabiomestre, line 53: This part of the test should run even when there is only one device

tests/catch/unit/event/Unit_hipEventQuery.cc
```diff
@@ -0,0 +1,101 @@
...
+    HIP_CHECK(hipEventSynchronize(event1));
+    HIP_CHECK(hipEventQuery(event1));  // Should be done
+
+    HIP_CHECK_ERROR(hipEventQuery(event2),
+                    hipErrorNotReady);  // Wont be done since kernel is waiting
+  }
+
+  // Query Event from different device
+  {
+    HIP_CHECK(hipSetDevice(1));
```
fabiomestre, line 72: This part of the test should only run when there is 2 devices available

jatinx: Done

tests/catch/unit/event/Unit_hipEventQuery.cc
jatinx: removed

tests/catch/unit/event/Unit_hipEventQuery.cc
jatinx: done

fabiomestre (dismissed):

chrispaquot (dismissed):

tests/catch/unit/event/Unit_hipEventQuery.cc
```diff
@@ -0,0 +1,100 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+#include <hip_test_common.hh>
+
+// Wait till *ptr is equal to old_val, replace it with new_val
+// Basically wrapper around CAS
+__global__ void blockKernel(int* ptr, int old_val, int new_val) {
+  while (atomicCAS_system(ptr, old_val, new_val) != new_val)
```
mangupta: I believe atomicCAS_system is available on nvidia for sm_60 and higher only. So we will get compilation failures on systems older than that.

mangupta (changes requested):

tests/catch/unit/event/Unit_hipEventQuery.cc
jatinx: Made it timing based

chrispaquot (dismissed):

mangupta:
@jatinx : The following test fails on NV - 

642 - Unit_hipEventQuery_DifferentDevice (Timeout)

PR #2722, opened 2022-06-07
EXSWCPHIPT-102 - Adding hipEventRecord Tests

jatinx:


chrispaquot (dismissed):

tests/catch/unit/event/Unit_hipEventRecord.cc
```diff
@@ -28,64 +28,90 @@ THE SOFTWARE.
...
+#if HT_AMD
+    WithFlags_ReleaseToDevice = hipEventReleaseToDevice,
+    WithFlags_ReleaseToSystem = hipEventReleaseToSystem,
+#endif
+    WithoutFlags
+  };
 
-    hipEvent_t start, stop;
+  auto flags = GENERATE(WithFlags_Default, WithFlags_Blocking, WithFlags_DisableTiming,
+#if HT_AMD
```
mangupta, line 54: You cannot have a ```#if ... #endif``` block inside the ```GENERATE``` since the latter is a macro.

mangupta (changes requested):
The change will cause a build failure. We need to move the ```#if...#endif``` block outside the ```GENERATE```.

tests/catch/unit/event/Unit_hipEventRecord.cc
```diff
@@ -28,64 +28,90 @@ THE SOFTWARE.
 #include <hip_test_common.hh>
```
fabiomestre, line 25: The copyright message still has 2021. It should be 2022

tests/catch/unit/event/Unit_hipEventRecord.cc
jatinx: Hey Maneesh, It works fine and behaves as expected i.e. AMD has more tests than NVIDIA since two flags are AMD specific.

on AMD: All tests passed (125 assertions in 2 test cases)

on NVIDIA: All tests passed (83 assertions in 2 test cases)

tests/catch/unit/event/Unit_hipEventRecord.cc
jatinx: done

tests/catch/unit/event/Unit_hipEventRecord.cc
mangupta: @jatinx : Here is what I see when I build the tests:
```
HIP_PR-2722/hip/tests/catch/unit/event/Unit_hipEventRecord.cc:57:2: error: embedding a directive within macro arguments has undefined behavior [-Werror,-Wembedded-directive]
#if HT_AMD
 ^
HIP_PR-2722/hip/tests/catch/unit/event/Unit_hipEventRecord.cc:59:2: error: embedding a directive within macro arguments has undefined behavior [-Werror,-Wembedded-directive]
#endif
 ^
```

jatinx: I can see it too by adding a -Wembedded-directive explicitly.
Its not seen with -Wall and -Werror.

Will change it.
Thanks for letting me know.

mangupta:
@jatinx : Thanks. Can you also rebase this on top of develop. PR #2719 just got merged in resulting in merge conflicts.

chrispaquot (dismissed):

chrispaquot (approved):

PR #2727, opened 2022-06-09
EXSWCPHIPT-105: Add testing for hipEventDestroy

fabiomestre:


chrispaquot (dismissed):

tests/catch/unit/event/hipEventDestroy.cc
```diff
@@ -0,0 +1,107 @@
...
+TEST_CASE("Unit_hipEventDestroy_Unfinished") {
+  hipEvent_t event;
+  HIP_CHECK(hipEventCreate(&event));
+
+  float *A_h, *B_h, *C_h;
+  launchVectorAdd(A_h, B_h, C_h);
+
+  HIP_CHECK(hipEventRecord(event));
+  HIP_CHECK_ERROR(hipEventQuery(event), hipErrorNotReady);
+  HIP_CHECK(hipEventDestroy(event));
```
FMarno, line 64: This is a race condition since the event could complete before hipEventQuery causing the test to fail, or betwee hipEventQuery and hipEventDestroy making the test invalid. I suppose if we fix the stream tests then the same solution could be applied here.

tests/catch/unit/event/hipEventDestroy.cc
```diff
@@ -0,0 +1,107 @@
...
+}
+
+TEST_CASE("Unit_hipEventDestroy_Negative") {
+  
+  SECTION("Invalid Event") {
+    hipEvent_t event{nullptr};
+    HIP_CHECK_ERROR(hipEventDestroy(event), hipErrorInvalidResourceHandle);
+  }
+
+#ifndef HT_AMD
```
FMarno: I'm pretty sure HT_AMD is defined in all cases so we should just use `#if`. See hip_test_context.hh line 39.
Also I think it would be better if we would use a test skip instead of just removing the test with the preprocessor.

tests/catch/unit/event/hipEventDestroy.cc
```diff
@@ -0,0 +1,107 @@
...
+  HIP_CHECK(hipMalloc(reinterpret_cast<void**>(&C_d), vectorSize));
+
+  hipEvent_t event;
+  HIP_CHECK(hipEventCreate(&event));
+
+  hipStream_t stream;
+  HIP_CHECK(hipStreamCreate(&stream));
+
+  float *A_h, *B_h, *C_h;
+  launchVectorAdd(A_h, B_h, C_h);
```
FMarno: I think it would be clearer to run the vector add on the created stream.

tests/catch/unit/event/hipEventDestroy.cc
```diff
@@ -0,0 +1,107 @@
...
+  float *A_d, *B_d, *C_d;
+  HIP_CHECK(hipMalloc(reinterpret_cast<void**>(&A_d), vectorSize));
+  HIP_CHECK(hipMalloc(reinterpret_cast<void**>(&B_d), vectorSize));
+  HIP_CHECK(hipMalloc(reinterpret_cast<void**>(&C_d), vectorSize));
+
+  hipEvent_t event;
+  HIP_CHECK(hipEventCreate(&event));
+
+  hipStream_t stream;
+  HIP_CHECK(hipStreamCreate(&stream));
```
FMarno, line 80: stream is not destroyed.

tests/catch/unit/event/hipEventDestroy.cc
```diff
@@ -0,0 +1,107 @@
...
+
+  float *A_h, *B_h, *C_h;
+  launchVectorAdd(A_h, B_h, C_h);
+
+  HIP_CHECK(hipEventRecord(event, stream));
+  HIP_CHECK_ERROR(hipEventQuery(event), hipErrorNotReady);
```
FMarno, line 86: I think this is also a race condition for the same reasons as the other test.

tests/catch/unit/event/hipEventDestroy.cc
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * @brief Check that destroying an event enqueued to a stream causes no errors.
+ *
+ */
+TEST_CASE("Unit_hipEventDestroy_WithWaitingStream") {
+  float *A_d, *B_d, *C_d;
+  HIP_CHECK(hipMalloc(reinterpret_cast<void**>(&A_d), vectorSize));
+  HIP_CHECK(hipMalloc(reinterpret_cast<void**>(&B_d), vectorSize));
+  HIP_CHECK(hipMalloc(reinterpret_cast<void**>(&C_d), vectorSize));
```
FMarno: this data isn't freed.

fabiomestre: This is actually a mistake. Forgot to delete this code.

fabiomestre:
Closing this pull request until a solution to EXSWCPHIPT-78 is found.

fabiomestre:
@chrispaquot This PR has been updated. It can be merged once https://github.com/ROCm-Developer-Tools/HIP/pull/2750 is merged

PR #2728, opened 2022-06-09
SWDEV-327563 - hiprtc functional test fix

agunashe:
SWDEV-327563 - hiprtc functional test fix

Windows fix - adding architecture to hiprtc functional test.


mangupta:
The changes proposed in this patch is not the right fix. Abandoned this change in favor of ad988ac01a9993295c70b4e4923fe87e63aa3d00 as a workaround for now.

PR #2750, opened 2022-06-20
EXSWCPHIPT-118 - Added testing for hipMemset Synchronous behavoiour.

ShanoToni:


chrispaquot (commented):
Is this kinda the same as https://github.com/ROCm-Developer-Tools/HIP/pull/2718 ?

ShanoToni:
> Is this kinda the same as #2718 ?

It is similar, These test cases check for synchronization by the hipMemset API's with respect to the memory used (Host or managed memory should be synchronous, while device memory operations should be asynchronous with respect to the host)
The PR is stacked on top of #2718 due to these test cases reuse the clock_kernel to ensure the device is busy, the use clock kernel and the function calling it have been moved to a common header file.

chrispaquot (dismissed):

mangupta (dismissed):
Reapproving

mangupta (dismissed):

chrispaquot (dismissed):

chrispaquot (dismissed):

tests/catch/unit/memory/hipFree.cc
```diff
@@ -0,0 +1,421 @@
...
+ *    Try to free memory that has been registered with hipHostRegister and check that
+ * hipErrorInvalidValue is returned.
+ */
+
+
+enum class FreeType { DevFree, ArrayFree, ArrayDestroy, HostFree };
+
+// Amount of time kernel should wait
+using namespace std::chrono_literals;
+const std::chrono::duration<uint64_t, std::milli> delay = 50ms;
```
FMarno, line 48:
```suggestion
constexpr std::chrono::milliseconds delay(50);
```

tests/catch/unit/memory/hipMemsetSync.cc
```diff
@@ -0,0 +1,505 @@
...
+}
+
+// Helper function to run tests for hipMemset allocation types
+template <typename T>
+void runTests(allocType type, memSetType memsetType, MultiDData data, hipStream_t stream) {
+  bool async = GENERATE(true, false);
+  CAPTURE(type, memsetType, data.width, data.height, data.depth, stream, async);
+  std::pair<T*, T*> aPtr = initMemory<T>(type, memsetType, data);
+  using namespace std::chrono_literals;
+  const std::chrono::duration<uint64_t, std::milli> delay = 100ms;
```
FMarno, line 408:
```suggestion
  constexpr std::chrono::milliseconds delay(100);
```

tests/catch/unit/memory/hipMemsetSync.cc
```diff
@@ -0,0 +1,505 @@
...
+
+template <typename T>
+static void devRegisteredCopy(memSetType memType, T* aPtr, T* hostMem, size_t dataW, size_t dataH,
+                              size_t dataD, size_t& dataPitch) {
+  size_t elementSize = sizeof(T);
+
+  switch (memType) {
+    case memSetType::hipMemset3D: {
+      hipMemcpy3DParms params{};
+      params.kind = hipMemcpyHostToHost;
```
FMarno, line 244: the kind is different to the other memcpy's I think it's likely wrong

tests/catch/unit/memory/hipMemsetSync.cc
```diff
@@ -0,0 +1,505 @@
...
+  }
+}
+
+// Copies device data to host and checks that each element is equal to the
+// specified value
+template <typename T>
+void verifyData(T* aPtr, size_t value, MultiDData& data, allocType type, memSetType memType) {
+  auto dataH = data.height == 0 ? 1 : data.height;
+  auto dataD = data.depth == 0 ? 1 : data.depth;
+  std::unique_ptr<T[]> hostPtr = std::make_unique<T[]>(data.pitch * dataH * dataD / sizeof(T));
```
FMarno, line 280: when copying onto the host for verification I think we should be using just the extent of the data instead of the extent with pitch included.

tests/catch/unit/memory/hipMemsetSync.cc
```diff
@@ -0,0 +1,505 @@
...
+  }
+
+
+  size_t idx;
+  bool allMatch = true;
+
+  for (size_t k = 0; k < dataD; k++) {
+    for (size_t j = 0; j < dataH; j++) {
+      for (size_t i = 0; i < data.width; i++) {
+        idx = data.pitch * dataH * k + data.pitch * j + i;
```
FMarno, line 300: again I think we should only be looking at the extent here so I think it would be:
```suggestion
        idx = data.width * dataH * k + data.width * j + i;
```
please double check yourself

FMarno (changes requested):

PR #2765, opened 2022-06-23
EXSWCPHIPT-125 - Add tests for hipGetDeviceCount and restructure some SpawnProc tests

jatinx:


tests/catch/include/hip_test_process.hh
```diff
@@ -53,18 +62,31 @@ class SpawnProc {
...
     exeName = dir.string();
+
+    INFO("Testing for exe exists: " << exeName);
+    REQUIRE(fs::exists(exeName));
+
     if (captureOutput) {
       auto path = fs::temp_directory_path();
       path /= getRandomString();
       tmpFileName = path.string();
+      INFO("Testing for capture file to not exists: " << tmpFileName);
```
FMarno:
```suggestion
      INFO("Testing that capture file does not exist already: " << tmpFileName);
```

tests/catch/include/hip_test_process.hh
```diff
@@ -53,18 +62,31 @@ class SpawnProc {
  public:
   SpawnProc(std::string exeName_, bool captureOutput_ = false)
       : exeName(exeName_), captureOutput(captureOutput_) {
-    auto dir = fs::path(TestContext::get().currentPath()).parent_path();
+    auto dir = fs::path(TestContext::get().currentPath());
     dir /= exeName;
     exeName = dir.string();
+
+    INFO("Testing for exe exists: " << exeName);
```
FMarno:
```suggestion
    INFO("Testing that exe exists: " << exeName);
```

tests/catch/unit/printf/printfSpecifiers.cc
```diff
@@ -89,7 +89,7 @@ x
 )here");
```
FMarno, line 89: update copyright header.

tests/catch/unit/device/hipGetDeviceCount.cc
```diff
@@ -32,3 +33,24 @@ TEST_CASE("Unit_hipGetDeviceCount_NegTst") {
...
+TEST_CASE("Unit_hipGetDeviceCount_HideDevices") {
+  int deviceCount = HipTest::getDeviceCount();
+  if (deviceCount < 2) {
+    HipTest::HIP_SKIP_TEST("This test requires more than 2 GPUs. Skipping.");
+    return;
+  }
+
+  for (int i = deviceCount; i >= 1; i--) {
+    std::string visibleStr;
+    for (int j = 0; j < i; j++) {  // Generate a string which has all devices except ith
```
FMarno: This comment is a little confusing since I don't see you skipping a value each time, the code seems to just include one less each time.

tests/catch/unit/device/getDeviceCount_exe.cc
```diff
@@ -0,0 +1,53 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+#include <hip/hip_runtime.h>
+
+// Expects 1 command line arg, which is the Device Visible String
+int main(int argc, char** argv) {
+  if (argc != 2) {
```
FMarno, line 28: you could use cout to add some detail about why the test might have failed.

tests/catch/unit/device/getDeviceCount_exe.cc
```diff
@@ -0,0 +1,53 @@
...
+  HIP_CHECK(hipInit(0));
+#else
+  unsetenv("ROCR_VISIBLE_DEVICES");
+  unsetenv("HIP_VISIBLE_DEVICES");
+  setenv("ROCR_VISIBLE_DEVICES", argv[1], 1);
+  setenv("HIP_VISIBLE_DEVICES", argv[1], 1);
+#endif
+
+  int count = 0;
+  if (hipSuccess != hipGetDeviceCount(&count)) return -1;
```
FMarno: if this fails then the env variables are not unset. I'm not sure which processes the environment variables are actually exported to, but you did take the time to unset them.

tests/catch/README.md
```diff
@@ -162,11 +162,11 @@ These macros are to be called in multi process tests, inside a process which get
 There is a special interface available for process isolation. ```hip::SpawnProc``` in ```hip_test_process.hh```. Using this interface test can spawn a process and place passing conditions on its return value or its output to stdout. This can be useful for testing printf output.
 Sample Usage:
 ```cpp
-hip::SpawnProc proc(<relative path of exe with test folder>, <optional bool value, if output is to be recorded>);
+hip::SpawnProc proc(<name of exe>, <optional bool value, if output is to be recorded>);
```
FMarno, line 165: Are relative paths not allowed now?

tests/catch/README.md
```diff
@@ -162,11 +162,11 @@ These macros are to be called in multi process tests, inside a process which get
...
 REQUIRE(0 == proc.run()); // Test of return value of the proc
 REQUIRE(exepctedOutput == proc.getOutput()); // Test on expected output of the process
 ```
-The process can be a standalone exe (see tests/catch/unit/printfExe for more information).
+The process can be a standalone exe, inside the same folder as other tests.
```
FMarno: is this can or must? also I don't think the "," is needed.

FMarno (commented):
Looks pretty good, just some small suggestions

tests/catch/README.md
jatinx: No, it was causing issues with standalone tests.

Its simple to have both the exe in same folder.

tests/catch/README.md
jatinx: I change it to "must"

tests/catch/include/hip_test_process.hh
jatinx: done

tests/catch/unit/device/getDeviceCount_exe.cc
jatinx: This is per process and should not matter after return statement.

tests/catch/unit/device/hipGetDeviceCount.cc
```diff
@@ -32,3 +33,24 @@ TEST_CASE("Unit_hipGetDeviceCount_NegTst") {
...
+      visibleStr += std::to_string(j);
+      if (j != (i - 1)) {
+        visibleStr += ",";
+      }
+    }
+
+    hip::SpawnProc proc("getDeviceCount");
```
FMarno: I think you need to add a flag here to capture the output.

FMarno (commented):
You've added the output to getDeviceCount_exe but not captured it in hipGetDeviceCount!

FMarno (dismissed):
LGTM

chrispaquot (approved):

PR #2766, opened 2022-06-23
EXSWCPHIPT-133 - Added support for offsetting to arrays to hipStreamValue tests

FMarno:
Changes to hipStreamWithCUMask were needed to allow c++17 to be used (the c++17 nodiscard attribute is on those function calls). C++17 was very helpful by allowing the use of if constexpr.

While making the changes I also fixed some small memory bugs and removed some narrowing casts that seemed unintentional.

NB: I've been unable to test these tests on an NVIDIA device that supports the calls, so there could be some problems.

### DIFF
--- a/tests/catch/unit/stream/hipStreamValue.cc
+++ b/tests/catch/unit/stream/hipStreamValue.cc
@@ -17,276 +17,377 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 */
 #include <hip_test_common.hh>
+#include <memory>
+#include <type_traits>
 
 constexpr unsigned int writeFlag = 0;
 
-#define DEFINE_HIP_STREAM_VALUE(TYPE, BITS, ...) hipStream##TYPE##Value##BITS(__VA_ARGS__)
+template <typename UIntT, typename... Args> auto waitFunc(Args... args) {
+  if constexpr (std::is_same<UIntT, uint32_t>::value) {
+    return hipStreamWaitValue32(args...);
+  } else {
+    return hipStreamWaitValue64(args...);
+  }
+};
 
-#define CHECK_HIP_STREAM_VALUE(TYPE, BITS, ...)                                                    \
-  HIP_CHECK(DEFINE_HIP_STREAM_VALUE(TYPE, BITS, __VA_ARGS__));
+template <typename UIntT, typename... Args> auto writeFunc(Args... args) {
+  if constexpr (std::is_same<UIntT, uint32_t>::value) {
+    return hipStreamWriteValue32(args...);
+  } else {
+    return hipStreamWriteValue64(args...);
+  }
+};
 
-#define NEG_TEST_ERROR_CHECK(TYPE, BITS, errorCode, ...)                                           \
-  HIP_CHECK_ERROR(DEFINE_HIP_STREAM_VALUE(TYPE, BITS, __VA_ARGS__), errorCode);
+// Random predefined 32 and 64 bit values
+using value32_t = std::integral_constant<uint32_t, 0x70F0F0FF>;
+using value64_t = std::integral_constant<uint64_t, 0x7FFF0000FFFF0000>;
+template <typename UIntT>
+using testValue =
+    typename std::conditional<std::is_same<UIntT, uint32_t>::value, value32_t, value64_t>::type;
 
-#if HT_AMD
-// Random predefiend 32 and 64 bit values
-constexpr uint32_t value32 = 0x70F0F0FF;
-constexpr uint64_t value64 = 0x7FFF0000FFFF0000;
 constexpr uint32_t DATA_INIT = 0x1234;
 constexpr uint32_t DATA_UPDATE = 0X4321;
 
-template <typename intT> struct TEST_WAIT {
-  using uintT = typename std::make_unsigned<intT>::type;
-  int compareOp;
-  uintT mask;
-  uintT waitValue;
-  intT signalValueFail;
-  intT signalValuePass;
+template <typename UIntT> struct TEST_WAIT {
+  static_assert(std::is_same<UIntT, uint32_t>::value or std::is_same<UIntT, uint64_t>::value,
+                "only implemented for 32 bit and 64 bit unsigned integers");
+  unsigned int compareOp;
+  UIntT mask = ~static_cast<UIntT>(0);
+  UIntT waitValue;
+  UIntT signalValueFail;
+  UIntT signalValuePass;
 
-  TEST_WAIT(int compareOp, uintT waitValue, intT signalValueFail, intT signalValuePass)
+  TEST_WAIT(unsigned int compareOp, UIntT waitValue, UIntT signalValueFail, UIntT signalValuePass)
       : compareOp{compareOp},
         waitValue{waitValue},
         signalValueFail{signalValueFail},
-        signalValuePass{signalValuePass} {
-    mask = static_cast<uintT>(0xFFFFFFFFFFFFFFFF);
-  }
+        signalValuePass{signalValuePass} {}
 
-  TEST_WAIT(int compareOp, uintT mask, uintT waitValue, intT signalValueFail, intT signalValuePass)
+  TEST_WAIT(unsigned int compareOp, UIntT mask, UIntT waitValue, UIntT signalValueFail,
+            UIntT signalValuePass)
       : compareOp{compareOp},
         mask{mask},
         waitValue{waitValue},
         signalValueFail{signalValueFail},
         signalValuePass{signalValuePass} {}
 };
-typedef TEST_WAIT<int32_t> TEST_WAIT32;
-typedef TEST_WAIT<int64_t> TEST_WAIT64;
+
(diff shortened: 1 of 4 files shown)


tests/catch/unit/stream/hipStreamValue.cc
```diff
@@ -190,102 +268,108 @@ void syncAndCheckData(hipStream_t stream, intT* dataPtr, int64_t* signalPtr, TES
...
-  HIP_CHECK(hipHostRegister(&(dataPtr.get()[0]), sizeof(intT), 0));
-  HIP_CHECK(hipHostRegister(&(dataPtr.get()[1]), sizeof(intT), 0));
+  HIP_CHECK(hipHostRegister(dataPtr.get(), sizeof(UIntT) * 2, 0));
+  std::fill(dataPtr.get(), dataPtr.get() + 2, DATA_INIT);
 
-  // Allocate Signal Memory
-  int64_t* signalPtr;
-  HIP_CHECK(hipExtMallocWithFlags((void**)&signalPtr, 8, hipMallocSignalMemory));
+  std::array<hipEvent_t, 2> events;
+  HIP_CHECK(hipEventCreate(&events[0]));
```
FMarno, line 327: matching hipEventDestroy is missing

chrispaquot:
Was this passing before on Nvidia?

FMarno:
> Was this passing before on Nvidia?

@chrispaquot Unfortunately we don't have an nvidia device that is capable of running these tests but I would guess no.

I have a found a few failing tests with the AMD backend. Do you know if the CI is actually running these tests or just skipping them?

tests/catch/unit/stream/hipStreamValue.cc
```diff
@@ -17,171 +17,261 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
   for (int device_id = 0; device_id < device_num; ++device_id) {
     HIP_CHECK(hipSetDevice(device_id));
-    waitValueSupport = 0;
-    HIP_CHECK(hipDeviceGetAttribute(&waitValueSupport, hipDeviceAttributeCanUseStreamWaitValue,
-                                    device_id));
+    int waitValueSupport = 0;
+    auto getAttributeError = hipDeviceGetAttribute(
+        &waitValueSupport, hipDeviceAttributeCanUseStreamWaitValue, device_id);
+    if (getAttributeError != hipSuccess) {
+      HipTest::HIP_SKIP_TEST("attribute not supported");
```
fabiomestre, line 87: HIP_SKIP_TEST is being called twice. Isn't it enough to call it only after this function returns false?

tests/catch/unit/stream/hipStreamValue.cc
```diff
@@ -17,171 +17,261 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
-  *signalPtr = 0x0;
-  *hostPtr64 = 0x0;
-  *hostPtr32 = 0x0;
-
-  SECTION("Registered host memory hipStreamWriteValue32") {
-    INFO("Test writting to registered host pointer using hipStreamWriteValue32");
-    HIP_CHECK(hipStreamWriteValue32(stream, hostPtr32.get(), value32, writeFlag));
-    HIP_CHECK(hipStreamSynchronize(stream));
-    HIP_ASSERT(*hostPtr32 == value32);
+  TestPtr(intT* ptr, B toFree) : toFree(std::move(toFree)), ptr(ptr) {}
```
fabiomestre: Why is ptr passed twice in here? toFree seems to be a unique_pointer which also contains ptr. toFree name is a bit misleading as well

tests/catch/unit/stream/hipStreamValue.cc
```diff
@@ -17,171 +17,261 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
 }
 
-// hipStreamWriteValue Tests
-TEST_CASE("Unit_hipStreamValue_Write") {
-  int64_t* signalPtr;
+enum class PtrType { HostPtr, DevicePtr, DevicePtrToHost, Signal };
 
-  hipStream_t stream{nullptr};
-  HIP_CHECK(hipStreamCreate(&stream));
+template <PtrType type, typename intT, typename B> struct TestPtr {
```
fabiomestre: Some documentation on how to use this struct would be helpful. Since it's not a POD, maybe it should be renamed to "class"

tests/catch/unit/stream/hipStreamValue.cc
```diff
@@ -17,171 +17,261 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
-  dataPtr[0] = DATA_INIT;
-  dataPtr[1] = DATA_INIT;
+  CAPTURE(offset);
 
+  // Allocate Host Memory
+  auto ptr = allocMem<ptrType, UIntT>();
+  UIntT* target = ptr.ptr + offset;
+  HIP_CHECK(writeFunc(stream, target, writeValue, writeFlag));
+  HIP_CHECK(hipStreamSynchronize(stream));
+  HIP_ASSERT(ptr.getValue(offset) == writeValue);
```
fabiomestre: Why use HIP_ASSERT here? Isn't REQUIRE better because it uses catch2? 

Note: There is other places with the same issue. So this comment would apply there as well.

tests/catch/unit/stream/hipStreamValue.cc
```diff
@@ -190,103 +280,112 @@ void syncAndCheckData(hipStream_t stream, intT* dataPtr, int64_t* signalPtr, TES
...
   }
+#if HT_AMD
+  HipTest::HIP_SKIP_TEST("EXSWCPHIPT-128");
+  return;
+#endif
+
+  using UIntT = typename TestType::UIntType;
+  constexpr auto ptrType = TestType::ptrType;
+  constexpr UIntT defaultMask = ~static_cast<UIntT>(0);
+  constexpr auto waitFunc = [](auto... args) {
```
fabiomestre: waitFunc and writeFunc are defined more than once in different tests. Could we define them globally instead?

tests/catch/unit/stream/hipStreamValue.cc
```diff
@@ -352,93 +471,86 @@ DEFINE_STREAM_WAIT_VAL_TEST_CASES_INT64("NoMask_And",
                                                     0x0F0F0F0F0F0F0F0F, 0X1F0F0F0F0F0F0F0F))
 DEFINE_STREAM_WAIT_VAL_TEST_CASES_INT64("NoMask_Nor",
                                         TEST_WAIT64(hipStreamWaitValueNor, 0x4724724747247247,
-                                                    static_cast<int64_t>(0xbddbddbdbddbddbd),
-                                                    static_cast<int64_t>(0xbddbddbdbddbddb3)))
+                                                    0xbddbddbdbddbddbd, 0xbddbddbdbddbddb3))
 #undef DEFINE_STREAM_WAIT_VAL_TEST_CASES_INT64
 
-#endif
+#define DEFINE_HIP_STREAM_VALUE(TYPE, BITS, ...) hipStream##TYPE##Value##BITS(__VA_ARGS__)
```
fabiomestre: This macros are not used anywhere anymore:

DEFINE_HIP_STREAM_VALUE
NEG_TEST_ERROR_CHECK

tests/catch/unit/stream/hipStreamValue.cc
FMarno: I am hoping that this will document why the attribute is not working. It might help someone that might have otherwise thought their device didn't support the attribute.

tests/catch/unit/stream/hipStreamValue.cc
FMarno: Depending on the test, the pointer that is used may be different to the pointer that is freed. The idea is to expose the ptr that is used with hipStreamWrite/WaitValue and then store a unique pointer with the deleter to simplify cleanup. I'm struggling a bit with what to name "toFree"

tests/catch/unit/stream/hipStreamValue.cc
FMarno: HIP_ASSERT was already in use so this an attempt to reduce changes, but I will just replace them now.

tests/catch/unit/stream/hipStreamValue.cc
FMarno: done

tests/catch/unit/stream/hipStreamValue.cc
FMarno: done

tests/catch/unit/stream/hipStreamValue.cc
FMarno: I've made it a class, renamed a couple things, and added some (hopefully) helpful documentation.

tests/catch/unit/stream/hipStreamValue.cc
fabiomestre: Fair enough

fabiomestre (dismissed):

chrispaquot (dismissed):

PR #2820, opened 2022-07-25
SWDEV-347383 - Remove code that sets -std and -fms-* flags

searlmc1:


yxsamliu (approved):

emankov (approved):
Totally agree with `-std=c++`, but who sets `-fms-*` now on the Windows path? hiprts, cmake, both?

searlmc1:
> Totally agree with `-std=c++`, but who sets `-fms-*` now on the Windows path? hiprts, cmake, both?

Per Sam, those -fms-* flags are included in clang so no need to set them separately.

PR #2848, opened 2022-08-04
Document of modern cpp features supported by the compiler

FMarno:
This document includes a non-complete list of modern C++ features that the hipcc compiler supports with examples.


scchan (approved):
LGTM

yxsamliu (approved):

PR #2849, opened 2022-08-06
fix logic for finding nvcc even if clang exists

### LINKED ISSUE #1650: Better way to detect hip_platform

acxz:
Issue: if a system has both `nvcc` and `clang++` (compiled w/o amd) then the HIP-PLATFORM autodetects to `amd`, instead of `nvidia`.

this PR fixes it by only checking for `clang++` compiled with amd with the provided alias of `amdclang++`.

Follow up of https://github.com/ROCm-Developer-Tools/HIP/pull/2623

Fixes: https://github.com/ROCm-Developer-Tools/HIP/issues/1650

Fixes comment: https://github.com/ROCm-Developer-Tools/HIP/issues/2256#issuecomment-818801109

cc: @mangupta

### DIFF
--- a/bin/hipvars.pm
+++ b/bin/hipvars.pm
@@ -121,7 +121,7 @@ if (defined $HIP_RUNTIME and $HIP_RUNTIME eq "rocclr" and !defined $HIP_ROCCLR_H
 }
 
 if (not defined $HIP_PLATFORM) {
-    if (can_run("$HIP_CLANG_PATH/clang++") or can_run("clang++")) {
+    if (can_run("$HIP_CLANG_PATH/clang++") or can_run("amdclang++")) {
         $HIP_PLATFORM = "amd";
     } elsif (can_run("$CUDA_PATH/bin/nvcc") or can_run("nvcc")) {
         $HIP_PLATFORM = "nvidia";


bin/hipvars.pm
```diff
@@ -121,7 +121,7 @@ if (defined $HIP_RUNTIME and $HIP_RUNTIME eq "rocclr" and !defined $HIP_ROCCLR_H
 }
 
 if (not defined $HIP_PLATFORM) {
-    if (can_run("$HIP_CLANG_PATH/clang++") or can_run("clang++")) {
```
scchan, line 124: I'd suggest changing it to `can_run(amdclang++)`, which is an alias to AMD's version of clang.
I'd recommend to our HIP team to run a more extensive set of tests to see if any app breaks due to this dependence

scchan (changes requested):

bin/hipvars.pm
acxz: What is the path of `amdclang++`? I don't seem to have that alias when I build rocm-llvm.

In any case, this PR is simply an atomic change removing system clang++. For changing `HIP_CLANG_PATH/clang++` you should open up another PR and I can rebase off of it.

acxz: Looking at your comment again, I understand what you meant now, basically replacing `clang++` with `amdclang++`.

acxz: Added your suggestion!

Snektron:
Can this be merged yet?

mangupta:
@Snektron : hipcc and corresponding dependencies such as hipvars.pm have been moved out of the hip repo. See #3216. These scripts are now part of the [hipcc project](https://github.com/ROCm-Developer-Tools/hipcc/).

acxz:
Thanks for mentioning that @mangupta
Closing this PR in favor of https://github.com/ROCm/HIPCC/pull/145

PR #2871, opened 2022-08-23
[WIP] Fix hipcc lib dir

awehrfritz:
This fixes various issues around the lib path in `hipvars.pm` and `hipcc.pl`:
- `hipvars.pm`: Detect the correct lib directory (i.e. 'lib64' or 'lib')
- `hipcc.pl`: Use the correct lib path in the linker arguments
- `hipcc.pl`: Set the `DEVICE_LIB_PATH` correctly for system installation (e.g. in /usr/lib64/amdgcn/bitcode/)

Note that this does not yet fix the issues around the include directories, which requires not only changes in HIP, but also ROCm-CompilerSupport.

@raramakr, this addresses some of the issues mentioned in https://github.com/ROCm-Developer-Tools/hipamd/issues/43#issuecomment-1222959814. Could you please have a look and merge those changes as you have been working on updating HIP to use GNUInstallDirs (https://github.com/ROCm-Developer-Tools/hipamd/commit/c92a12faf39210e24d329cc2c7e94dd124e69fed)?

### DIFF
--- a/bin/hipvars.pm
+++ b/bin/hipvars.pm
@@ -93,22 +93,19 @@ if ($isWindows) {
 # HIP_ROCCLR_HOME is used by Windows builds
 $HIP_ROCCLR_HOME=$ENV{'HIP_ROCCLR_HOME'};
 
-if (defined $HIP_ROCCLR_HOME) {
-    # Only used for Windows builds
+if ($isWindows) {
     $LIB = "lib";
-    $HIP_INFO_PATH = "$HIP_ROCCLR_HOME/lib/.hipInfo";
 } else {
-    # Detect lib directory (i.e. 'lib64' or 'lib')
-    if (-e "$HIP_PATH/lib64/.hipInfo") {
-        $LIB = "lib64";
-    } else {
-        $LIB = "lib";
-    }
-    $HIP_INFO_PATH = "$HIP_PATH/$LIB/.hipInfo";
+    # Look for the HSA runtime library to determine the lib directory name
+    $LIB = `find $HIP_PATH/lib $HIP_PATH/lib64 -name libhsa-runtime64* | grep -m1 -o '.*/lib.*/'`;
+    chomp($LIB); # Remove trailing newline characters
+    $LIB = basename($LIB); # Get the lib directory name
 }
-if (! -f $HIP_INFO_PATH) {
-    printf ("error: HIP_INFO_PATH does not exist ('$HIP_INFO_PATH')\n");
-    exit (-1);
+
+if (defined $HIP_ROCCLR_HOME) {
+    $HIP_INFO_PATH= "$HIP_ROCCLR_HOME/$LIB/.hipInfo";
+} else {
+    $HIP_INFO_PATH= "$HIP_PATH/$LIB/.hipInfo"; # use actual file
 }
 #---
 #HIP_PLATFORM controls whether to use nvidia or amd platform:
--- a/bin/hipvars.pm
+++ b/bin/hipvars.pm
@@ -97,10 +97,10 @@ if ($isWindows) {
     $LIB = "lib";
 } else {
     # Look for the AMD HIP library to determine the lib directory name
-    if (my @libs = glob("\E$HIP_PATH/lib*/*libamdhip64*.so*")) {
+    if (my @libs = glob("$HIP_PATH/lib*/libamdhip64*.so*")) {
         # Multi-lib layout (Fedora and others: /usr/lib64)
         $LIB = basename(dirname($libs[0]));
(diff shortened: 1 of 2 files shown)


awehrfritz:
@Mystro256, I would appreciate if you could have a look at this and forward it to the right people to get this fixed.

Mystro256:
Neat idea, but something to note about hipinfo is that the Debian packagers actually remove this file in their test packages, since it violates the Linux FHS (I think) and is not strictly required. Since Fedora has this policy too, I would advise against detecting the libdir using this logic, since it would be best to use "%exclude %{_libdir}/.hipInfo" in the spec file.

Alternatively the script can be updated to look for "libhsa-runtime64*" in /usr/lib and /usr/lib64, which should also work for Debian, since they use /usr/lib/x86_64-linux-gnu. I'm not sure how insane of an idea of doing this is:

find /usr/lib /usr/lib64 -name libhsa-runtime64* | grep -m1 -o '^/usr/lib.*/'

awehrfritz:
Good point about the hidden hipInfo file. Ideally we would get rid of those altogether and set those variables during the build process. By maybe that is another problem.

I was toying around with a similar idea of searching for the actual libraries already earlier. If I find the time I will revise the patches to do that. Also happfor somebody else to take the lead.

Mystro256:
@cgmb Do you know how the Debian guys deal with this? I figure it might have come up, but I'm less versed in HIP.

cgmb:
> hipcc.pl: Set the DEVICE_LIB_PATH correctly for system installation (e.g. in /usr/lib64/amdgcn/bitcode/)

It would be nice to also handle `/usr/lib/x86_64-linux-gnu/amdgcn/bitcode/`, which is where the files are currently placed on Debian. Would that be supported by this change?

awehrfritz:
@Mystro256 I have updated this branch now to adopt your suggestion to look for the HSA runtime library. This works well in my Fedora packages where `.hipInfo` and `.hipVersion` are not present anymore.

@cgmb at the moment, this patch does not handle the Debian case for `DEVICE_LIB_PATH`. With the latest change, I only replace the hardcoded `'lib'` to `$LIB` which corresponds `lib*/` without the trailing slash. It does not account for the additional multi-arch directory (i.e. `x86_64-linux-gnu`) in the Debian case. However, that could be added probably fairly easily, but as I am not really familiar with the Debian filesystem layout, I wasn't even aware of this.

The `$LIB` directory is also used for setting the search directories for the linker (via `-L` or `-rpath`). In the Fedora case, this leads to having `-L/usr/lib64` added in various places, which is as far as I can tell not needed at all since that is in the default search path already anyway. I'm at a loss with this in the Debian case. Should the linker arguments correspond then to `-L/usr/lib/x86_64-linux-gnu` in Debian's multi-arch setup?

bin/hipvars.pm
```diff
@@ -93,10 +93,19 @@ if ($isWindows) {
 # HIP_ROCCLR_HOME is used by Windows builds
 $HIP_ROCCLR_HOME=$ENV{'HIP_ROCCLR_HOME'};
 
+if ($isWindows) {
+    $LIB = "lib";
+} else {
+    # Look for the HSA runtime library to determine the lib directory name
+    $LIB = `find $HIP_PATH/lib $HIP_PATH/lib64 -name libhsa-runtime64* | grep -m1 -o '.*/lib.*/'`;
+    chomp($LIB); # Remove trailing newline characters
+    $LIB = basename($LIB); # Get the lib directory name
```
awehrfritz: @cgmb, we could change this part to the following in order to allow for the Debian layout to be caught by this as well. However, this has the potential to catch also some other weird cases, and I think it might be better to check on the different cases explicitly. I will push another change soon.
``` perl
    # Look for the AMD HIP library to determine the lib directory name
    $LIB = `find $HIP_PATH/lib $HIP_PATH/lib64 -name libamdhip64* | grep -m1 -o 'lib.*'`;
    chomp($LIB); # Remove trailing newline characters
    $LIB = dirname($LIB); # Get the lib directory name
```

awehrfritz: I have pushed another commit now which makes the checks more explicit and the code more readable.

cgmb (dismissed):
LGTM.

awehrfritz:
Thanks for your review and comments @cgmb. I have converted this PR to a draft since I realised that there are still various entangled issues in the perl scripts for hipcc related to an installation in `/usr`. I previously thought the issues around `lib` and `inlcude` paths were somewhat separate, but it's just one big mess there.

I have now a fully working version of all this, and was able to compile and run the HIP examples using only Fedora packages. But this required some major surgery, and will require local patches either way, as ROCM 5.2 doesn't work with upstream Clang 14 anymore and requires AMD's Clang or additional patches for hipcc.

I will push my changes when I find time to clean this all up.

PR #2880, opened 2022-08-23
EXSWHTEC-10 - Implement additional tests for hipIpcOpenMemHandle

mirza-halilcevic:
- Verify that opening the same handle in different contexts on same device causes an error
- Verify that opening the handle in the process that created it causes an error


chrispaquot (commented):
Would it be possible to make it work on Windows too?

mirza-halilcevic:
> Would it be possible to make it work on Windows too?

@chrispaquot The IPC APIs are only supported on Linux systems.

chrispaquot (dismissed):

chrispaquot (dismissed):

mangupta:
@mirza-halilcevic : Some tests currently failing. Shared details separately.

mirza-halilcevic:
> @mirza-halilcevic : Some tests currently failing. Shared details separately.

PR updated, you can proceed with the merge.

PR #2881, opened 2022-08-23
EXSWHTEC-9 - Implement additional tests for hipIpcCloseMemHandle

mirza-halilcevic:
- Verify that memory stays mapped when reference count is non zero
- Verify that memory stays mapped if handle is closed in a second process
- Verify that attempting to close handle in the process that created it causes an error.


chrispaquot (commented):
Is it working on Windows?

chrispaquot (commented):
This won't build on Windows I believe.

mirza-halilcevic:
> This won't build on Windows I believe.

@chrispaquot You're right. We added a Linux-only guard around the tests.

chrispaquot:
> > This won't build on Windows I believe.
> 
> @chrispaquot You're right. We added a Linux-only guard around the tests.

Can't we just remove it from the Windows build?

mirza-halilcevic:
> > > This won't build on Windows I believe.
> > 
> > 
> > @chrispaquot You're right. We added a Linux-only guard around the tests.
> 
> Can't we just remove it from the Windows build?

The test is now removed from the Windows build. Please proceed with the review.

chrispaquot (dismissed):

mangupta:
@mirza-halilcevic : Some tests fail CI. Details shared separately.

mirza-halilcevic:
> @mirza-halilcevic : Some tests fail CI. Details shared separately.

PR updated, you can proceed with the merge.

PR #2887, opened 2022-08-24
EXSWHTEC-5, EXSWHTEC-7 - Implement abstraction for multithreaded tests

mirza-halilcevic:
- Guarantees total ordering between parent and child thread.

### DIFF
--- a/tests/catch/include/threaded_zig_zag_test.hh
+++ b/tests/catch/include/threaded_zig_zag_test.hh
@@ -98,6 +98,11 @@ template <typename T> class ThreadedZigZagTest {
     HIP_CHECK_THREAD_FINALIZE();
   }
 
+  void TestPart1() const {}
+  void TestPart2() const {}
+  void TestPart3() const {}
+  void TestPart4() const {}
+
  private:
   std::mutex mtx_;
   std::condition_variable cv_;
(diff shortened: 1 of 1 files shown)


chrispaquot (dismissed):

tests/catch/include/threaded_zig_zag_test.hh
```diff
@@ -0,0 +1,110 @@
...
+
+    // Finalize
+    t.join();
+    HIP_CHECK_THREAD_FINALIZE();
+  }
+
+  void TestPart1() const {}
+  void TestPart2() const {}
+  void TestPart3() const {}
+  void TestPart4() const {}
```
chrispaquot, line 104: Don't they have to be virtual?

mirza-halilcevic: No. These are only default implementations in case a derived class doesn't need to implement one of the methods. If a derived class implements a method of the same name as one of the above it will shadow the parent class implementation and will be called instead. Virtual dispatch was not intended and is not necessary.

chrispaquot: I get it. Because it's static_cast<T*> that it doesn't need to be virtual.

PR #2904, opened 2022-08-29
EXSWHTEC-16 - Implement hipEventSynchronize tests

nives-vukovic:
- Add hipEventSynchronize.cc file with positive unit tests: Simple recorded event synchronization and without event recording
- Add newly created file into CMakeLists.txt


chrispaquot (commented):
Because all operations are made on the same stream (null), the test will pass w/o the stream synchronize. You need to have a different stream (non-blocking) for the kernel launch to test stream synchronization.

nives-vukovic:
@chrispaquot In these tests we just wanted to verify hipEventSynchronize API, as it is in the Event Management group. The point of the test was to see if further execution will be blocked until specified event is completed. If you think additional test with multiple streams or modifying this test would make the testing process more robust, we can expand it.

nives-vukovic:
@chrispaquot In these tests we mainly wanted to verify functionality of the hipEventSynchronize API, as extensive joint testing of multiple Event Management APIs is present in the Unit_hipEvent.cc. The point of the implemented test was to see if further execution will be blocked until specified event is completed, I have expanded it to include the case when kernel is launched in a different stream. Please review the changes.

chrispaquot (dismissed):

PR #2910, opened 2022-08-30
EXSWHTEC-17 - Implement additional hipStreamWaitEvent tests

nives-vukovic:
- Add simple test that waits for an event - Unit_hipStreamWaitEvent_Default
- Add negative test when stream is uninitialized - Unit_hipStreamWaitEvent_WithUninitializedStream
- Minor modification of Unit_hipStreamWaitEvent_DifferentStreams

### DIFF
--- a/tests/catch/unit/stream/hipStreamWaitEvent.cc
+++ b/tests/catch/unit/stream/hipStreamWaitEvent.cc
@@ -16,7 +16,13 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 */
-
+/*
+Testcase Scenarios :
+Unit_hipStreamWaitEvent_Negative - Test unsuccessful hipStreamWaitEvent when either event or flags are invalid
+Unit_hipStreamWaitEvent_UninitializedStream_Negative - Test unsuccessful hipStreamWaitEvent when stream is uninitialized
+Unit_hipStreamWaitEvent_Default - Test simple waiting for an event with hipStreamWaitEvent api
+Unit_hipStreamWaitEvent_DifferentStreams - Test waiting for an event on a different stream with hipStreamWaitEvent api
+*/
 
 #include <hip_test_common.hh>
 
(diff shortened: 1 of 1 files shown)


nives-vukovic:
Marko Veniger, Mirza Halilcevic, please review.

marko-veniger:
OK to be moved to Open status, reviewed by Marko Veniger

tests/catch/unit/stream/hipStreamWaitEvent.cc
```diff
@@ -59,6 +65,19 @@ TEST_CASE("Unit_hipStreamWaitEvent_Negative") {
   }
 }
 
+#if !HT_NVIDIA
```
rorake, line 69: Please put a comment why this test is disabled on Nvidia. If there is JIRA defect created for that then mention that in the comment.

tests/catch/unit/stream/hipStreamWaitEvent.cc
```diff
@@ -74,6 +93,35 @@ __global__ void waitKernel(int clockRate, int seconds) {
...
+  hipEvent_t waitEvent{nullptr};
+
+  HIP_CHECK(hipStreamCreate(&stream));
+  HIP_CHECK(hipEventCreate(&waitEvent));
+
+  REQUIRE(stream != nullptr);
+  REQUIRE(waitEvent != nullptr);
+
+  hipDeviceProp_t prop{};
+  HIP_CHECK(hipGetDeviceProperties(&prop, 0));
```
rorake: Instead of using hard-coded 0th device, get the default device using hipGetDevice & use it in hipGetDeviceProperties.

nives-vukovic: @rorake Code is refactored according to the comments. Please review.

rorake (dismissed):

tests/catch/unit/stream/hipStreamWaitEvent.cc
chrispaquot: I would like to understand why this happens and try to solve the issue instead of avoiding it on NV.

nives-vukovic: This define has been set following the example of hipStreamQuery tests. The API is not supposed to cause Segmentation fault, as this is not the observed behavior when CUDA API is used. Defect is created to address the issue.

mangupta (approved):

PR #2922, opened 2022-09-05
EXSWHTEC-40 - Implement additional Stream Management negative tests

nives-vukovic:
- Implement negative test for hipStream<Write/Wait>Value when stream is uninitialized
- Implement negative test for hipStreamSynchronize when stream is uninitialized

### DIFF
--- a/tests/catch/unit/stream/hipStreamValue.cc
+++ b/tests/catch/unit/stream/hipStreamValue.cc
@@ -499,7 +499,40 @@ TEST_CASE("Unit_hipStreamValue_Negative_InvalidMemory") {
   HIP_CHECK(hipStreamDestroy(stream));
 }
 
-TEMPLATE_TEST_CASE("Unit_hipStreamWaitValue_Negative_InvalidFlag", "", uint32_t, uint64_t) {
+TEMPLATE_TEST_CASE("Unit_hipStreamValue_Negative_UninitializedStream", "", uint32_t, uint64_t) {
+  if (!streamWaitValueSupported()) {
+    HipTest::HIP_SKIP_TEST("hipStreamWaitValue not supported on this device.");
+    return;
+  }
+
+  hipStream_t stream{reinterpret_cast<hipStream_t>(0xFFFF)};
+
+  // Allocate Host Memory
+  auto hostPtr = std::make_unique<TestType>();
+
+  // Register Host Memory
+  HIP_CHECK(hipHostRegister(hostPtr.get(), sizeof(TestType), 0));
+
+  // Set dummy data
+  *hostPtr = 0x0;
+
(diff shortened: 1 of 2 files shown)


nives-vukovic:
Marko Veniger, Dino Music, please review.

marko-veniger:
OK to be moved to Open status, reviewed by Marko Veniger

tests/catch/unit/stream/hipStreamValue.cc
```diff
@@ -512,7 +512,44 @@ TEST_CASE("Unit_hipStreamValue_Negative_InvalidMemory") {
   HIP_CHECK(hipStreamDestroy(stream));
 }
 
-TEMPLATE_TEST_CASE("Unit_hipStreamWaitValue_Negative_InvalidFlag", "", uint32_t, uint64_t) {
+TEMPLATE_TEST_CASE("Unit_hipStreamValue_Negative_UninitializedStream", "", uint32_t, uint64_t) {
+#if HT_AMD
```
chrispaquot: Can we remove this and just not have the test run (using json)?

nives-vukovic: @chrispaquot This approach was used because it was the way already implemented tests have been disabled. I don't have access to the jira defect, so I don't know if its content affected the choice of the disable method. If you have more information about the defect and could let me know if it still active, I can switch the json config file instead.

chrispaquot: @mangupta can you help here?

mangupta: I believe EXSWCPHIPT-96 was resolved via https://github.com/rocm-developer-tools/hipamd/commit/45a715c877391460fd02eebd2ebf27ab80b4a6a8. So we should be able to enable these tests on AMD as well.

mangupta:
@nives-vukovic  : I see some merge conflicts. Could you rebase your changes on top of current develop branch.

tests/catch/unit/stream/hipStreamValue.cc
nives-vukovic: @mangupta I have rebased the changes, however now when the tests have been enabled on AMD some of them fail and defects will have to be created

mangupta: @nives-vukovic : I don't see any failures with your updated patch. Can you please cross-check once more?

nives-vukovic: @mangupta I have run the tests once more, and several negative tests fail. However, we are currently only able to test it using rocm-5.2.3, so it is possible that these issues have been solved on develop.

PR #2941, opened 2022-09-15
Use llvm-readelf rather than readelf

searlmc1:


bin/hipcc.pl
```diff
@@ -475,7 +475,7 @@ BEGIN
             my $fileType = `file $obj`;
             my $isObj = ($fileType =~ m/ELF/ or $fileType =~ m/COFF/);
             if ($fileType =~ m/ELF/) {
-                my $sections = `readelf -e -W $obj`;
+                my $sections = `llvm-readelf -e -W $obj`;
```
mangupta: I don't think llvm-readelf is present in any of the folders of the default $PATH. At least not on my system. So maybe we should change this to $HIP_CLANG_PATH/llvm-readelf?
CC: @yxsamliu, @david-salinas

keryell: It comes with the `llvm` package on Debian for example.

david-salinas: I agree with Maneesh, we should use $HIP_CLANG_PATH here.  Because this ensures (in theory) that this hipcc.pl uses the same llvm-readelf that is was packaged/installed with this hipcc.pl's ROCm package/install directory.  The linux distro's LLVM may not be on the same level that is needed by this invocation of hipcc.pl.

mangupta: Also distro's LLVM package might not be installed on the system. So you need to add it to the dependencies for ROCm packages to ensure that it is available when hipcc.pl is invoked.

searlmc1: Hi @mangupta , @david-salinas ,

PR updated to use $HIP_CLANG_PATH/llvm-readelf ; "distro's llvm package" - not sure if I'm following; ROCm's llvm-readelf should be used. ROCm's llvm is installed as part of ROCm; hipcc already relies on pieces of llvm so I believe the dependencies are already setup; let me know if I am missing something.

mangupta: @searlmc1 : The "distro's llvm package" was a response to the comment from @keryell . The patch looks fine now to me.

searlmc1: @mangupta - does it look fine enough to Approve it ? 

:)

mangupta (approved):

PR #2972, opened 2022-09-30
EXSWHTEC-65  - Implement additional Initialization and Version tests

nives-vukovic:
- Split Unit_hipDeviceGetUuid and Unit_hipRuntimeGetVersion_NegAndValTst into positive and negative tests
- Expand hipDeviceComputeCapability negative test with case when device is invalid
- Expand hipDeviceGetP2PAttribute negative test with case when src and dst devices are the same and device is out of bounds
- Implement hipDeviceGetPCIBusId negative test when the bus id is partially filled into a char array


music-dino:
O.K. to be moved to open status.

chrispaquot (dismissed):

mangupta:
@nives-vukovic : Some tests fail on CI. Details shared separately.

PR #2976, opened 2022-10-05
CHIP-SPV Integration + Standalone tests 

pvelesko:
This branch includes the CHIP-SPV integration and also adds the option for compiling the unit tests as standalone executables. 

In case of CHIP-SPV, we are JITting all the kernels at startup so in order to run a single test in the suite, all the kernels need to be compiled first leading to slow testing. 

The standalone test options can be controlled via CMake option `STANDALONE_TESTS`. This will modify the `hip_add_exe_to_target` function so the changes to the original testing workflow are minimal. 

This PR still requires some work. In particular, the standalone `hip_add_exe_to_target` expects the unit test to be self-contained in a single source file. There are cases where you build something like `common.cc` - those functions need to be converted to inline .hh 

I'd like to get some feedback first, @gargrahul 

- [ ] get rid of `CMake Warning contains no tests` warnings
- [ ] Workaround for tests that are not contained in a single .cc file
- [ ] Compared to using the latest Catch2 as a submodule (using Catch::Catch2WithMain target), the current unit tests perform test discovery every single time you run ctest. This takes a long time and would be nice to have an option to register at compilation time.


b-sumner:
@pvelesko I'm not sure I agree with treating SPIR-V as a platform.  I prefer to think of it as an additional offload-arch, so, e.g., on a given platform one could use '--offload-arch=ISAarch --offload-arch=spirv64' allowing the binary to carry both platform+device specific ISA and SPIR-V (possibly using platform specific SPIR-V extensions) to be selected at runtime depending on what the system includes.

PR #2978, opened 2022-10-06
EXSWHTEC-94 - Implement helper classes and functions for memory tests

music-dino:
- Implement RAII resource guards for linear allocations 
- Implement RAII resource guards for stream
- Implement utility kernels for setting, incrementing and delayed execution
- Implement utility function for checking device attributes
- Implement utility functions for checking array contents

### DIFF
--- a/tests/catch/unit/memory/CMakeLists.txt
+++ b/tests/catch/unit/memory/CMakeLists.txt
@@ -183,5 +183,4 @@ endif()
 
 hip_add_exe_to_target(NAME MemoryTest
                       TEST_SRC ${TEST_SRC}
-                      TEST_TARGET_NAME build_tests
-                      COMPILE_OPTIONS -std=c++14)
+                      TEST_TARGET_NAME build_tests)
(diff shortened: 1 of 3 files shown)


chrispaquot (dismissed):

tests/catch/unit/memory/CMakeLists.txt
```diff
@@ -183,5 +183,4 @@ endif()
 
 hip_add_exe_to_target(NAME MemoryTest
                       TEST_SRC ${TEST_SRC}
-                      TEST_TARGET_NAME build_tests
-                      COMPILE_OPTIONS -std=c++14)
+                      TEST_TARGET_NAME build_tests)
```
mangupta, line 186: Any reason why we need to remove c++14 here?

music-dino: I noticed that tests/catch/CMakeLists sets the C++ version as C++17, with the stated intent in a comment being "enforce C++17 for all tests". The removed line overrode that setting for memory tests. 
In addition, this pull request uses a fold expression in DeviceAttributesSupport, and a constexpr lambda, both C++17 features. Future pull requests will use features like if constexpr and structured bindings as well.

mangupta: Thanks for the clarification. I am resolving some of the build failures seen by removing this flag override. Once it is done, this PR should pass tests and be ready to merge.

PR #3002, opened 2022-10-17
EXSWHTEC-81 - Implement tests for hipGetSymbolAddress and hipGetSymbolSize

music-dino:
- Negative parameter tests for hipGetSymbolAddress
- Negative parameter tests for hipGetSymbolSize
- Positive tests for fetching symbol address and size with symbols of several different types.


marko-veniger (dismissed):

tests/catch/unit/memory/hipGetSymbolSizeAddress.cc
```diff
@@ -0,0 +1,115 @@
...
+  std::array<T, N> read_buffer;
+  HIP_CHECK(hipMemcpy(read_buffer.data(), symbol_ptr, symbol_size, hipMemcpyDeviceToHost));
+  ArrayFindIfNot(read_buffer.data(), expected_value, read_buffer.size());
+}
+
+#if HT_AMD
+#define SYMBOL(expr) &HIP_SYMBOL(expr)
+#else
+#define SYMBOL(expr) HIP_SYMBOL(expr)
+#endif
```
chrispaquot, line 83: Why is there a difference here?

music-dino: The behavior differs according to platform. On Nvidia, HIP_SYMBOL is replaced with a value of symbol pointer type, on AMD a value of symbol type. This addresses the discrepancy.

PR #3003, opened 2022-10-17
EXSWHTEC-83 - Implement new and reimplement existing tests for hipMemAdvise

music-dino:
- Negative parameter tests
- Validate that setting and resetting of flags works as expected
- Validate that setting one flag does not lead to the resetting of another
- Validate round down for base address and roundup for base address + count behavior
- Validate that memory can still be read and write after applying each advice
- Validate that prefetching after setting flags does not interfere with them.


chrispaquot (dismissed):

mangupta (dismissed):

mangupta:
@music-dino : Some tests fail on CI. Details shared separately.

music-dino:
> @music-dino : Some tests fail on CI. Details shared separately.

PR updated, please proceed.

chrispaquot (approved):

PR #3053, opened 2022-11-01
EXSWHTEC-80 - Implement tests for hipModuleGetGlobal

music-dino:
- Implement positive tests to check basic behavior for globals of different types
- Implement positive and negative parameter tests.


gargrahul:

PR #3074, opened 2022-11-15
EXSWHTEC-149 - Implement tests for hipExternalMemoryGetMappedBuffer for the Vulkan API

music-dino:
- Test that mapped memory can be read from and written to
- Test if using mapped memory offset works correctly
- Negative parameter tests.


gargrahul:

PR #3076, opened 2022-11-15
EXSWHTEC-146 - Implement tests for hipSignalExternalSemaphoresAsync for the Vulkan API

music-dino:
- Test signaling a binary semaphore
- Test signaling a timeline semaphore(Nvidia only)
- Test signaling two semaphores
- Negative parameter tests.


gargrahul:

PR #3077, opened 2022-11-15
EXSWHTEC-134 - Implement tests for hipGLGetDevices

mirza-halilcevic:
- Basic positive test
- Positive parameter tests
- Negative parameter tests.

PR #3086, opened 2022-11-15
EXSWHTEC-141 - Implement tests for hipGraphicsUnregisterResource 

mirza-halilcevic:
- Basic negative test.

Depends-on: #3077


chrispaquot (commented):
Do these tests pass? I am not sure we support EGL.

mirza-halilcevic:
> Do these tests pass? I am not sure we support EGL.

I added the option to choose between a headless EGL context and a regular GLUT context via the environment variable GL_CONTEXT_TYPE. By default, GLUT will be used.

The tests do not pass with EGL on AMD, but they pass on NVIDIA. There seem to be some issues with GLUT as well. I will file a defect that explains in further detail.

mirza-halilcevic:
PR moved to: https://github.com/ROCm-Developer-Tools/hip-tests/pull/69

PR #3128, opened 2023-01-09
SWDEV-355608 - deprecate/cleanup hipcc link flags

scchan:
- deprecate -use-staticlib, -use-sharedlib which no longer provide any functional values
- use --hip-link instead of specifying the HIP runtime by name when linking
- fix linker option bug in HIT test's cmake
- update build options for unit tests requiring pthread or rt

Change-Id: Ib49978773c80fb40c71dc52b050ce921943ee3e4


scchan:
Rebased the patch in https://github.com/ROCm-Developer-Tools/HIP/pull/3093 and added minor updates

gargrahul:
@scchan Please check the CI failures.

scchan:
@gargrahul I fixed the Unit_hipEventMgpuMThreads.cc test, which should be linking explicitly to the pthread library.
I don't think the Windows test failure is related to my change though.

yxsamliu (dismissed):
LGTM. Thanks

gargrahul (approved):

PR #3153, opened 2023-02-05
CHIP-SPV Integration

pvelesko:
Changes enabling the use of SPIR-V backend

### DIFF
--- a/include/hip/channel_descriptor.h
+++ b/include/hip/channel_descriptor.h
@@ -28,12 +28,22 @@ THE SOFTWARE.
 // on NVCC path:
 
 
-#if (defined(__HIP_PLATFORM_HCC__) || defined(__HIP_PLATFORM_AMD__)) && !(defined(__HIP_PLATFORM_NVCC__) || defined(__HIP_PLATFORM_NVIDIA__))
+#if (defined(__HIP_PLATFORM_HCC__) || defined(__HIP_PLATFORM_AMD__)) &&                            \
+    !(defined(__HIP_PLATFORM_NVCC__) || defined(__HIP_PLATFORM_NVIDIA__)) &&                       \
+    !(defined(__HIP_PLATFORM_CLANG__) || defined(__HIP_PLATFORM_SPIRV__))
 #include <hip/amd_detail/amd_channel_descriptor.h>
-#elif !(defined(__HIP_PLATFORM_HCC__) || defined(__HIP_PLATFORM_AMD__)) && (defined(__HIP_PLATFORM_NVCC__) || defined(__HIP_PLATFORM_NVIDIA__))
-#include <hip/nvidia_detail/nvidia_channel_descriptor.h>
-#else
-#error("Must define exactly one of __HIP_PLATFORM_AMD__ or __HIP_PLATFORM_NVIDIA__");
-#endif
 
-#endif
+#elif (defined(__HIP_PLATFORM_NVCC__) || defined(__HIP_PLATFORM_NVIDIA__)) &&                      \
+    !(defined(__HIP_PLATFORM_HCC__) || defined(__HIP_PLATFORM_AMD__)) &&                           \
+    !(defined(__HIP_PLATFORM_CLANG__) || defined(__HIP_PLATFORM_SPIRV__))
+#include <hip/nvidia_detail/nvidia_channel_descriptor.h>
+
+#elif (defined(__HIP_PLATFORM_CLANG__) || defined(__HIP_PLATFORM_SPIRV__)) &&                      \
+    !(defined(__HIP_PLATFORM_HCC__) || defined(__HIP_PLATFORM_AMD__)) &&                           \
+    !(defined(__HIP_PLATFORM_NVCC__) || defined(__HIP_PLATFORM_NVIDIA__))
+#include <hip/spirv_channel_descriptor.h>
+
+#else
+#error("Must define exactly one of __HIP_PLATFORM_AMD__, __HIP_PLATFORM_NVIDIA__ or __HIP_PLATFORM_SPIRV__");
+#endif // HIP PLATFORM SELECTION
+#endif // HIP_INCLUDE_HIP_CHANNEL_DESCRIPTOR_H
--- a/include/hip/hip_runtime_api.h
+++ b/include/hip/hip_runtime_api.h
@@ -78,7 +78,7 @@ typedef struct hipUUID_t {
 } hipUUID;
 
 //---
-// Common headers for both NVCC and HCC paths:
+// Common headers for SPIRV, NVCC and HCC paths:
 
 /**
  * hipDeviceProp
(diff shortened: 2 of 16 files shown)


b-sumner:
SPIR-V as a platform is not following the existing precedent.  And I'm not clear if the compiler generating the SPIR-V is expected to produce code using only KHR extensions.

Also, suppose one of existing platforms wanted to add SPIR-V support to existing ISA or PTX support?  Then I might want to consider SPIR-V as an --offload-arch instead of as a platform.

yxsamliu:
I have the same question as Brian. Is the 'SPIRV platform' intended to be a target-neutral platform that works for all devices supporting SPIRV, or is it intended to support SPIRV on Intel platform only? If it is the latter, then it is better called Intel platform.

This also raises a question about supporting SPIRV on other platforms, e.g. ROCm. We need to come up with a structure that allows SPIRV support for different platforms.

linehill:
> And I'm not clear if the compiler generating the SPIR-V is expected to produce code using only KHR extensions.

Practically the SPIR-V code emission is implementation detail in CHIP-SPV from the HIP users’ point of view. The SPIR-V code generation orchestrated by CHIP-SPV and the emitted code should be constrained by target’s (a CHIP-SPV backend, e.g. OpenCL) capabilities and the SPIR-V environment specification. 

> Is the 'SPIRV platform' intended to be a target-neutral platform that works for all devices supporting SPIRV, or is it intended to support SPIRV on Intel platform only?

From CHIP-SPV’s readme: “CHIP-SPV that aims to make HIP and CUDA portable to platforms which support SPIR-V as the device intermediate representation. Currently CHIP-SPV supports OpenCL and Level Zero as the low-level runtime alternatives”. 
In a sense, the CHIP-SPV is target neutral (to some degree*1) but mostly the development effort has been on supporting HIP applications to run on Intel devices  - so no guarantees yet the CHIP-SPV would run SPIR-V code on other devices.

*1: some capabilities are required from the target for modeling some HIP features - e.g. for OpenCL backend, including but not limited to: image support for textures, subgroup extension for __shfl* functions and coarse-grained SMV (at least) for presenting device allocations as pointers on the host side code.

> This also raises a question about supporting SPIRV on other platforms, e.g. ROCm.

I would imagine that each platform might have a different flavor for the SPIR-V code (e.g. ROCm would probably use SPV_AMD* extensions). Should/would HIP applications pack SPIR-V code objects for different platforms? How could we tell which platform SPIR-V code objects, embedded in HIP binaries (more specific: in clang offload bundles), are targeted at?

> We need to come up with a structure that allows SPIRV support for different platforms.

Perhaps, extend the Clang offload bundle’s [bundle entry ID](https://clang.llvm.org/docs/ClangOffloadBundler.html#id6) with a (mandatory?) target platform field?

My two cents on the platform naming: could we call it “CHIPSPV” or just “CHIP”?

pjaaskel:
Yes, the intention is to support all targets with SPIR-V support, and not be Intel specific. However, some features of HIP cannot be yet expressed with khr/ext extensions only (there are only a couple in the list), but we are in progress making them such. So - meanwhile some of the features work only if the required Intel extension is supported. Note, however, that if those features are not used by the HIP kernel, it won't utilize that extension. Having Intel or AMD "flavors" of the SPIR-V target is thus not wanted nor sensible since they will be obsoleted as soon as we standardize those extensions.

include/hip/channel_descriptor.h
```diff
@@ -28,12 +28,22 @@ THE SOFTWARE.
...
-#endif
+#elif (defined(__HIP_PLATFORM_CLANG__) || defined(__HIP_PLATFORM_SPIRV__)) &&                      \
+    !(defined(__HIP_PLATFORM_HCC__) || defined(__HIP_PLATFORM_AMD__)) &&                           \
+    !(defined(__HIP_PLATFORM_NVCC__) || defined(__HIP_PLATFORM_NVIDIA__))
+#include <hip/spirv_channel_descriptor.h>
+
+#else
+#error("Must define exactly one of __HIP_PLATFORM_AMD__, __HIP_PLATFORM_NVIDIA__ or __HIP_PLATFORM_SPIRV__");
+#endif // HIP PLATFORM SELECTION
+#endif // HIP_INCLUDE_HIP_CHANNEL_DESCRIPTOR_H
```
keryell, line 49: A lot of missing EOL.

include/hip/channel_descriptor.h
```diff
@@ -28,12 +28,22 @@ THE SOFTWARE.
 // on NVCC path:
 
 
-#if (defined(__HIP_PLATFORM_HCC__) || defined(__HIP_PLATFORM_AMD__)) && !(defined(__HIP_PLATFORM_NVCC__) || defined(__HIP_PLATFORM_NVIDIA__))
+#if (defined(__HIP_PLATFORM_HCC__) || defined(__HIP_PLATFORM_AMD__)) &&                            \
+    !(defined(__HIP_PLATFORM_NVCC__) || defined(__HIP_PLATFORM_NVIDIA__)) &&                       \
+    !(defined(__HIP_PLATFORM_CLANG__) || defined(__HIP_PLATFORM_SPIRV__))
```
yxsamliu, line 33: What is the purpose of `__HIP_PLATFORM_CLANG__` and under what situation is it defined?

include/hip/channel_descriptor.h
```diff
@@ -28,12 +28,22 @@ THE SOFTWARE.
...
 #include <hip/nvidia_detail/nvidia_channel_descriptor.h>
-#else
-#error("Must define exactly one of __HIP_PLATFORM_AMD__ or __HIP_PLATFORM_NVIDIA__");
-#endif
 
-#endif
+#elif (defined(__HIP_PLATFORM_CLANG__) || defined(__HIP_PLATFORM_SPIRV__)) &&                      \
+    !(defined(__HIP_PLATFORM_HCC__) || defined(__HIP_PLATFORM_AMD__)) &&                           \
+    !(defined(__HIP_PLATFORM_NVCC__) || defined(__HIP_PLATFORM_NVIDIA__))
+#include <hip/spirv_channel_descriptor.h>
```
yxsamliu, line 44: For HIP/AMD and HIP/NVCC, the header files containing implementation details are in a separate repo [hipamd](https://github.com/ROCm-Developer-Tools/hipamd). However, the installed header files form a unified directory structure:
```
include - hip+
                     |-amd_detail
                     |-nvidia_detail
```
 where include/hip contains generic HIP headers, include/hip/amd_detail contains AMD/HIP specific headers and include/hip/nvidia_detail contains NV/HIP specific headers.

If more platforms are supported, it is desirable to keep this directory structure consistent and clean.

Will SPIRV specific HIP headers be upstreamed to https://github.com/ROCm-Developer-Tools/hipamd ? If not, where will they be? Is it possible to keep them in include/hip/spirv_detail to keep the directory structure clean?

include/hip/hip_runtime_api.h
```diff
@@ -477,14 +477,35 @@ enum hipComputeMode {
...
+#endif // HIP PLATFORM SELECTION
+
+#ifdef _USE_HIPCOMMON_RUNTIME_API_
 
 #include <stdint.h>
 #include <stddef.h>
 #ifndef GENERIC_GRID_LAUNCH
 #define GENERIC_GRID_LAUNCH 1
-#endif
-#include <hip/amd_detail/host_defines.h>
```
yxsamliu, line 487: This changes the order of host_defines.h vs other headers and macros, which may have unexpected effect.

Better keep the original order by adding extra conditions to differentiate AMD and SPIRV.

include/hip/hip_runtime_api.h
```diff
@@ -696,6 +717,7 @@ enum hipLimit_t {
 // Stream per thread
 /** Implicit stream per application thread.*/
 #define hipStreamPerThread ((hipStream_t)2)
+#define hipStreamLegacy ((hipStream_t)3)
```
yxsamliu, line 720: why is this needed?

include/hip/hip_runtime_api.h
```diff
@@ -7360,11 +7378,13 @@ static inline hipError_t hipMallocManaged(T** devPtr, size_t size,
     return hipMallocManaged((void**)devPtr, size, flags);
 }
 
-#endif
-#endif
+#endif // __cplusplus
 
-#include <hip/amd_detail/amd_hip_runtime_pt_api.h>
+#if (defined(__HIP_PLATFORM_HCC__) || defined(__HIP_PLATFORM_AMD__)) && !(defined(__HIP_PLATFORM_NVCC__) || defined(__HIP_PLATFORM_NVIDIA__))
+#include "hip/amd_detail/amd_hip_runtime_pt_api.h"
```
yxsamliu, line 7384: why `<>` is changed to `""`?

include/hip/hip_runtime_api.h
```diff
@@ -7360,11 +7378,13 @@ static inline hipError_t hipMallocManaged(T** devPtr, size_t size,
...
+#endif // TODO
 
 #if USE_PROF_API
-#include <hip/amd_detail/hip_prof_str.h>
+#include "hip/amd_detail/hip_prof_str.h"
```
yxsamliu, line 7388: same question as above

include/hip/channel_descriptor.h
pvelesko: We define it for all our paths. It's here to match the AMD and NVIDIA equivalents `__HIP_PLATFORM_HCC__` and `__HIP_PLATFORM_NVCC__`

include/hip/channel_descriptor.h
pvelesko: > Will SPIRV specific HIP headers be upstreamed to https://github.com/ROCm-Developer-Tools/hipamd ? 

no plans for that. 

>  If not, where will they be?

In the current repo much like hipamd

>  Is it possible to keep them in include/hip/spirv_detail to keep the directory structure clean?

ok we can rename the directories as you requested.

include/hip/hip_runtime_api.h
pvelesko: It's defined for CUDA and gives you access to the legacy default queue. Why isn't it defined for AMD?

include/hip/hip_runtime_api.h
pvelesko: no good reason, will fix.

PR #3171, opened 2023-02-22
WIP Progress Documentation Refactor

saadrahim:
Please do not review formally but suggestions appreciated.

PR #3192, opened 2023-03-27
Doxysphinx docs

Maetveis:
Replaces: #3171

~~Requires https://github.com/RadeonOpenCompute/rocm-docs-core/pull/36~~ (merged)


saadrahim:
@Maetveis the RTD build is not passing. Please help solve that.

saadrahim:
I suspect the builds are failing as you separated out the doxygen fixes. Perhaps it was wrong advice to suggest that.

build-docs.sh
saadrahim, line 1: Should this be removed? I don't think it is necessary.

docs/.doxygen/Doxyfile
```diff
@@ -32,19 +32,19 @@ DOXYFILE_ENCODING      = UTF-8
...
 
-PROJECT_NAME           = "HIP: Heterogenous-computing Interface for Portability"
+PROJECT_NAME           = "HIP Runtime API Reference"
 
 # The PROJECT_NUMBER tag can be used to enter a project or revision number. This
 # could be handy for archiving the generated documentation or if some version
 # control system is used.
 
-PROJECT_NUMBER         =
+PROJECT_NUMBER         = v3.0.1.0
```
saadrahim: Why is there a version number here?

build-docs.sh
Maetveis: Agreed, it was part of the original PR that I took over, removed.

docs/.doxygen/Doxyfile
Maetveis: Removed now. I reviewed the changes to the doxyfile to only change relevant values.

chrispaquot (approved):

PR #3221, opened 2023-05-02
Restoring doxygen files

saadrahim:
This should fix hip-docs package builds.


gargrahul:
Merging it to check internal builds.

PR #3239, opened 2023-05-31
Fixes related to distro's filesystems

Mystro256:
When packaging hipamd with Fedora, these are some issues that came up.

They seem generic to most distros, so we should pull them in if there's no issue.


Mystro256:
Moved to:
https://github.com/ROCm-Developer-Tools/HIPCC/pull/83

PR #3242, opened 2023-05-31
[doc] documentation updates and standardization changes

samjwu:
update dependencies
add version number to documentation
rename .sphinx/.doxygen to sphinx/doxygen
enable htmlzip, pdf, epub formats when publishing on Read the Docs 
add an "About" section with a license page
removes docBin from URL


samjwu:
Made CR in Gerrit to do this

PR #3319, opened 2023-08-29
delete workspace before starting the build

arjun-raj-kuppala:


.jenkins/Jenkinsfile
```diff
@@ -1,7 +1,9 @@
 def hipBuildTest(String backendLabel) {
         node(backendLabel) {
           stage("SYNC - ${backendLabel}") {
-
+            // Delete the workspace
+            sh "rm ${env.WORKSPACE}/* -fr"
```
keryell: Using consistent coding style?
```suggestion
            sh "rm -fr ${env.WORKSPACE}/*"
```

keryell: Note that this might not delete files and directories starting with '.', which might be an issue.

PR #3405, opened 2024-02-02
SWDEV-427628 - Documentation restructure

MathiasMagnus:
This PR fixes the associated JIRA ticket. It is broken into multiple commits for easier review.

PR #3318, opened 2023-08-24
HIP Programming model

MathiasMagnus:

PR #3432, opened 2024-03-18
Add Programming model documentation

MathiasMagnus:
This PR succeeds #3318 and builds on top of #3405, adds a single commit on top the doc reorg.


docs/data/understand/programming_model/cdna3_cu_dark.png
MKKnorr, line 1: This image is not used currently, and has some additional notes compared to the light version

docs/data/understand/programming_model/rdna3_cu.drawio
MKKnorr, line 1: This file seems to be empty :) But I guess it's a placeholder to replace rdna3_cu.png

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+Threads) onto various architectures, primarily GPUs. While the model may be
+expressed in most imperative languages, (eg. Python via PyHIP) this document
+will focus on the original C/C++ API of HIP.
+
+Threading Model
+===============================================================================
+
+The SIMT nature of HIP is captured by the ability to execute user-provided
+device programs, expressed as single-source C/C++ functions or sources compiled
+online/offline to binaries in bulk.
```
MKKnorr, line 21: What does "in bulk" mean here? Is it relating to the fact that it is executed in a highly parallel way? That might be more of a property of the gpus, not HIP.

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+
+Multiple instances of the device program (aka. kernel) may execute in parallel,
```
MKKnorr: It might be useful defining the term thread here (which you basically already did). E.g.
```suggestion
Multiple instances of the device program (aka. kernel), so called threads, may execute in parallel,
```
but this suggestion seems a bit hard to read

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+all uniquely identified by a set of integral values which are referred to as
+thread IDs. The set of integers identifying a thread relate to the hierarchy in
```
MKKnorr: Being pedantic: It is not exactly clear what you mean here. If by thread ID you mean `threadIdx`, then they don't globally uniquely identify the thread. However if you mean the set of `blockIdx` and `threadIdx`, then this holds true, but I wouldn't refer to them as "thread IDs". Also `threadIdx` doesn't (fully) relate to the hierarchy in which the threads execute.

I'd suggest just ommiting the second part
```suggestion
all uniquely identified by a set of integral values. The set of integers identifying a thread relate to the hierarchy in
```

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+which threads execute.
+
+.. _inherent_thread_model:
+
+Inherent Thread Model
+-------------------------------------------------------------------------------
+
+The thread hiearchy inherent to how AMD GPUs operate manifest as depicted in
```
MKKnorr: Using simpler words and a typo:
```suggestion
The thread hierarchy inherent to how AMD GPUs operate is depicted in
```

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+:numref:`inherent_thread_hierarchy`.
+
+.. _inherent_thread_hierarchy:
+
+.. figure:: ../data/reference/programming_model/thread_hierarchy.svg
+  :alt: Diagram depicting nested rectangles of varying color. The outermost one
+        titled "Grid", inside sets of uniform rectangles layered on oneanother
+        titled "Block". Each "Block" containing sets of uniform rectangles
+        layered on oneanother titled "Warp". Each of the "Warp" titled
```
MKKnorr: "typo"
```suggestion
        titled "Grid", inside sets of uniform rectangles layered on one another
        titled "Block". Each "Block" containing sets of uniform rectangles
        layered on one another titled "Warp". Each of the "Warp" titled
```
Or thinking about it, shouldn't it be:
```suggestion
        titled "Grid", inside sets of uniform rectangles layered on top of each other
        titled "Block". Each "Block" containing sets of uniform rectangles
        layered on top of each other titled "Warp". Each of the "Warp" titled
```

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+
+Inherent Thread Model
+-------------------------------------------------------------------------------
+
+The thread hiearchy inherent to how AMD GPUs operate manifest as depicted in
+:numref:`inherent_thread_hierarchy`.
+
+.. _inherent_thread_hierarchy:
+
+.. figure:: ../data/reference/programming_model/thread_hierarchy.svg
```
MKKnorr, line 37: It would be nice to add in the description and alt text, that the arrows represent threads

Actually, when being pedantic: the blocks don't logically consist of warps, that's just how they are executed on the hardware. It would be more precise if the blocks contained the threads directly. This could also convey the wrong idea, that the amount of threads in a block has to be a multiple of a warp (which is of course advisable, but by no means necessary)

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+        layered on oneanother titled "Warp". Each of the "Warp" titled
+        rectangles filled with downward pointing arrows inside.
+
+  Hierarchy of thread groups.
+
+* The innermost grouping is called a warp, or a wavefront in ISA terms. A warp
+  is the most tightly coupled groups of threads, both physically and logically.
+
+  When referring to threads inside a warp, they may be called lanes, and the
+  integral value identifying them the lane ID. Lane IDs aren't quieried like
```
MKKnorr: Typo
```suggestion
  integral value identifying them the lane ID. Lane IDs aren't queried like
```

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+  other thread IDs, but are user-calculated. As a consequence they are only as
+  multi-dimensional as the user interprets the calculated values to be.
+
+  The size of a warp is architecture dependent and always fixed. Warps are
+  signified by the set of communication primitives at their disposal, detailed
+  under :ref:`warp_cross_lane_functions`.
+
+* The middle grouping is called a block or thread block. The defining feature
+  of a block is that all threads in a block will share an instance of memory
+  which they may use to share data or synchronize with oneanother.
```
MKKnorr: Typo - several occurences in the file
```suggestion
  which they may use to share data or synchronize with one another.
```

MKKnorr: Is shared memory really the defining feature of a thread block? Blocks don't even have to use shared memory. Maybe the fact that the threads in a block "run in parallel" is a better defining feature. Of course shared memory is still an important aspect

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+
+  The size of a block is user-configurable but is maxmized by the queryable
+  capabilites of the executing hardware. The unique ID of the thread within a
```
MKKnorr: Typos
```suggestion
  The size of a block is user-configurable but is maximized by the queryable
  capabilities of the executing hardware. The unique ID of the thread within a
```
Or maybe better wording:

```suggestion
  The size of a block is user-configurable up to a maximum that depends
  on the executing hardware. The unique ID of the thread within a
```
Something along those lines, I might not have used the best wording. I just want to avoid the notion, that the block size is something that must be maximized. (Also it not only depends on the hardware, but also the resource requirements of the kernel, but that might be secondary here)

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+  block is 3-dimensional as provided by the API. When linearizing thread IDs
```
MKKnorr: Is this the appropriate point to actually mention what API function this is? This might help get a better connection between the model and what it actually looks like in code.

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+  within a block, assume the "fast index" being dimension ``x``, followed by
+  the ``y`` and ``z`` dimensions.
+
+* The outermost grouping is called a grid. A grid manifests as a single
+  dispatch of kernels for execution. The unique ID of each block within a grid
```
MKKnorr:
```suggestion
  dispatch of a kernel for execution. The unique ID of each block within a grid
```

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+  is 3-dimensional, as provided by the API and is queryable by every thread
```
MKKnorr: Again: should the actual function be mentioned here?

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+  within the block.
+
```
MKKnorr, line 76: For the upper section: Nice explanation, but I'm missing, that the hierarchy is just plainly stated: A grid is made up of (several) blocks, a block is made up of (several) threads.


The fact that threads are grouped in warps is important, but not a direct consequence of the thread model of hip.

This way the logic of the section could actually be reversed, explaining from top down - a grid consists of blocks, blocks consist of threads. But this does not necessarily have to be changed, the other approach is just as fine, whichever fits better.

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+The Cooperative Groups API introduces new APIs to launch, group, subdivide,
+synchronize and identify threads, as well as some predefined group-collective
+algorithms, but most importantly a matching threading model to think in terms
+of. It relaxes some of the restrictions of the :ref:`inherent_thread_model`
+imposed by the strict 1:1 mapping of architectural details to the programming
+model.
+
+The rich set of APIs introduced by Cooperative Groups allow the programmer
+to define their own groups based on run-time predicates, but a set of implicit
+groups manifest based on kernel launch parameters.
```
MKKnorr: I'm not sure I fully understand this "but" here - is it a restriction, or an additional set of implicit groups?

Also, but this might just be me again, the use of the word "to manifest" throws me off in the reading flow a bit

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+
+The thread hiearchy abstraction of Cooperative Groups manifest as depicted in
```
MKKnorr: Typo
```suggestion
The thread hierarchy abstraction of Cooperative Groups manifest as depicted in
```
And another "manifest" ;) if it's not for reading flow (which might be personal to me) at least replace some of them for the sake of reducing repetition :)

docs/reference/programming_model.rst
MKKnorr: Addendum: The passive voice might not be the best choice here, as it could also be misunderstood. Also the layering of the rectangles on top of each other might not be as important, the layout (4 rectangles layed out in 2 rows and 2 columns) also hasn't been explained, and I think that's not the important part this image has to convey, but rather the relation of blocks being contained in the grid and warps being contained in the block

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+:numref:`coop_thread_hierarchy`.
+
+.. _coop_thread_hierarchy:
+
+.. figure:: ../data/reference/programming_model/thread_hierarchy_coop.svg
+  :alt: Diagram depicting nested rectangles of varying color. The outermost one
+        titled "Grid", inside sets of different sized rectangles layered on
+        oneanother titled "Block". Each "Block" containing sets of uniform
+        rectangles layered on oneanother titled "Warp". Each of the "Warp"
+        titled rectangles filled with downward pointing arrows inside.
```
MKKnorr: This alt text is not fully accurate, the outermost rectangle is called "Multi Grid", and there are no warps in this picture (which is good that way imo).
I think this is just a copy-paste mistake, but just to make sure: this should convey, that the grids can have different amounts and layouts of blocks (here the layout is actually important, compared to the first figure)

The remarks from the first one also mostly apply - the passive voice makes this harder to understand.

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+
+  Cooperative group thread hierarchy.
+
+* Multi Grid is an abstraction of potentially multiple simultaneous launches of
+  the same kernel over multiple devices. Grids inside a multi device kernel
+  launch need not be of uniform size, thus allowing taking into account
+  different device capabilities and preferences.
+
+  .. deprecated:: 5.0
```
MKKnorr, line 114: The rendered note does not make it fully clear, which part is deprecated

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+
+* Same as the :ref:`inherent_thread_model` Grid entity. The ability to
+  synchronize over a grid requires the kernel to be launched using the
+  Cooperative Groups API.
+
+* The defining feature of a cluster or block cluster is that all threads in a
+  cluster will share a common set of distributed shared memory which they may
+  use to share data or synchronize with oneanother.
+
+* Same as the :ref:`inherent_thread_model` Block entity.
```
MKKnorr: The usage of "Same as [...]" does not say, *what* is actually the same

Using [definition lists](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#definition-lists) would make this way easier to understand than a simple bullet-point list. The Inherent Thread Model explanation would also benefit from this. This might require some text to be adjusted, but I think it would be worth it

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+
+.. note::
+
+  Explicit warp-level thread handling is absent from the Cooperative Groups API.
+  In order to exploit the known hardware SIMD width on which built-in
+  functionality translates to simpler logic, one may use the group partitioning
+  part of the API, typically but not necessarily ``tiled_partition``.
```
MKKnorr:
```suggestion
  part of the API, typically, but not necessarily, ``tiled_partition``.
```

Also: this is a weird statement. What else could be used for that? Is the API really that inconsistent? Not sure what to do here, but this threw me off, when reading it.

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+by the memory subsystem of GPUs. :numref:`memory_hierarchy` summarizes that memory namespaces and
+how they relate to the various levels of the threading model.
+
+.. _memory_hierarchy:
+
+.. figure:: ../data/reference/programming_model/memory_hierarchy.svg
+  :alt: Diagram depicting nested rectangles of varying color. The outermost one
+        titled "Grid", inside on the upper half a rectangle titled "Cluster".
+        Inside it are two identical rectangles titled "Block", inside them are
+        ones titled "Local" with multiple "Warp" titled rectangles. Blocks have
```
MKKnorr: Same criticism as earlier, this being written in passive makes it harder to understand.
Actually, this one has a good example of how to make it less ambiguous - by adding "it" after "inside", it already makes a huge difference. Compare:
```suggestion
  :alt: Diagram depicting nested rectangles of varying color. The outermost one
        titled "Grid", inside (of it) on the upper half (is) a rectangle titled "Cluster".
        Inside it are two identical rectangles titled "Block", inside them are
        ones titled "Local" with multiple "Warp" titled rectangles. Blocks have
```

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+        not just Local inside, but also rectangles titled "Shared". The Shared
+        rectangles of Blocks in the same Cluster are grouped together with a
+        translucent halo titled "Cluster shared". Outside the Cluster but
+        inside the Grid is a rectangle titled "Global" with three others
+        inside: "Surface", "Texture" (same color) and "Constant" (different
+        color).
+
+  Memory hierarchy.
+
```
MKKnorr, line 162: Same as earlier: a definition list gives more clarity on what part is being explained

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+* Local or per-thread memory is read-write storage only visible to the
+  threads defining the given variables. The size of a block for a given kernel,
+  the number of concurrent warps are limited by local memory usage.
```
MKKnorr:
```suggestion
* Local or per-thread memory is read-write storage only visible to the
  thread defining the given variables. The size of a block for a given kernel,
  and thereby the number of concurrently running warps, are limited by local memory usage.
```
Due to this explanation I'd argue, that the depiction in figure three is wrong - "Local" should not surround the warps (which probably shouldn't even be in that figure), but should only surround one single thread.

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+    and accessible through additional APIs.
+
+  * Surface is a writable version of texture memory.
+
+Execution Model
+===============================================================================
+
+HIP programs consist of two distinct scopes:
+
+* The host-side API running on the host processor. There are to APIs available:
```
MKKnorr: :)
```suggestion
* The host-side API running on the host processor. There are two APIs available:
```

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+  * The HIP runtime API which enables use of the single-source programming
+    model.
+
+  * The HIP driver API which sits at a lower level and most importantly differs
+    by removing some of the facilities provided by the runtime API, most
+    importantly around kernel launching and argument setting. It is geared
+    towards implementing abstractions atop, such as the runtime API itself.
+
+* The device-side kernels running on GPUs. Both the host and the device-side
+  APIs have synchronous and asynchronous functions in them.
```
MKKnorr: Since this is a nested list of each "two" points (two distinct "scopes", two "APIs") I'd think making this a bit more visually appealing could help avoiding mixing this up. A definition list or a numbered list for the "outer" list could help.

And also changing "The host-side API running on the host processor" to just "The host-side running on the host processor", so that it won't be confused with the immediately following "two APIs available"

Also it doesn't really help, that the earlier mentioned "scopes" are actually "APIs", which might make it easier to confuse the nested lists.

MKKnorr:
```suggestion
* The device-side kernels running on GPUs.

Both the host and the device-side APIs have synchronous and asynchronous functions in them.
```
This statement should be separated from the device-side, as it talks about both points.

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+
+Host-side execution
+-------------------------------------------------------------------------------
+
+The part of the host-side API which deals with device management and their
+queries are synchronous. All asynchronous APIs, such as kernel execution, data
+movement and potentially data allocation/freeing all happen in the context of
+device streams.
+
+Streams are FIFO buffers of commands to execute relating to a given device.
```
MKKnorr:
```suggestion
Streams are FIFO buffers of commands to execute on a given device.
```

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+Commands which enqueue tasks on a stream all return promptly and the command is
+executed asynchronously. All side-effects of a command on a stream are visible
+to all subsequent commands on the same stream. Multiple streams may point to
+the same device and those streams may be fed from multiple concurrent host-side
+threads. Execution on multiple streams may be concurrent but isn't required to
+be.
+
+Asynchronous APIs involving a stream all return a stream event which may be
+used to synchronize the execution of multiple streams. A user may enqueue a
+barrier onto a stream referencing an event. The barrier will will block until
```
MKKnorr: :)
```suggestion
barrier onto a stream referencing an event. The barrier will block until
```

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+the command related to the event does not complete, at which point all
```
MKKnorr: Shouldn't it be waiting until it completes?

```suggestion
the command related to the event does complete, at which point all
```

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+side-effects of the command shall be visible to commands following the barrier,
+even if those side-effects manifest on different devices.
```
MKKnorr: This "manifest" is actually fine, I think because here it is easy to figure out whether the verb or the noun is meant 😅

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+intended use-cases.
+
+* Using the triple-chevron ``<<<...>>>`` operator on a ``__global__`` annotated
+  function.
+
+* Using ``hipLaunchKernelGGL()`` on a ``__global__`` annotated function.
+
+  .. tip::
+
+    This name by default is a macro expanding to triple-chevron. In cases where
```
MKKnorr:
```suggestion
    ``hipLaunchKernelGGL()`` by default is a macro expanding to triple-chevron. In cases where
```

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+    language syntax extensions are undesirable, or where launching templated
+    and/or overloaded kernel functions define the
+    ``HIP_TEMPLATE_KERNEL_LAUNCH`` preprocessor macro before including the HIP
+    headers to turn it into a templated function.
+
+* Using the
+  :doxygen:`launch APIs supporting the triple-chevron syntax <Clang>` directly.
```
MKKnorr: This does not seem as if it's supposed to look this way. Especially in the rendered version.

docs/reference/programming_model.rst
```diff
@@ -0,0 +1,259 @@
...
+
+  .. caution::
+
+    These APIs are intended to be used/generated by tools such as the HIP
+    compiler itself and not intended towards end-user code. Should you be
```
MKKnorr: This feels slightly easier to understand
```suggestion
    compiler itself and not by end-user code. Should you be
```

docs/understand/programming_model.rst
```diff
@@ -0,0 +1,158 @@
...
+                between the programmer and the compiler/runtime executing the
+                code, how it maps to the hardware.
+  :keywords: AMD, ROCm, HIP, CUDA, API design
+
+*******************************************************************************
+Programming Model
+*******************************************************************************
+
+The HIP programming model makes it as easy as reasonably possible to map
+data-parallel C/C++ algorithms and map them to massively parallel, wide SIMD
```
MKKnorr: repetition
```suggestion
The HIP programming model makes it as easy as reasonably possible to map
data-parallel C/C++ algorithms to massively parallel, wide SIMD
```

docs/understand/programming_model.rst
```diff
@@ -0,0 +1,158 @@
...
+architectures, such as GPUs. As a consequence, one needs a basic understanding
+of the underlying device architecture to make efficient use of HIP and GPGPU
+(General Purpose Graphics Processing Unit) programming in general.
+
+RDNA & CDNA Architecture Summary
+===============================================================================
+
+Most GPU architectures, much like RDNA and CDNA have a hierarchical structure.
+The inner-most piece is some Single Instruction Multiple Data (SIMD) enabled
```
MKKnorr:
```suggestion
The inner-most piece is a Single Instruction Multiple Data (SIMD) enabled
```
I think it's okay to use the singular here, as this doesn't imply, that it's the only SIMD, just that it's one of many pieces. Alternatively you could switch piece with "building block"

docs/understand/programming_model.rst
```diff
@@ -0,0 +1,158 @@
...
+vector Arithmetic Logical Unit (ALU). Most recent GPUs beside the vector ALU
+also house some matrix ALU for accelerating algorithms of well defined shapes.
```
MKKnorr:
```suggestion
also house matrix ALUs for accelerating algorithms of well defined shapes.
```
Also, and this is nit-picking, I'm not sure if "well defined shapes" is the right wording. The shape is not that important once the matrices surpass the minimum size needed to use the matrix ALUs. Then not the whole operation might be offloaded to the matrix ALUs, but it would still "accelerate" them, even if only partially. "[...] for accelerating matrix multiplications." is the easiest way out, but then again, it doesn't add much information.

docs/understand/programming_model.rst
```diff
@@ -0,0 +1,158 @@
...
+
+Some number of vector and matrix ALUs comprise a larger block, often referred
```
MKKnorr:
```suggestion
A number of vector and matrix ALUs comprise a larger block, often referred
```
And maybe add "The exact number of vector and matrix ALUs per Compute Unit depends on the architecture."

docs/understand/programming_model.rst
```diff
@@ -0,0 +1,158 @@
...
+  Block Diagram of an RDNA3 Compute Unit.
+
+.. _cdna3_cu:
+
+.. figure:: ../data/understand/programming_model/cdna3_cu.png
+  :alt: Block diagram showing components: "Scheduler", "Local Data Share",
+        "Matrix Core Unit", "Shader Core" and "L1 Cache" along with some
+        unnamed blocks.
+
+  Block Diagram of an CDNA3 Compute Unit.
```
MKKnorr:
```suggestion
  Block Diagram of a CDNA3 Compute Unit.
```

docs/understand/programming_model.rst
```diff
@@ -0,0 +1,158 @@
...
+
+For hardware implementation's sake, some number of Multi Processors are grouped
```
MKKnorr:
```suggestion
For hardware implementation's sake, multiple Compute Units are grouped
```

docs/understand/programming_model.rst
```diff
@@ -0,0 +1,158 @@
...
+.. _rdna3_cu:
+
+.. figure:: ../data/understand/programming_model/rdna3_cu.png
+  :alt: Block diagram showing components mostly duplicated on the upper and
+        lower halves of the image with some spanning over both parts (hinting
+        at them being shared). Both the top and the bottom have two sets of
+        identical hardware blocks. a "Scheduler" having "Vector GPR"s with an
+        associated Vector ALU with the following noted capabilities:
+        Float/INT/Matrix SIMD32, Float/Matrix SIMD32, Transcendental SIMD8, AI
+        MATRIX Accelerator, DPFP (1). Both top and bottom have a single "Ray
```
MKKnorr:
```suggestion
        Float/INT/Matrix SIMD32, Float/Matrix SIMD32, Transcendental SIMD8, AI
        MATRIX Accelerator, double precision Floating-Point. Both top and bottom have a single "Ray
```
This list is a bit overkill for a screen reader imo, but I'm not too versed in that. But still, the "(1)"  definitely doesn't belong in there, and DPFP is an abbreviation I've never heard, and also hasn't been explained previously. If it means what I think it means, I would write that out, or just ommit it. The list could probably also be shortened, the capabilities of the SIMD32s are slightly redundant, at least when the list is all the information one is given

docs/understand/programming_model.rst
```diff
@@ -0,0 +1,158 @@
...
+Single Instruction Multiple Threads
+===============================================================================
+
+The SIMT programming model behind the HIP device-side execution is a
+middle-ground between SMT (Simultaneous Multi-Threading) programming known from
+multi-core CPUs, and SIMD (Single Instruction, Multiple Data) programming
+mostly known from exploiting relevant instruction sets on CPUs (eg.
+SSE/AVX/Neon).
+
+A HIP device compiler maps our SIMT code written in HIP C++ to an inherently
```
MKKnorr: This is a change of tone from the previous sections. "the" could also be ommited
```suggestion
A HIP device compiler maps (the) SIMT code written in HIP C++ to an inherently
```

docs/understand/programming_model.rst
```diff
@@ -0,0 +1,158 @@
...
+  {
+    int tid = threadIdx.x;
+    int bid = blockIdx.x;
+    int dim = blockDim.x;
+  
+    a[tid] += (tid + bid - dim) * b[tid];
+  }
+
+The incoming four-vector of floating-point values ``a`` is multiplied by a
+scalar and then multiplied element-wise by another four-vector. On modern
```
MKKnorr: This description is not right, `b` is multiplied by the scalar and then added to `a`.

Also I'm not sure if it's a good idea to introduce the vector types at the same time as trying to explain the concept of "vecotrizing" the code. That's an additional layer of parallelism added on top

docs/understand/programming_model.rst
```diff
@@ -0,0 +1,158 @@
...
+  :alt: Two large arrows pointing downward with blocks inside and ellipses
+        between the arrows. Inside the arrows the same series of blocks with
+        the following texts inside from top to bottom: "ADD", "DIV", "FMA",
+        "FMA", "FMA" and "FMA".
+
+  Instruction flow of the sample SIMT program.
+
+In HIP, lanes of a SIMD architecture are fed by mapping threads of a SIMT
+execution, one thread down each lane of a SIMD engine. Execution parallelism
+isn't exploited from the width of the built-in vector types, but via the thread
```
MKKnorr: This is where presenting ILP and TLP at the same time can become complicated - there are vector instructions, that can also be executed in parallel

docs/understand/programming_model.rst
```diff
@@ -0,0 +1,158 @@
...
+code are signified by the ``__host__`` and ``__device__`` decorators. There are
+a few key differences between the two:
+
+* The C++ abstract machine assumes a unified memory address space, meaning that
+  one can always access any given address in memory (assuming the absence of
+  data races). HIP however introduces several memory namespaces, an address
+  from one means nothing in another. Moreover not all address spaces are
+  accessible from all contexts.
+
+  If one were to look at {ref}`cdna2_gcd` and inside the {ref}`cdna3_cu`,
```
MKKnorr: These references are not being rendered correctly

docs/understand/programming_model.rst
```diff
@@ -0,0 +1,158 @@
...
+  by HIP aren't general purpose devices, at least not in the sense a CPU is.
+  HIP focuses on data parallel computations and as such caters to throughput
+  optimized architectures, such as GPUs or accelerators derived from GPU
+  architectures.
+
+* Asynchrony is at the forefront of the HIP API. Computations launched by HIP
+  execute asynchronously on the device and it is the user's responsibility to
+  synchronize their data dispatch/fetch with computations on the device. HIP
+  does perform implicit synchronization on occasions, but unlike some APIs
+  (OpenCL, SYCL) by and large places the onus of synchronization on the user.
```
MKKnorr: Simpler wording
```suggestion
  (OpenCL, SYCL) by and large places the responsibility of synchronization on the user.
```

docs/understand/programming_model.rst
```diff
@@ -0,0 +1,158 @@
...
+  some are very expensive (meaning: slow) to implement on GPU devices, therefor
+  they are forbidden in device contexts to avoid users tapping into features
+  unexpectedly decimating their program's performance. Offload devices targeted
+  by HIP aren't general purpose devices, at least not in the sense a CPU is.
+  HIP focuses on data parallel computations and as such caters to throughput
+  optimized architectures, such as GPUs or accelerators derived from GPU
+  architectures.
+
+* Asynchrony is at the forefront of the HIP API. Computations launched by HIP
+  execute asynchronously on the device and it is the user's responsibility to
```
MKKnorr:
```suggestion
* Asynchrony is at the forefront of the HIP API. Computations launched on the device
  execute asynchronously with respect to the host and it is the user's responsibility to
```
I think it's important to emphasize, that the execution is not asynchronously on the device tiself

docs/understand/programming_model.rst
```diff
@@ -0,0 +1,158 @@
...
+.. _cdna2_gcd:
+
+.. figure:: ../data/understand/programming_model/cdna2_gcd.png
+  :alt: Block diagram showing four "Compute Engine"s each with 28 "CU"s
+        (Compute Unit) inside. These four Compute Engines share one block of
+        "L2 Cache and Controllers". Around them are four "Memory Controller"s,
+        each having a "Memory Phy" block next to them. To the top and bottom of
+        all these are eight blocks of "Infinity Fabric Link" with one of the
+        eight reading "Infinity Fabric or PCIe". Two sole "VCN" blocks sit in
+        top corners. At the very bottom spans a colored seaction reading
```
MKKnorr: Typo
```suggestion
        top corners. At the very bottom spans a colored section reading
```

docs/understand/programming_model.rst
```diff
@@ -0,0 +1,158 @@
...
+  one can always access any given address in memory (assuming the absence of
+  data races). HIP however introduces several memory namespaces, an address
+  from one means nothing in another. Moreover not all address spaces are
+  accessible from all contexts.
+
+  If one were to look at {ref}`cdna2_gcd` and inside the {ref}`cdna3_cu`,
+  every Compute Unit has an instance of storage backing the namespace
+  ``__shared__``. Even if the host were to have access to these regions of
+  memory, the performance benefits of the segmented memory subsystem are
+  supported by the inability of meaningful asynchronous accesss from the host.
```
MKKnorr: Typo
```suggestion
  supported by the inability of meaningful asynchronous access from the host.
```

docs/reference/programming_model.rst
MathiasMagnus: Bulk here indeed refers to the massively parallel nature of the kernel launch, more specifically that the runtime doesn't track kernels independently, but identifies the entire launch as one entity. The language "bulk" is borrowed from ISO C++ [A Unified Executors Proposal for C++ | P0443R12](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p0443r12.html) which standardizes this behavior under the `bulk_executor` type.

docs/reference/programming_model.rst
MathiasMagnus: While I do agree with the sentiment of not conflating the physical realization from the logical structure, I don't agree with warps not being part of the API. Warp-collective functions explicitly operate at this granularity, and if the notion isn't introduced, it can't be referred to by those functions.

A block consisting of warps where the last has unnecessary threads is a valid concern though. We may mention that in text. Highlighting that capability in image may be too much in the opposite direction: encouraging using the API as such.

docs/reference/programming_model.rst
MathiasMagnus: The layering wishes to accentuate the multi-dimensional indexing of blocks/threads.

docs/reference/programming_model.rst
MathiasMagnus: In OpenCL it definitely is, and I'm not aware of it being any different in HIP. It may not be explicitly stated, but I'm not aware of runtimes being allowed at treating them otherwise: allowing a block to span multiple CUs (even when not using shared memory). The maximum block size is limited by the program counters of a given CU, which also happen to share the shared memory. (IMHO OpenCL does this better by explicitly nailing this down.)

> Compute Unit
>
>    An OpenCL device has one or more compute units. A work-group executes on a single compute unit. A compute unit is composed of one or more processing elements and local memory. A compute unit may also include dedicated texture filter units that can be accessed by its processing elements.
>
> Work-group
>
>    A collection of related work-items that execute on a single compute unit. The work-items in the group execute the same kernel-instance and share local memory and work-group functions.

docs/reference/programming_model.rst
MathiasMagnus: Top-down and bottom-up explanation could both be done.

Again, if there were no explicit warp-collective primitives in the API, I'd also say they're implementation detail, but I think they're not.

docs/reference/programming_model.rst
MathiasMagnus: I think the first "but" is understandable. The second paragraph needs a rewrite/clarification.

docs/reference/programming_model.rst
MathiasMagnus: Needs fix.

docs/reference/programming_model.rst
MathiasMagnus: Fixing...

docs/reference/programming_model.rst
MathiasMagnus: Fixing...

docs/reference/programming_model.rst
MathiasMagnus: Should say "event completes".

docs/reference/programming_model.rst
MathiasMagnus: Separate out "generate" and "used" for the two parts of the sentence.

docs/understand/programming_model.rst
MathiasMagnus: The "matrix ALU" already implies what it's meant to accelerate, but I wouldn't want to corner myself too much with the definition. As is with any fixed-function hardware, people always try to find way to use them for what it's not meant for: [exhibit A](https://arxiv.org/abs/2209.11287).

docs/reference/programming_model.rst
MKKnorr: I'd still say that first and foremost (thread) blocks are a group of threads. When using warp-level functions, the user is required to take care of making sure that the right threads participate, so it is not as straight forward, especially as there is no built in "warpIdx".

And, even though we don't have to conform to that, the cuda programming guide doesn't even mention warps in its programming model chapter at all. It first explains it in the chapter about hardware implementation. Granted, those chapters can be interleaved somewhat, and I would also argue, that warps should be mentioned here, for completeness sake, but they don't belong in the thread hierarchy. There is no address space or anything related to warps, that makes them stand out. Anything you can do on a warp level should also be possible on a block level, just not as efficient/performant

docs/reference/programming_model.rst
MKKnorr: One could argue that warps are an implementation detail in the sense, that their size is not defined - an architecture with a warp size of one is imaginable as well

docs/reference/programming_model.rst
MKKnorr: Ok. Just realized it should also be "all the command`s`", just so you don't overlook it when fixing it

docs/reference/programming_model.rst
MKKnorr: Then I think 
```suggestion
    compiler itself and not intended for end-user code. Should you be
```
is still easier to understand

docs/understand/programming_model.rst
MKKnorr: Yes, I agree, but "algorithms" is too broad of a term for my taste. The linked paper even states

> any computation that can be expressed as MMA operations can leverage TCs

so "accelerating algorithms involving matrix-multiply accumulate operations" or "accelerating algorithms representable as matrix-multiply operations" is more precise, while not leaving the "unorthodox" use cases out

docs/reference/programming_model.rst
MathiasMagnus: Needs extra text or clearer depiction in image. (note to self: register file)

docs/reference/programming_model.rst
MathiasMagnus: Fixed.

docs/reference/programming_model.rst
MathiasMagnus: Fixed.

docs/reference/programming_model.rst
MathiasMagnus: Fixed.

docs/reference/programming_model.rst
MathiasMagnus: Fixed.

docs/reference/programming_model.rst
MathiasMagnus: Fixed.

docs/data/reference/programming_model/thread_hierarchy_coop.svg
gandryey, line 1: I think we should add the depth for Multi Grid picture. It's multiple grids, thus it should show the depth of gray rectangles called "Grid"

docs/data/understand/programming_model/cdna3_cu_dark.png
MKKnorr: Removed image

docs/data/understand/programming_model/rdna3_cu.drawio
MKKnorr: Removed file

docs/reference/programming_model.rst
MKKnorr: Added a sentence mentioning threads, at least. I'll open another issue to keep track of this - this image should be updated

docs/reference/programming_model.rst
MKKnorr: Changed the description - blocks need to run on a single CU. The need for shared memory is not good enough, then the whole grid could be seen as a block, as it shares the global memory, even if that would be slow. The possibility of synchronizing on a finer grained level is definitely more important, and that doesn't even necessarily require shared memory.

docs/reference/programming_model.rst
MKKnorr: Added the actual function

docs/reference/programming_model.rst
MKKnorr: Added the actual function

docs/reference/programming_model.rst
MKKnorr: Fixed

docs/reference/programming_model.rst
MKKnorr: Fixed

docs/reference/programming_model.rst
MKKnorr: I'll add an issue for it

docs/reference/programming_model.rst
MKKnorr: Slightly changed/reworded

docs/reference/programming_model.rst
MKKnorr: Couldn't fix it either, not even sure what the original target of this was - removed the reference

docs/understand/programming_model.rst
MKKnorr: Reworded the section, also added scalar ALU, as it is needed for the description of the CUs

docs/understand/programming_model.rst
MKKnorr: Reworded

docs/understand/programming_model.rst
MKKnorr: Fixed

docs/understand/programming_model.rst
MKKnorr: Fixed

docs/understand/programming_model.rst
MKKnorr: Slightly reworded

docs/data/reference/programming_model/thread_hierarchy_coop.svg
MKKnorr: This is already portraied by having several Grids in the picture. The grids don't necessarily have to be of uniform size, so laying them on top of each other could be misleading

MKKnorr:
I took over this PR and fixed all the remarks I still had on the original PR, in a hopefully sensible manner, and will create new issues for the docs team for the rest

@jujiang-del please review and merge in Gerrit when appropriate

jujiang-del:
Merging to Gerrit
https://gerrit-git.amd.com/c/compute/ec/hip/+/1034056

jujiang-del:
Merging to Gerrit
https://gerrit-git.amd.com/c/compute/ec/hip/+/1034056

MKKnorr:
Closing, as this has been upstreamed via gerrit. Thanks!

PR #3449, opened 2024-04-14
Fix README dead-end links

FattiMei:
Some links in README.md gave a 404 status, I searched corresponding files in the repository and updated the links


neon60:
The links had been fixed, and the section already also removed from README file. Closing the PR.

PR #3455, opened 2024-04-24
Add HIP Performance Guidelines

matyas-streamhpc:


docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+developers optimize the performance of AMD GPUs. They cover established
+parallelization and optimization techniques, coding metaphors, and idioms that
+can greatly simplify programming for HIP-capable GPU architectures.
+
+By following four main cornerstones, we can exploit the performance 
+optimization potential of HIP.
+
+- parallel execution
+- memory usage optimization
+- optimization for maximum instruction
```
Melirius: This is about throughput effectively. Will be better
```suggestion
- optimization for maximum throughput
```

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+- minimizing memory thrashing
+
+In the following chapters, we will show you their benefits and how to use them
+effectively.
+
+.. _parallel execution:
+Parallel execution
+===============================================================================
+For optimal use, the application should reveal and efficiently allocate as much
```
Melirius:
```suggestion
For optimal use, the application should reveal and efficiently imply as much
```
?

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+-------------------------------------------------------------------------------
+The application should optimize parallel execution across the host and devices 
+using asynchronous calls and streams. Workloads should be assigned based on 
+efficiency: serial to the host, parallel to the devices.
+
+For parallel workloads, when threads need to synchronize to share data, if they
+belong to the same block, they should use `__syncthreads()` (see:
+:ref:`synchronization functions`) within the same kernel invocation. If they 
+belong to different blocks, they must use global memory with two separate 
+kernel invocations. The latter should be minimized as it adds overhead.
```
Melirius: Hmm, e.g. SYCL has a guaranty that lowest global index block that is still executing is progressing. It allows some limited use of synchronization for later blocks in the same invocation. However, I don't remember, if it is the case for HIP, and cannot google it fast. Should we ask Young about?

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+-------------------------------------------------------------------------------
+Multiprocessor-level optimization involves maximizing parallel execution within
+each multiprocessor on a device. Each multiprocessor can execute a number of 
+threads concurrently, and the total number of threads that can run in parallel
+is determined by the number of concurrent threads each multiprocessor can 
+handle.
+
+The key to multiprocessor-level optimization is to efficiently utilize the 
+various functional units within a multiprocessor. This can be achieved by 
+ensuring a sufficient number of resident warps, and at every instruction issue
```
Melirius:
```suggestion
ensuring a sufficient number of resident warps, as at every instruction issue 
```

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+resources of the GPU, from the individual multiprocessors to the device as a
+whole.
+
+.. _memory optimization:
+Memory optimization
+===============================================================================
+The first step in maximizing memory throughput is to minimize low-bandwidth 
+data transfers. This involves reducing data transfers between the host and the 
+device, as these have lower bandwidth than transfers between global memory and 
+the device.
```
Melirius: What is it "transfers between global memory and the device"? Global memory is a part of the device itself. Or you have in mind unified address space? Then it is not always true and should be rewritten more clearly anyway.

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+
+Additionally, data transfers between global memory and the device should be 
+minimized by maximizing the use of on-chip memory: shared memory and caches. 
+Shared memory acts as a user-managed cache, where the application explicitly
```
Melirius: Aha, from here it is clear that "device" = "CUs". That is slightly misleading, could we elaborate another term for it?

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+the device.
+
+Additionally, data transfers between global memory and the device should be 
+minimized by maximizing the use of on-chip memory: shared memory and caches. 
+Shared memory acts as a user-managed cache, where the application explicitly 
+allocates and accesses it. A common programming pattern is to stage data from 
+device memory into shared memory. This involves each thread of a block loading 
+data from device memory to shared memory, synchronizing with all other threads 
+of the block, processing the data in shared memory, synchronizing again if 
+necessary, and writing the results back to device memory.
```
Melirius:
```suggestion
necessary, and writing the results back to device global memory.
```

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+depending on the access pattern for each type of memory. Therefore, the next 
+step in maximizing memory throughput is to organize memory accesses as 
+optimally as possible. This is especially important for global memory accesses,
+as global memory bandwidth is low compared to available on-chip bandwidths and
+arithmetic instruction throughput. Thus, non-optimal global memory accesses
+generally have a high impact on performance.
+
+Data Transfer
+-------------------------------------------------------------------------------
+applications should aim to minimize data transfers between the host and the
```
Melirius:
```suggestion
Applications should aim to minimize data transfers between the host and the 
```

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+
+Batching small transfers into a single large transfer can improve performance 
+due to the overhead associated with each transfer. On systems with a front-side
+bus, using page-locked host memory can enhance data transfer performance.
+
+When using mapped page-locked memory, there is no need to allocate device 
+memory or explicitly copy data between device and host memory. Data transfers 
+occur implicitly each time the kernel accesses the mapped memory. For optimal
+performance, these memory accesses should be coalesced, similar to global
+memory accesses.
```
Melirius: There is no definition of "coalescing" here or before, but it is mentioned in such a way like it was introduced earlier. It should be accompanied with explanation or at least linked to explanation in another place.

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+
+Device Memory Access
+-------------------------------------------------------------------------------
+Memory access instructions may be repeated due to the spread of memory
+addresses across warp threads. The impact on throughput varies with memory type
+and is generally reduced when addresses are more scattered, especially in
+global memory.
+
+Device memory is accessed via 32-, 64-, or 128-byte transactions that must be
+naturally aligned. Maximizing memory throughput involves coalescing memory
```
Melirius: I think a short glossary in the beginning can be very valuable. For example, I'm not sure here, is "device memory" and "global memory" the same thing in these sentences, or it is different concepts. And again "coalescing" without explanation.

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+accesses of threads within a warp into minimal transactions, following optimal
+access patterns, using properly sized and aligned data types, and padding data
+when necessary.
+
+Global memory instructions support reading or writing data of specific sizes 
+(1, 2, 4, 8, or 16 bytes) that are naturally aligned. If the size and alignment
+requirements are not met, it leads to multiple instructions, reducing 
+performance. Therefore, using data types that meet these requirements, ensuring
+alignment for structures, and maintaining alignment for all values or arrays is
+crucial for correct results and optimal performance.
```
Melirius: Hmm, correct results should be obtained even on bad alignment, I suggest
```suggestion
crucial for optimal performance.
```

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+
+threads often access 2D arrays at an address calculated by
```
Melirius:
```suggestion
Threads often access 2D arrays at an address calculated as
```

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+(1, 2, 4, 8, or 16 bytes) that are naturally aligned. If the size and alignment
+requirements are not met, it leads to multiple instructions, reducing 
+performance. Therefore, using data types that meet these requirements, ensuring
+alignment for structures, and maintaining alignment for all values or arrays is
+crucial for correct results and optimal performance.
+
+threads often access 2D arrays at an address calculated by 
+`BaseAddress + xIndex + width * yIndex`. For efficient memory access, the array
+and thread block widths should be multiples of the warp size. If the array 
+width is not a multiple of the warp size, it is more efficient to allocate it
```
Melirius:
```suggestion
width is not a multiple of the warp size, it is usually more efficient to allocate it
```
The pb can be with large arrays.

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+serviced at the throughput of the constant cache for cache hits, or the 
+throughput of the device memory otherwise.
+
+Texture and surface memory are stored in device memory and cached in texture 
+cache. This setup optimizes 2D spatial locality, leading to better performance
+for threads reading close 2D addresses. Reading device memory through texture
+or surface fetching can be advantageous, offering higher bandwidth for local
+texture fetches or surface reads, offloading addressing calculations,
+allowing data broadcasting, and optional conversion of 8-bit and 16-bit integer
+input data to 32-bit floating-point values.
```
Melirius:
```suggestion
input data to 32-bit floating-point values on-the-fly.
```

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+
+.. _instruction optimization:
+Optimization for maximum instruction
```
Melirius:
```suggestion
Optimization for maximum instruction throughput
```

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+===============================================================================
+To maximize instruction throughput:
+
+- minimize low throughput arithmetic instructions
+- minimize divergent warps inflicted by control flow instructions
+- minimize the number of instruction as possible
```
Melirius:
```suggestion
- minimize the number of instruction as possible
- maximize instruction parallelism
```

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+
+Arithmetic instructions
+-------------------------------------------------------------------------------
+The type and complexity of arithmetic operations can significantly impact the 
+performance of your application. We are highlighting some hints how to maximize
+it.
+
+Using efficient operations: Some arithmetic operations are more costly than 
+others. For example, multiplication is typically faster than division, and 
+integer operations are usually faster than floating-point operations.
```
Melirius:
```suggestion
integer operations are usually faster than floating-point operations,
especially with double-precision.
```

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+
+Minimizing low-throughput instructions: This might involve trading precision 
+for speed when it does not affect the final result. For instance, consider 
+using single-precision arithmetic instead of double-precision.
+
+Leverage intrinsic functions: Intrinsic functions are pre-defined functions 
+available in HIP that can often be executed faster than equivalent arithmetic 
+operations. They can help optimize performance by replacing more complex
```
Melirius:
```suggestion
operations (subject to some input or accuracy restrictions). They can help optimize performance by replacing more complex 
```

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+warp follow different execution paths. This can happen due to conditional 
+statements that lead to different arithmetic operations being performed by 
+different threads. Divergent warps can significantly reduce instruction 
+throughput, so try to structure your code to minimize divergence.
+
+Optimizing memory access: The efficiency of memory access can impact the speed
+of arithmetic operations. Coalesced memory access, where threads in a warp 
+access consecutive memory locations, can improve memory throughput and thus 
+the speed of arithmetic operations.
+
```
Melirius:
```suggestion

Maximizing instruction parallelism: Some GPU architectures could issue parallel independent instructions simultaneously, for example integer and floating point, or two operations with independent inputs and outputs. Mostly this is a work for compiler, but expressing parallelism in the code explicitly can improve instructions throughput.
```

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+For example, when the control condition depends on (`threadIdx` / `warpSize`), 
+no warp diverges. The compiler may optimize loops or short if or switch blocks 
+using branch predication, preventing warp divergence. With branch predication, 
+instructions associated with a false predicate are scheduled but not executed, 
+avoiding unnecessary operations.
+
+Synchronization
+-------------------------------------------------------------------------------
+ Synchronization ensures that all threads within a block have completed their 
+ computations before moving forward, which is critical when threads are
```
Melirius:
```suggestion
 computations and memory accesses before moving forward, which is critical when threads are 
```

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+ computations before moving forward, which is critical when threads are 
+ dependent on the results of other threads. However, synchronization can also 
+ lead to performance overhead, as it requires threads to wait, potentially
+ leading to idle GPU resources.
+
+`__syncthreads()` is used to synchronize all threads in a block, ensuring that
+all threads have reached the same point in the code and that shared memory is
+visible to all threads after the point of synchronization.
+
+ As for alternative ways to synchronize is using streams. Different streams
```
Melirius:
```suggestion
 An alternative way to synchronize is using streams. Different streams 
```

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+ leading to idle GPU resources.
+
+`__syncthreads()` is used to synchronize all threads in a block, ensuring that
+all threads have reached the same point in the code and that shared memory is
+visible to all threads after the point of synchronization.
+
+ As for alternative ways to synchronize is using streams. Different streams 
+ can execute commands out of order with respect to one another or concurrently.
+ This allows for more fine-grained control over the execution order of 
+ commands, which can be beneficial in certain scenarios.
```
Melirius: Hmm, using streams for intra-block synchronization is a definite overkill. I suggest to extend this paragraph to explain, what level of synchronization is provided by streams, and make a link to their description.

docs/reference/performance.rst
```diff
@@ -0,0 +1,277 @@
...
+  application and deallocate only when the application no longer needs it. 
+  Minimize the number of `hipMalloc` and `hipFree` calls in your application,
+  particularly in areas critical to performance.
+- if an application is unable to allocate sufficient device memory, consider 
+  resorting to other memory types such as `hipHostMalloc` or 
+  `hipMallocManaged`. While these may not offer the same performance, they can 
+  allow the application to continue running.
+- For supported platforms, `hipMallocManaged` allows for oversubscription. With
+  the right memory advise policies, it can maintain most, if not all, of the 
+  performance of `hipMalloc`. hipMallocManaged does not require an allocation
```
Melirius:
```suggestion
  performance of `hipMalloc`. `hipMallocManaged` does not require an allocation 
```

docs/reference/performance.rst
matyas-streamhpc: fixed

docs/reference/performance.rst
matyas-streamhpc: Thanks! Fixed.

docs/reference/performance.rst
matyas-streamhpc: Thanks! Fixed.

docs/reference/performance.rst
matyas-streamhpc: Fixed.

docs/reference/performance.rst
matyas-streamhpc: FIxed

docs/reference/performance.rst
matyas-streamhpc: Added

docs/reference/performance.rst
matyas-streamhpc: Fixed.

docs/reference/performance.rst
matyas-streamhpc: Thanks! Updated.

docs/reference/performance.rst
matyas-streamhpc: Added

docs/reference/performance.rst
matyas-streamhpc: Updated

docs/reference/performance.rst
matyas-streamhpc: Added

docs/reference/performance.rst
matyas-streamhpc: Added

docs/reference/performance.rst
matyas-streamhpc: Added

docs/reference/performance.rst
matyas-streamhpc: Fixed.

docs/reference/performance.rst
matyas-streamhpc: Fixed.

docs/reference/performance.rst
matyas-streamhpc: Fixed.

docs/index.md
```diff
@@ -17,6 +17,7 @@ portable applications for AMD and NVIDIA GPUs from single source code.
 
 :::{grid-item-card} Reference
 
+* {doc}`/reference/performance_guidelines`
```
MKKnorr: I would argue this document would rather fit in the "understand" section, than "reference"

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+  :description: This chapter describes a set of best practices designed to help
+   developers optimize the performance of HIP-capable GPU architectures.
+  :keywords: AMD, ROCm, HIP, CUDA, performance, guidelines
+
+*******************************************************************************
+Performance Guidelines
+*******************************************************************************
+
+The AMD HIP Performance Guidelines are a set of best practices designed to help
+developers optimize the performance of AMD GPUs. They cover established
```
MKKnorr:
```suggestion
developers optimize the performance of their code for AMD GPUs. They cover established
```
These guidelines don't magically make GPUs perform better, but the code that is executed on them :)

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+parallelization and optimization techniques, coding metaphors, and idioms that
```
MKKnorr: What is a coding metaphor? Are they really covered in here?

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+can greatly simplify programming for HIP-capable GPU architectures.
```
MKKnorr: I wouldn't say these techniques are for simplifying code, but for improving performance. At least that should be the goal of this document

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+
+By following four main cornerstones, we can exploit the performance
+optimization potential of HIP.
+
+- parallel execution
+- memory usage optimization
```
MKKnorr:
```suggestion
- memory bandwidth (usage) optimization
```
Reducing memory usage/footprint itself does not necessarily optimize the performance, but optimizing the bandwidth (usage) can

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+- optimization for maximum throughput
```
MKKnorr:
```suggestion
- optimizing for maximum instruction throughput
```

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+- minimizing memory thrashing
+
+In the following chapters, we will show you their benefits and how to use them
+effectively.
+
+.. _parallel execution:
+Parallel execution
+===============================================================================
+For optimal use, the application should reveal and efficiently imply as much
+parallelism as possible to keep all system components active.
```
MKKnorr:
```suggestion
For optimal use, the application should reveal and efficiently employ/provide as much
parallelism as possible to keep all system components busy.
```
"imply" doesn't really make sense here. Keeping the components "busy" also is more direct than "active"

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+
+Application level
+-------------------------------------------------------------------------------
+The application should optimize parallel execution across the host and devices
+using asynchronous calls and streams. Workloads should be assigned based on
+efficiency: serial to the host, parallel to the devices.
+
+For parallel workloads, when threads need to synchronize to share data, if they
+belong to the same block, they should use ``__syncthreads()`` (see:
+:ref:`synchronization functions`) within the same kernel invocation. If they
```
MKKnorr: This is not a recommendation ("should"), but a must. Also, this does not really belong to the performance guidelines, this is about correctness.

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+belong to different blocks, they must use global memory with two separate
+kernel invocations. The latter should be minimized as it adds overhead.
+
+Device level
+-------------------------------------------------------------------------------
+Device-level optimization primarily involves maximizing parallel execution
```
MKKnorr:
```suggestion
Device level optimization primarily involves maximizing parallel execution
```

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+across the multiprocessors of the device. This can be achieved by executing
+multiple kernels concurrently on a device. The management of these kernels is
+facilitated by streams, which allow for the overlapping of computation and data
+transfers, enhancing performance. The aim is to keep all multiprocessors busy
+by executing enough kernels concurrently. However, launching too many kernels
+can lead to resource contention, so a balance must be found for optimal
+performance. This approach helps in achieving maximum utilization of the
+resources of the device.
```
MKKnorr: This is not only achievable by running multiple kernels. The first pitfall for using GPUs and not getting the proper performance, is not having a big enough parallel workload in general. One kernel with a big enough grid can also fully occupy a gpu.

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+
+Multiprocessor level
+-------------------------------------------------------------------------------
+Multiprocessor-level optimization involves maximizing parallel execution within
```
MKKnorr:
```suggestion
Multiprocessor level
-------------------------------------------------------------------------------
Multiprocessor level optimization involves maximizing parallel execution within
```
Either use the dash also in the heading or in none of the two, same goes for "device-level" in the previous section

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+each multiprocessor on a device. Each multiprocessor can execute a number of
+threads concurrently, and the total number of threads that can run in parallel
+is determined by the number of concurrent threads each multiprocessor can
+handle.
```
MKKnorr: This is almost a tautology. Did you mean
```suggestion
each multiprocessor on a device. Each multiprocessor can execute a number of
threads concurrently, and the total number of threads that can run in parallel on a gpu
is determined by the number of concurrent threads each multiprocessor can
handle.
```
?
If yes, than that rather belongs to the device level optimization. If you mean occupancy, then you should rewrite this a bit

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+
+The key to multiprocessor-level optimization is to efficiently utilize the
+various functional units within a multiprocessor. This can be achieved by
+ensuring a sufficient number of resident warps, as at every instruction issue
+time, a warp scheduler selects an instruction that is ready to execute. This
```
MKKnorr:
```suggestion
various functional units within a multiprocessor. This can be achieved by
ensuring a sufficient number of resident warps, so that every clock cycle an instruction from a warp is ready for execution. This
```
Also: full occupancy is not a guarantee for peak performance, so "This can be achieved by" is not enitrely accurate

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+instruction can be another independent instruction of the same warp, exploiting
+:ref:`instruction optimization<instruction-level parallelism>`, or more
```
MKKnorr:
```suggestion
:ref:`instruction level parallelism<instruction-level parallelism>`, or more
```
this link is not working right now

Also: This is not necessarily called instruction level parallelism, that depends on the hardware implementation. But I think this is nitpicking, as it goes into too much detail

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+commonly an instruction of another warp, exploiting thread-level parallelism.
+
+In comparison, device-level optimization focuses on the device as a whole,
+aiming to keep all multiprocessors busy by executing enough kernels
```
MKKnorr: Again, this is not only achievable by executing enough kernels in parallel

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+resources of the GPU, from the individual multiprocessors to the device as a
+whole.
+
+.. _memory optimization:
+Memory optimization
+===============================================================================
+The first step in maximizing memory throughput is to minimize low-bandwidth
+data transfers. This involves reducing data transfers between the host and the
+device, as these have lower bandwidth than transfers between global memory and
+the device.
```
MKKnorr: I agree with @Melirius, this is written in a confusing way. I'd just ommit the second sentence and integrate it into the first.
```suggestion
data transfers between the host and the device.
```

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+
+In comparison, device-level optimization focuses on the device as a whole,
+aiming to keep all multiprocessors busy by executing enough kernels
+concurrently. Both levels of optimization are crucial for achieving maximum
+performance. They work together to ensure efficient utilization of the
+resources of the GPU, from the individual multiprocessors to the device as a
+whole.
+
+.. _memory optimization:
+Memory optimization
```
MKKnorr:
```suggestion
Memory throughput/bandwidth optimization
```
choose one. Memory optimization itself is too broad of a term when it comes to performance

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+===============================================================================
+The first step in maximizing memory throughput is to minimize low-bandwidth
+data transfers. This involves reducing data transfers between the host and the
+device, as these have lower bandwidth than transfers between global memory and
+the device.
+
+Additionally, data transfers between global memory and the device should be
```
MKKnorr: Same as @Melirius - "device" is not the proper term to use here.
Maybe rephrase to
```suggestion
Additionally, loads from and stores to global memory should be
```

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+minimized by maximizing the use of on-chip memory: shared memory and caches.
+Shared memory acts as a user-managed cache, where the application explicitly
+allocates and accesses it. A common programming pattern is to stage data from
+device memory into shared memory. This involves each thread of a block loading
+data from device memory to shared memory, synchronizing with all other threads
+of the block, processing the data in shared memory, synchronizing again if
```
MKKnorr:
```suggestion
of the block, processing the data which is stored in shared memory, synchronizing again if
```
Otherwise this sounds like in-memory computing, which it defintely is not

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+necessary, and writing the results back to device global memory.
+
+For some applications, a traditional hardware-managed cache is more appropriate
+to exploit data locality. On devices of certain compute capabilities, the same
+on-chip memory is used for both L1 and shared memory, and the amount dedicated
+to each is configurable for each kernel call.
```
MKKnorr:
```suggestion
to exploit data locality.
```
This is not true for AMD GPUs, and "compute capability" is not a valid term for AMD GPUs.

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+
+Finally, the throughput of memory accesses by a kernel can vary significantly
+depending on the access pattern for each type of memory. Therefore, the next
```
MKKnorr:
```suggestion
Finally, the throughput of memory accesses by a kernel can vary significantly
depending on the access pattern. Therefore, the next
```

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+-------------------------------------------------------------------------------
+Applications should aim to minimize data transfers between the host and the
+device. This can be achieved by moving more computations from the host to the
+device, even if it means running kernels that do not fully utilize the
+parallelism for device. Intermediate data structures can be created, used,
+and discarded in device memory without being mapped or copied to host memory.
+
+Batching small transfers into a single large transfer can improve performance
+due to the overhead associated with each transfer. On systems with a front-side
+bus, using page-locked host memory can enhance data transfer performance.
```
MKKnorr:
```suggestion
due to the overhead associated with each transfer.
```
This sentence really dates the documentation. I don't think there's any system out there with an FSB that is supported by ROCm (e.g. providing PCIe 3.0)

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+allowing data broadcasting, and optional conversion of 8-bit and 16-bit integer
+input data to 32-bit floating-point values on-the-fly.
+
+.. _instruction optimization:
+Optimization for maximum instruction throughput
+===============================================================================
+To maximize instruction throughput:
+
+- minimize low throughput arithmetic instructions
+- minimize divergent warps inflicted by control flow instructions
```
MKKnorr:
```suggestion
- minimize divergence within warps inflicted by control flow instructions
```
just to emphasize, that different warps may take different execution paths, but the important part is all threads within a warp taking the same path

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+- minimize the number of instruction as possible
```
MKKnorr:
```suggestion
- minimize the amount of instructions as much as possible
```
or even:
```suggestion
```
Also: this is kind of a weird performance optimization. Obviously one shouldn't do unnecessary work, but reducing the amount of instructions in general seems like a weird suggestion. This sounds more like an optimization on an algorithmical level, but not specific to GPUs

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+Arithmetic instructions
+-------------------------------------------------------------------------------
+The type and complexity of arithmetic operations can significantly impact the
+performance of your application. We are highlighting some hints how to maximize
+it.
+
+Using efficient operations: Some arithmetic operations are more costly than
+others. For example, multiplication is typically faster than division, and
+integer operations are usually faster than floating-point operations,
+especially with double-precision.
```
MKKnorr: "faster" is kind of a broad term here. When it comes to throughput, floating-point and integer operations are mostly on par, and some architectures even have higher single-precision throughput than integer throughput. Especially when it comes to 64 bit integer operations, as these have to be emulated. I'm not sure about latency though right now.

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+
+Minimizing low-throughput instructions: This might involve trading precision
+for speed when it does not affect the final result. For instance, consider
+using single-precision arithmetic instead of double-precision.
```
MKKnorr: I'm torn on these suggestions in the performance guidelines. The specific case ("when it does not affect the final result") is kind of a weird one. In general moving from double to single precision changes the outcome of the result.

This is an optimization that does not necessarily get you closer to peak performance of the device, but improves performance by doing less/different work

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+Leverage intrinsic functions: Intrinsic functions are pre-defined functions
+available in HIP that can often be executed faster than equivalent arithmetic
+operations (subject to some input or accuracy restrictions). They can help
+optimize performance by replacing more complex arithmetic operations.
+
+Avoiding divergent warps: Divergent warps occur when threads within the same
+warp follow different execution paths. This can happen due to conditional
+statements that lead to different arithmetic operations being performed by
+different threads. Divergent warps can significantly reduce instruction
+throughput, so try to structure your code to minimize divergence.
```
MKKnorr: I would argue that divergent warps don't reduce the instruction throughput, but the utilization of the ALUs/SIMDs. In theory you can still issue as many warp-level instructions as before - achieving the same throughput - but in the worst case only one thread actually executes the instruction

This paragraph probably fits better into the "Control flow instructions" section

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+
+Optimizing memory access: The efficiency of memory access can impact the speed
+of arithmetic operations. Coalesced memory access, where threads in a warp
+access consecutive memory locations, can improve memory throughput and thus
+the speed of arithmetic operations.
```
MKKnorr: Pedantic: arithmetic operations can't be "sped up", the time they can't be scheduled for execution however can depend on memory accesses. I would add a reference to the memory optimizations here

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+
+Maximizing instruction parallelism: Some GPU architectures could issue parallel
+independent instructions simultaneously, for example integer and floating
```
MKKnorr:
```suggestion
Maximizing instruction parallelism: Some GPU architectures could issue 
independent instructions simultaneously, for example integer and floating
```

Also: some AMD architectures have support for "packed math", adding the ability to execute e.g. two 16-bit floating-point instructions at once
https://gpuopen.com/learn/rdna-performance-guide/#bit-math

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+point, or two operations with independent inputs and outputs. Mostly this is a
+work for compiler, but expressing parallelism in the code explicitly can
+improve instructions throughput.
```
MKKnorr:
```suggestion
point, or two operations with independent inputs and outputs. This is mostly
done by the compiler, but expressing parallelism in the code explicitly can
improve instructions throughput.
```

docs/reference/performance_guidelines.rst
```diff
@@ -0,0 +1,287 @@
...
+(``threadIdx`` / ``warpSize``), no warp diverges. The compiler may optimize
+loops or short if or switch blocks using branch predication, preventing warp
+divergence. With branch predication, instructions associated with a false
+predicate are scheduled but not executed, avoiding unnecessary operations.
+
+Synchronization
+-------------------------------------------------------------------------------
+Synchronization ensures that all threads within a block have completed their
+computations and memory accesses before moving forward, which is critical when
+threads are dependent on the results of other threads. However,
```
MKKnorr:
```suggestion
computations and memory accesses before moving forward, which is critical when
threads depend on the results of other threads. However,
```

MKKnorr (changes requested):
Most of these sections are ***very*** close to the performance guidelines of the cuda programming guide, sometimes almost quoting it directly. I don't think that's a good practice, especially as some parts don't apply to AMDs GPUs at all, and on top of that the cuda programmign guide does not have a permissive license from what I can tell

A better place for inspiration might be gpuopen, that already has some performance guides for e.g. rdna https://gpuopen.com/learn/rdna-performance-guide/

docs/reference/performance_guidelines.rst
Melirius: Do not forget, HIP is also positioned as a tool for NVIDIA GPUs - "one-code-fits-all".

docs/reference/performance_guidelines.rst
Melirius: HIP compiler usually does not emit commands for fast-skip. Or now already does?

docs/reference/performance_guidelines.rst
Melirius: Heh, you're right. That is the timings for GCN architecture: https://github.com/CLRX/CLRX-mirror/blob/master/doc/GcnTimings.md Float is even faster than integer.

Melirius:
BTW, it should be mentioned somewhere that to fully utilize all SIMD lines/possible threads in the block x-block size should be a multiple of warp size.

docs/reference/performance_guidelines.rst
MKKnorr: That is true, but in that case it would be better to point to the cuda documentation, and not copy it here. Apart from being questionable when it comes to the license, it also increases the burden on keeping the documentation up to date.

docs/reference/performance_guidelines.rst
MKKnorr: Good question. Not sure about that. But it should still be emphasized, that divergence is only a problem when the threads within a warp diverge. If different warps take different code paths no damage is done (from the instruction-throughput point of view)

MKKnorr:
> BTW, it should be mentioned somewhere that to fully utilize all SIMD lines/possible threads in the block x-block size should be a multiple of warp size.

When being pedantic: the block size (`x*y*z`) has to be a multiple of the warp size for full utilization

docs/index.md
matyas-streamhpc: It is moved to `How-to`

docs/reference/performance_guidelines.rst
matyas-streamhpc: Fixed.

docs/reference/performance_guidelines.rst
matyas-streamhpc: Updated.

docs/reference/performance_guidelines.rst
matyas-streamhpc: Updated.

docs/reference/performance_guidelines.rst
matyas-streamhpc: true. fixed.

docs/reference/performance_guidelines.rst
matyas-streamhpc: Thanks! Updated.

docs/reference/performance_guidelines.rst
matyas-streamhpc: Updated

docs/reference/performance_guidelines.rst
matyas-streamhpc: Fixed

docs/reference/performance_guidelines.rst
matyas-streamhpc: Perhaps no. I removed from the document.

docs/reference/performance_guidelines.rst
matyas-streamhpc: Fixed.

docs/reference/performance_guidelines.rst
matyas-streamhpc: Yes, I was thinking about it. Thanks! Updated.

docs/reference/performance_guidelines.rst
matyas-streamhpc: Fixed.

docs/reference/performance_guidelines.rst
matyas-streamhpc: Fixed

docs/reference/performance_guidelines.rst
matyas-streamhpc: Removed. I think it is not necessary.

docs/reference/performance_guidelines.rst
matyas-streamhpc: Updated

docs/reference/performance_guidelines.rst
matyas-streamhpc: Fixed

docs/reference/performance_guidelines.rst
matyas-streamhpc: Updated

docs/reference/performance_guidelines.rst
matyas-streamhpc: removed

docs/reference/performance.rst
matyas-streamhpc: updated

matyas-streamhpc:
> Most of these sections are _**very**_ close to the performance guidelines of the cuda programming guide, sometimes almost quoting it directly. I don't think that's a good practice, especially as some parts don't apply to AMDs GPUs at all, and on top of that the cuda programmign guide does not have a permissive license from what I can tell
> 
> A better place for inspiration might be gpuopen, that already has some performance guides for e.g. rdna https://gpuopen.com/learn/rdna-performance-guide/

I am not sure that is best strategy either, but the concept was accepted as a first version. It is not quoting directly the mentioned document, but there is overlap in the content. Personally, I would have appreciated every pieces of recommendation, both in format and in content.

Nonetheless, we always have the opportunity to improve it and make the documentation better for the satisfaction of the developers.

docs/reference/performance.rst
matyas-streamhpc: As I know, no. But I do not find it either. It would be a good idea.

docs/reference/performance_guidelines.rst
matyas-streamhpc: Updated

docs/reference/performance_guidelines.rst
matyas-streamhpc: Removed (see Ivan's comment)

docs/reference/performance_guidelines.rst
matyas-streamhpc: Moved

docs/reference/performance.rst
matyas-streamhpc: Introduction is added.

docs/reference/performance.rst
matyas-streamhpc: Updated

docs/reference/performance.rst
matyas-streamhpc: Updated

docs/reference/performance.rst
matyas-streamhpc: Reference added to coalescing.

docs/reference/performance_guidelines.rst
matyas-streamhpc: Added

neon60:
Most of this PR changes has been merged in, while the leftover is here: #3483

PR #3474, opened 2024-05-06
Changes from docs/6.1.1

neon60:
- Update documentation requirements
- Add HSA functions from ROCR Runtime
- Add links to module api and gpu_arch samples
- Move programming_model to understand folder
- Add hardware implementation

PR #3480, opened 2024-05-10
Update math functions

neon60:
This PR focusing on math functions update and those changes are in kernel_language.rst.

I based my changes on PR: #3474 This will be merged in soon at Gerrit side.

### DIFF
--- a/docs/reference/kernel_language.rst
+++ b/docs/reference/kernel_language.rst
@@ -327,441 +327,441 @@ Following is the list of supported single precision mathematical functions.
       - **Supported on Host**
       - **Supported on Device**
 
-    * - | float acosf ( float  x ) 
+    * - | ``float acosf(float  x)``
         | Calculate the arc cosine of the input argument.  
       - ✓
       - ✓
 
-    * - | float acoshf ( float  x ) 
+    * - | ``float acoshf(float  x)``
         | Calculate the nonnegative arc hyperbolic cosine of the input argument.
       - ✓
       - ✓
 
-    * - | float asinf ( float  x )
+    * - | ``float asinf(float  x)``
         | Calculate the arc sine of the input argument
       - ✓
       - ✓
 
-    * - | float asinhf ( float  x )
+    * - | ``float asinhf(float  x)``
         | Calculate the arc hyperbolic sine of the input argument.
       - ✓
       - ✓
 
-    * - | float atan2f ( float  y, float  x ) 
+    * - | ``float atan2f(float  y, float  x)``
         | Calculate the arc tangent of the ratio of first and second input arguments.
       - ✓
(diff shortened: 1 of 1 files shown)


randyh62 (dismissed):
Seems like some of the images in common between understand/hardware_implementation and understand/programming_model could be consolidated to use a single reference? It seems like this is being done in the kernel_language.rst file.

neon60:
> Seems like some of the images in common between understand/hardware_implementation and understand/programming_model could be consolidated to use a single reference? It seems like this is being done in the kernel_language.rst file.

The develop branch is slightly behind, this PR changes are focusing on kernel_language.rst.

Because of the data structure contribution rule. We have file duplication, or we do not state how to handle those.

docs/reference/kernel_language.rst
```diff
@@ -327,441 +327,441 @@ Following is the list of supported single precision mathematical functions.
...
+    * - | ``bool signbit(float  a)``
         | Return the sign bit of the input.
       - ✓
       - ✓
 
-    * - | void sincosf ( float  x, float* sptr, float* cptr )
+    * - | ``void sincosf(float  x, float* sptr, float* cptr)``
         | Calculate the sine and cosine of the first input argument.
       - ✓
       - ✗
```
matyas-streamhpc: Can you check whether it is supported now?

It seems from this page that it is supported now, but I did not check it.
https://cgmb-hip.readthedocs.io/en/latest/autotoc_md76.html

Thanks!

docs/reference/kernel_language.rst
```diff
@@ -1372,255 +1462,255 @@ HIP supports the following atomic operations.
...
+    * - ``double atomicAdd_system(double* address, double val)``
       - ✓
       - ✓
 
-    * - float unsafeAtomicAdd(float* address, float val)
+    * - ``float unsafeAtomicAdd(float* address, float val)``
       - ✓
       - ✗
 
-    * - float safeAtomicAdd(float* address, float val)
```
matyas-streamhpc, line 1419: missing apostrophes:
` * - ``float safeAtomicAdd(float* address, float val)`` `

docs/reference/kernel_language.rst
neon60: Supported.

neon60: I have to check this, thanks for the heads up.

docs/reference/kernel_language.rst
neon60: Fixed. Thanks

docs/reference/kernel_language.rst
neon60: Fixed.

PR #3487, opened 2024-05-17
MathiasMagnus SAXPY tutorial

neon60:
The SAXPY tutorial.

### DIFF
--- a/docs/tutorial/saxpy.rst
+++ b/docs/tutorial/saxpy.rst
@@ -1,0 +1,751 @@
+.. meta::
+  :description: The SAXPY tutorial on HIP
+  :keywords: AMD, ROCm, HIP, SAXPY, tutorial
+
+*******************************************************************************
+Tutorial: SAXPY - Hello, HIP
+*******************************************************************************
+
+This tutorial explains the basic concepts of the single-source
+Heterogeneous-computing Interface for Portability (HIP) programming model and
+the essential tooling around it. It also reviews some commonalities of
+heterogenous APIs in general. This topic assumes basic familiarity with the
+C/C++ compilation model and language.
+
+Prerequisites
+=============
+
+To follow this tutorial, you'll need installed drivers and a HIP compiler
+toolchain to compile your code. Because HIP for ROCm supports compiling and
+running on Linux and Windows with AMD and NVIDIA GPUs, the combination of
+install instructions is more than worth covering as part of this tutorial. For
+more information about installing HIP development packages, see
+:doc:`/install/install`.
+
+.. _hip-tutorial-saxpy-heterogeneous-programming:
+
+Heterogeneous programming
+=========================
+
+*Heterogeneous programming* and *offloading APIs* are often mentioned together. Heterogeneous programming deals with devices of varying capabilities simultaneously. Offloading focuses on the "remote" and asynchronous aspects of computation. HIP encompasses both. It exposes GPGPU (general-purpose GPU) programming much like ordinary host-side CPU programming and lets you move data across various devices.
+
+When programming in HIP (and other heterogenous APIs for that matter), remember that target devices are built for a specific purpose. They are designed with different tradeoffs than traditional CPUs and therefore have very different performance characteristics. Even subtle changes in code might adversely affect execution time.
+
+Your first lines of HIP code
+============================
+
+First, let's do the "Hello, World!" of GPGPU: SAXPY. Single-precision A times X Plus Y (*SAXPY*) is a mathematical acronym; a vector equation :math:`a\cdot x+y=z` where :math:`a\in\mathbb{R}` is a scalar and :math:`x,y,z\in\mathbb{V}` are vector quantities of some large dimensionality. This vector space is defined over the set of reals. Practically speaking, you can compute this using a single ``for`` loop over three arrays.
+
+.. code-block:: C++
+
+    for (int i = 0 ; i < N ; ++i)
+        z[i] = a * x[i] + y[i];
+
+In linear algebra libraries, such as BLAS (Basic Linear Algebra Subsystem) this operation is defined as AXPY "A times X Plus Y". The "S" comes from *single-precision*, meaning that array element is ``float`` -s (IEEE 754 binary32 representation).
+
+To quickly get started, use the set of `HIP samples from GitHub <https://github.com/amd/rocm-examples/>`_. With Git configured on your machine, open a command-line and navigate to your desired working directory, then run:
+
+.. code-block:: shell
+
+  git clone https://github.com/amd/rocm-examples.git
+
+A simple implementation of SAXPY resides in the ``HIP-Basic/saxpy/main.hip`` file in this repository. The HIP code here mostly deals with where data has to be and when, and how devices transform this data. The first HIP calls deal with allocating device-side memory and copying data from host-side memory to device side in a C runtime-like fashion.
+
+.. code-block:: C++
+
+  // Allocate and copy vectors to device memory.
+  float* d_x{};
+  float* d_y{};
+  HIP_CHECK(hipMalloc(&d_x, size_bytes));
+  HIP_CHECK(hipMalloc(&d_y, size_bytes));
+  HIP_CHECK(hipMemcpy(d_x, x.data(), size_bytes, hipMemcpyHostToDevice));
+  HIP_CHECK(hipMemcpy(d_y, y.data(), size_bytes, hipMemcpyHostToDevice));
+
+``HIP_CHECK`` is a custom macro borrowed from the examples utilities which checks the error code returned by API functions for errors and reports them to the console. It is not essential to the API, but it is a good practice to check the error codes of the HIP APIs in case you pass on incorrect values to the API, or the API might be out of resources.
+
+The code selects the device to allocate to and to copy to. Commands are issued to the HIP runtime per thread, and every thread has a device set as the target of commands. The default device is ``0``, which is equivalent to calling ``hipSetDevice(0)``.
+
+Launch the calculation on the device after the input data has been prepared.
+
+.. code-block:: C++
+
+  __global__ void saxpy_kernel(const float a, const float* d_x, float* d_y, const unsigned int size)
+  {
+      // ...
+  }
+
+  int main()
+  {
+      // ...
+
+      // Launch the kernel on the default stream.
+      saxpy_kernel<<<dim3(grid_size), dim3(block_size), 0, hipStreamDefault>>>(a, d_x, d_y, size);
+  }
+
+Analyze at the signature of the offloaded function:
+
+- ``__global__`` instructs the compiler to generate code for this function as an
+  entrypoint to a device program, such that it can be launched from the host.
+- The function does not return anything, because there is no trivial way to
+  construct a return channel of a parallel invocation. Device-side entrypoints
+  may not return a value, their results should be communicated using output
+  parameters.
+- Device-side functions are typically called compute kernels, or just kernels
+  for short. This is to distinguish them from non-graphics-related graphics
+  shaders, or just shaders for short.
+- Arguments are taken by value and all arguments shall be
+  `TriviallyCopyable <https://en.cppreference.com/w/cpp/named_req/TriviallyCopyable>`_,
+  meaning they should be `memcpy`-friendly. _(Imagine if they had custom copy
+  constructors. Where would that logic execute? On the host? On the device?)_
+  Pointer arguments are pointers to device memory, one typically backed by
+  VRAM.
+- We said that we'll be computing :math:`a\cdot x+y=z`, however we only pass
+  two pointers to the function. We'll be canonically reusing one of the inputs
+  as outputs.
+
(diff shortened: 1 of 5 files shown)


docs/tutorials/reduction.rst
randyh62, line 1: in C++ it's been signidfied by...

Spelling needs to be corrected, though I would change it to "in C++ it is called..."

randyh62, line 1: doesn not

randyh62, line 1: One such is what ``std::reduce`` does, is to require the user-provided binary...

The approach taken by ``std::reduce`` is to require the user-provided binary...

randyh62, line 1: Implementing reductions on GPUs will require a basic...

requires a basic...

randyh62, line 1: through the {ref}`inherent_thread_model`, and

This is not the correct reference syntax for RST. It does not render correctly in the output.

randyh62, line 1: Synchronizing _all_ the threads running on a GPU at any given time, while
possible, is a costly and/or intricate operation.

randyh62, line 1: We'll be exploring

You'll be exploring...generally we should avoid the use of we.

randyh62, line 1: If not absolutely necessary, we'll map our parallel algorithm as such that Multi Processors / Blocks can make independent progress and need not sync often.

If synchronization is not absolutely necessary, map the parallel algorithm so that multi-processors and blocks can make independent progress and need not sync often.

I don't think Blocks needs to be capitalized in the text.

randyh62, line 1: One may reduce divergence by keeping dataflow

You can reduce divergence...

randyh62, line 1: the coming code segments

the following code segments...

randyh62, line 1: strictly* lockstep

Not sure what the * is needed for

randyh62, line 1: Like we've mentioned in the previous step,

As mentioned in the prior step,

randyh62 (commented):
provided comments

docs/tutorials/reduction.rst
neon60: Fixed

neon60: Fixed

neon60: Fixed

neon60: Fixed

neon60: Fixed.

neon60: Fixed.

neon60: Fixed.

neon60: Changed.

neon60: Fixed.

neon60: Fixed.

neon60:
``IMAGE OF FINAL ALGO`` is missing

docs/tutorials/reduction.rst
neon60: Fixed.

neon60:
> `IMAGE OF FINAL ALGO` is missing

Fixed.

peterjunpark:
Added language suggestions from @lpaoletti. Review the diff if needed [2b4a8be](https://github.com/ROCm/HIP/pull/3487/commits/2b4a8beb8de2fea01faebd7cf911a61fed9cc578)

neon60:
@lpaoletti Can I get an approval?

neon60:
The reduction tutorial PR: #3513

peterjunpark:
Should we remove the tab sync?

I don't like how the page jumps around when you select a tab due to the length of content changing all over the place at the same time.

neon60:
> Should we remove the tab sync?
> 
> I don't like how the page jumps around when you select a tab due to content lengths changing everywhere at the same time.

Probably we should. I am thinking what we can do with it.

randyh62:
> > Should we remove the tab sync?
> > I don't like how the page jumps around when you select a tab due to content lengths changing everywhere at the same time.
> 
> Probably we should. I am thinking what we can do with it.

I think we should keep it. I think a user is going to pick their flow in the first tab, and the page will update the content to display their flow. I don't think people will be flipping back and forth between flows except to see what differences there might be.

docs/tutorial/saxpy.rst
```diff
@@ -0,0 +1,786 @@
...
+install instructions is more than worth covering as part of this tutorial. For
+more information about installing HIP development packages, see
+:doc:`/install/install`.
+
+.. _hip-tutorial-saxpy-heterogeneous-programming
+
+Heterogeneous programming
+=========================
+
+*Heterogeneous programming* and *offloading APIs* are often said in the same
```
lpaoletti: Heterogeneous programming and offloading APIs are often mentioned together.

docs/tutorial/saxpy.rst
```diff
@@ -0,0 +1,786 @@
...
+breath. Heterogeneous programming deals with devices of varying capabilities
```
peterjunpark:
```suggestion
*Heterogeneous programming* and *offloading APIs* are often mentioned together. Heterogeneous programming deals with devices of varying capabilities
```

docs/tutorials/reduction.rst
neon60: Move the reduction to a separate PR.

docs/tutorial/saxpy.rst
```diff
@@ -0,0 +1,780 @@
+.. meta::
+  :description: The SAXPY tutorial on HIP
+  :keywords: AMD, ROCm, HIP, SAXPY, tutorial
+
+*******************************************************************************
+Tutorial: SAXPY - Hello, HIP
+*******************************************************************************
+
+This tutorial explains the basic concepts of the single-source
+Heterogeneous-computing Interface for Portability (HIP) programming model and
```
cjatin, line 10: Can we change it to HIP (Heterogeneous-computing Interface) programming model. I think most people will call it HIP instead of its full name.

docs/tutorial/saxpy.rst
```diff
@@ -0,0 +1,780 @@
...
+that target devices are built for a specific purpose. They are designed with
+different tradeoffs than traditional CPUs and therefore have very different
+performance characteristics. Even subtle changes in code might adversely affect
+execution time.
+
+Your first lines of HIP code
+============================
+
+First, let's do the "Hello, World!" of GPGPU: SAXPY. *SAXPY* is a mathematical
+acronym; a vector equation :math:`a\cdot x+y=z` where :math:`a\in\mathbb{R}` is
```
cjatin: I think you have mentioned SAXPY's full form below. Can we move it up here.

docs/tutorial/saxpy.rst
```diff
@@ -0,0 +1,780 @@
...
+  float* d_x{};
+  float* d_y{};
+  HIP_CHECK(hipMalloc(&d_x, size_bytes));
+  HIP_CHECK(hipMalloc(&d_y, size_bytes));
+  HIP_CHECK(hipMemcpy(d_x, x.data(), size_bytes, hipMemcpyHostToDevice));
+  HIP_CHECK(hipMemcpy(d_y, y.data(), size_bytes, hipMemcpyHostToDevice));
+
+``HIP_CHECK`` is a custom macro borrowed from the examples utilities which
+checks the error code returned by API functions for errors and reports them to
+the console. It's not essential to the API.
```
cjatin: It is not essential to the API but it is a good practice to check the error codes of the HIP APIs incase you pass on incorrect values to the API or the API might be out of resources.

docs/tutorial/saxpy.rst
```diff
@@ -0,0 +1,780 @@
...
+  Pointer arguments are pointers to device memory, one typically backed by
+  VRAM.
+- We said that we'll be computing :math:`a\cdot x+y=z`, however we only pass
+  two pointers to the function. We'll be canonically reusing one of the inputs
+  as outputs.
+
+This function is launched from the host using a language extension often called
+the triple chevron syntax. Inside the angle brackets, provide the following.
+
+- The number of blocks to launch (our grid size)
```
cjatin: Do we have any documentation describing grid and block sizes? If yes can we point the user to it from here. I think these are non trivial concepts and user who just wants to exec stuff on their GPU can go it it if they want to.

docs/tutorial/saxpy.rst
```diff
@@ -0,0 +1,780 @@
...
+      }
+  }
+
+- The unique linear index identifying the thread is computed from the block id
+  the thread is a member of, the block's size and the id of the thread within
+  the block.
+- A check is made to avoid overindexing the input.
+- The useful part of the computation is carried out.
+
+Retrieval of the result from the device is done much like its dispatching:
```
cjatin: dispatching can mean different thing in context of HIP, can we change the wording here. Also the first memcpy should also be explained and here we can mention in simple terms that we are copying result back to host.

docs/tutorial/saxpy.rst
```diff
@@ -0,0 +1,780 @@
...
+      nvcc ./HIP-Basic/saxpy/main.hip -o saxpy -I ./Common -I /opt/rocm/include -O2 -x cu
+
+  .. tab-item:: Windows and AMD
+    :sync: windows-amd
+
+    .. code-block:: powershell
+
+      clang++ .\HIP-Basic\saxpy\main.hip -o saxpy.exe -I .\Common -lamdhip64 -L ${env:HIP_PATH}lib -O2
+
+  .. tab-item:: Windows and NVIDIA
```
cjatin, line 322: Have you tested this particular config? windows + nvidia?

docs/tutorial/saxpy.rst
```diff
@@ -0,0 +1,780 @@
...
+device and back. It has also discussed the device code as source, with the HIP
+runtime arguing that the correct binary to dispatch for execution. How can you
+find out what device binary flavors are embedded into the executable?
+
+.. tab-set::
+
+  .. tab-item:: Linux and AMD
+    :sync: linux-amd
+
+    The utilities included with ROCm help significantly to inspect binary
```
cjatin, line 345: I think binary inspection deserves its own doc, rather then putting it alongside saxpy explanation. In my head person looking for saxpy documentation will be least bothered with code object extraction and the person wanting to play with code objects will be least bothered with saxpy.

cjatin (changes requested):

docs/tutorial/saxpy.rst
neon60: This is defined by our documentation styling guide, and we are using the same at other locations. "The full text and then the acronym in parenthesis, and the following references use the acronym."

Example:
https://github.com/ROCm/HIP/blob/develop/docs/index.md

docs/tutorial/saxpy.rst
neon60: Added.

docs/tutorial/saxpy.rst
neon60: Added.

docs/tutorial/saxpy.rst
neon60: Added.

docs/tutorial/saxpy.rst
neon60: We can put this feedback to a separate issue. Currently, we do not have binary inspection tutorial and at least we have some here now. 

If we finally have the binary inspection page, we can rethink this suggestion.  Maybe I can put this into a dropdown:
https://sphinx-design.readthedocs.io/en/latest/dropdowns.html

What do you think?

docs/tutorial/saxpy.rst
neon60: Removed dispatching.

docs/tutorial/saxpy.rst
neon60: As I know, all the commands were tested. I can test this, today. Do you so problems with it?

docs/tutorial/saxpy.rst
cjatin: I would rather have its own page, playing with code objects might require a bit more detail and populating here will overcomplicate this tutorial.

docs/tutorial/saxpy.rst
cjatin: I never really tested this config so was not sure. If it works I am more than OK with it.

cjatin (approved):
Looks good

docs/tutorial/saxpy.rst
neon60: OK. I am creating a separate issue to remove the code inspection from here to a new code inspection tutorial. 

The Boy Scout Rule: Leave the campground cleaner than you found it. → I am merging this in. One tutorial better than zero, and we will improve this later.

neon60:
> Looks good

Thanks for the review.

PR #3532, opened 2024-06-27
Programming interface

neon60:
@randyh62 The programming interface currently.


docs/understand/programming_interface.rst
```diff
@@ -0,0 +1,42 @@
+.. meta::
+  :description: This chapter describes the HIP runtime API and the compilation workflow of the HIP compilers.
+  :keywords: AMD, ROCm, HIP, CUDA, HIP runtime API
+
+*******************************************************************************
+Programming interface
+*******************************************************************************
+
+The programming interface document will focus on the HIP runtime API. The runtime API provides C and C++ functions for event, stream,  device and memory managements, etc. The HIP runtime on AMD platform uses the Common Language Runtimes (CLR), while on NVIDIA platform HIP runtime is only a thin layer over the CUDA runtime.
```
randyh62: This document focuses on the HIP Runtime API and the functions provided for event, stream, device and memory management. The HIP runtime on AMD platform uses the Common Language Runtimes (CLR), while on NVIDIA platform HIP runtime is only a thin layer over the CUDA runtime.

docs/understand/programming_interface.rst
```diff
@@ -0,0 +1,42 @@
...
+
+- **CLR** contains source codes for AMD's compute languages runtimes: ``HIP`` and ``OpenCL™``. CLR includes the implementation of ``HIP`` language on the AMD platform `hipamd <https://github.com/ROCm/clr/tree/develop/hipamd>`_ and the Radeon Open Compute Common Language Runtime (rocclr). rocclr is a virtual device interface, that HIP runtime interact with different backends such as ROCr on Linux or PAL on Windows. (CLR also include the implementation of `OpenCL <https://github.com/ROCm/clr/tree/develop/opencl>`_, while it's interact with ROCr and PAL)
```
randyh62: **CLR** contains source codes for AMD's common languages runtimes:

docs/understand/programming_interface.rst
```diff
@@ -0,0 +1,42 @@
...
+- **CUDA runtime** is built over the CUDA driver API (lower-level C API). For further information about the CUDA driver and runtime API, check the :doc:`hip:how-to/hip_porting_driver_api`. On non-AMD platform, HIP runtime determines, if CUDA is available and can be used. If available, HIP_PLATFORM is set to ``nvidia`` and underneath CUDA path is used.
+
+.. I am not sure we should share this.
```
randyh62: Then pull it.

docs/understand/programming_interface.rst
```diff
@@ -0,0 +1,42 @@
...
+The different runtimes interactions are represented on the following figure.
+
+.. figure:: ../data/understand/programming_interface/runtimes.svg
+
+HIP compilers
+=============
+
+The HIP runtime API and HIP C++ extensions are available with HIP compilers. On AMD platform ROCm currently provides two compiler interfaces: ``hipcc`` and ``amdclang++``. The ``hipcc`` command-line interface aims to provide a more familiar user interface to users who are experienced in CUDA but relatively new to the ROCm/HIP development environment. On the other hand, ``amdclang++`` provides a user interface identical to the clang++ compiler. (For further details, check `llvm <llvm-project-docs:index>`_). On NVIDIA platform ``hipcc`` invoke the locally installed ``NVCC`` compiler, while on AMD platform it's invoke ``amdclang++``.
```
randyh62: The HIP runtime API and HIP C++ extensions are available with HIP compilers. On AMD platform ROCm currently provides two compiler interfaces: ``amdclang++`` and ``hipcc``. The ``hipcc`` command-line interface aims to provide a more familiar user interface to users who are experienced in CUDA but relatively new to the ROCm/HIP development environment. On the other hand, ``amdclang++`` provides a user interface identical to the clang++ compiler. See `llvm <llvm-project-docs:index>`_ for more information. On NVIDIA platform ``hipcc`` invokes a locally installed ``NVCC`` compiler.

docs/understand/programming_interface.rst
```diff
@@ -0,0 +1,42 @@
...
+
+.. Need to update the link later.
+For AMD compiler options, check the `GPU compiler option page <https://rocm.docs.amd.com/en/docs-5.2.3/reference/rocmcc/rocmcc.html#amd-gpu-compilation>`_.
```
randyh62: For AMD compiler options, see :doc:`ROCm compilers reference <llvm-project:reference/rocmcc>`.

docs/understand/programming_interface.rst
```diff
@@ -0,0 +1,42 @@
...
+
+HIP compilation workflow
+------------------------
+
+The source code compiled with HIP compilers are separated to device code and host. The HIP compilers:
+
+.. WIP
+
+* Compiling the device code into an assembly.
```
randyh62: Compile the device code into an assembly object file

docs/understand/programming_interface.rst
```diff
@@ -0,0 +1,42 @@
...
+* Modify the host code by replacing the <<<...>>> syntax introduced in Kernels by the necessary CUDA runtime function calls to load and launch each compiled kernel from the PTX code and/or cubin object.
```
randyh62: Modify the host code by replacing the triple chevron syntax (<<<...>>>) introduced in Kernels by the necessary CUDA runtime function calls to load and launch each compiled kernel from the PTX code and/or cubin object.

randyh62: Is this necessary? Don't both HIP and CUDA support triple chevron syntax?

docs/understand/programming_interface.rst
```diff
@@ -0,0 +1,42 @@
...
+
+``NVCC`` and ``amdclang++`` target different architectures and use different code object formats: ``NVCC`` is ``cubin`` or ``ptx`` files, while the ``amdclang++`` path is the ``hsaco`` format.
+
+For example of compiling from command line, check the :ref:` SAXPY tutorial compiling <compiling_on_the_command_line>`.
+
+HIP runtime API
```
randyh62: This appears to be an orphaned heading? There is nothing following it?

.wordlist.txt
```diff
@@ -17,6 +17,7 @@ clr
 coroutines
 cuBLASLt
 cuCtx
+CUDA's
```
MKKnorr, line 20: Did this really get caught by the spellcheck? "CUDA" itself passes fine, and an apostrophe shouldn't trigger the spellcheck

docs/.gitignore
```diff
@@ -6,3 +6,5 @@
 /doxygen/html
 /doxygen/xml
 /sphinx/_toc.yml
+understand/user-kernel-space-compat-matrix-content.rst
+understand/algorithm-libraries-support-matrix-content.rst
```
MKKnorr: This seems like it shouldn't belong in here :)

docs/conf.py
```diff
@@ -6,6 +6,7 @@
 
 import re
 from typing import Any, Dict, List
+from urllib.request import urlretrieve
```
MKKnorr: What is this additional import needed for? I don't see any changes that would require this

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+*******************************************************************************
+HIP runtime API
+*******************************************************************************
+
+The HIP runtime API is the programming interface, which provides C and C++
+functions for event, stream,  device and memory managements, etc. The HIP
+runtime on AMD platform uses the :doc:`Common Language Runtime (CLR) <hip:understand/amd_clr>`,
+while on NVIDIA platform HIP runtime is only a thin layer over the CUDA runtime
+or Driver API.
```
MKKnorr:
```suggestion
The HIP runtime API provides C and C++ functionality to manage GPUs, like event, stream and memory management.
On AMD platforms the HIP runtime uses the :doc:`Common Language Runtime (CLR) <hip:understand/amd_clr>`,
while on NVIDIA platforms it is only a thin layer over the CUDA runtime or Driver API.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+- **CLR** contains source codes for AMD's compute languages runtimes: ``HIP``
```
MKKnorr:
```suggestion
- **CLR** contains source code for AMD's compute language runtimes: ``HIP``
```
Is it common or compute language runtimes? Even the clr repository doesn't agree with itself on this - the readme title uses "compute", and the explanation immediately beneath uses "common". But at least this document should be consistent

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+  and ``OpenCL™``. CLR includes the implementation of ``HIP`` language on the AMD
```
MKKnorr:
```suggestion
  and ``OpenCL™``. CLR includes the implementation of the ``HIP`` language on the AMD
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+  platform `hipamd <https://github.com/ROCm/clr/tree/develop/hipamd>`_ and the
+  Radeon Open Compute Common Language Runtime (rocclr). rocclr is a virtual device
+  interface, that HIP runtime interact with different backends such as ROCr on
```
MKKnorr:
```suggestion
  interface, that enables the HIP runtime to interact with different backends such as ROCr on
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+  Linux or PAL on Windows. (CLR also include the implementation of `OpenCL <https://github.com/ROCm/clr/tree/develop/opencl>`_,
+  while it's interact with ROCr and PAL)
```
MKKnorr: I can't make sense of the second part of this sentence. It's also questionable whether this note about OpenCL belongs here at all

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+- **CUDA runtime** is built over the CUDA driver API (lower-level C API). For
+  further information about the CUDA driver and runtime API, check the :doc:`hip:how-to/hip_porting_driver_api`.
```
MKKnorr:
```suggestion
- The **CUDA runtime** is built on top of the CUDA driver API, which is a C API with lower-level access to NVIDIA GPUs.
  For further information about the CUDA driver and runtime API and its relation to HIP check the :doc:`CUDA driver API porting guide<hip:how-to/hip_porting_driver_api>`.
```
Either drop the part in parantheses or integrate it into the sentence

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+  On non-AMD platform, HIP runtime determines, if CUDA is available and can be
+  used. If available, HIP_PLATFORM is set to ``nvidia`` and underneath CUDA path
+  is used.
```
MKKnorr:
```suggestion
  hipcc determines if CUDA is available. If it is available HIP_PLATFORM is set to ``nvidia`` and the CUDA compilation toolchain is used.
```
Does this even need to be here? This seems like something that belongs to the compilation workflow

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+The different runtimes interactions are represented on the following figure.
```
MKKnorr:
```suggestion
The relation between the different runtimes and their backends is presented in the following figure.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+The HIP runtime API and HIP C++ extensions are available with HIP compilers. On
+AMD platform ROCm currently provides two compiler interfaces: ``hipcc`` and
+``amdclang++``. The ``hipcc`` command-line interface aims to provide a more
+familiar user interface to users who are experienced in CUDA but relatively new
+to the ROCm/HIP development environment. On the other hand, ``amdclang++``
+provides a user interface identical to the clang++ compiler. (For further
+details, check :doc:`llvm <llvm-project:index>`). On NVIDIA platform ``hipcc``
+invoke the locally installed ``NVCC`` compiler, while on AMD platform it's
+invoke ``amdclang++``.
```
MKKnorr: The hip runtime/host-side can also be compiled with a "normal" compiler (see https://github.com/ROCm/HIP/blob/amd-staging/include/hip/hip_runtime_api.h#L27)

Also the focus of hipcc should lie more on portability - using amdclang++ is not portable to nvidia systems afaik. At least not easily

```suggestion

ROCm provides the compiler driver ``hipcc``, that can be used on AMD and NVIDIA platforms.
hipcc takes care of setting the default library and include paths for HIP, as well as some environment variables,
and takes care of invoking the appropriate compiler - ``amdclang++`` on AMD platforms and ``nvcc`` on NVIDIA platforms.
``amdclang++`` is based on the ``clang++`` compiler. For further details, check :doc:`the llvm project<llvm-project:index>`.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+.. Need to update the link later.
+
+For AMD compiler options, check the :doc:`GPU compiler option page <llvm-project:reference/rocmcc>`.
```
MKKnorr:
```suggestion
For AMD compiler options, check the :doc:`GPU compiler options page <llvm-project:reference/rocmcc>`.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+HIP compilation workflow
+------------------------
+
+Offline compilation
+^^^^^^^^^^^^^^^^^^^
+
+The source code compiled with HIP compilers are separated to device code and
+host.
```
MKKnorr: The splitting of the files is more of an implementation detail, it's more important to highlight the different compilation stages
```suggestion
The compilation of HIP code is separated into a host- and a device-code compilation stage.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+Offline compilation
+^^^^^^^^^^^^^^^^^^^
+
+The source code compiled with HIP compilers are separated to device code and
+host. 
+
+The device code compiled into assembly and/or binary. ``NVCC`` and 
+``amdclang++`` target different architectures and use different code object
+formats: ``NVCC`` is ``cubin`` or ``PTX`` files, while the ``amdclang++`` path
+is the ``hsaco`` format.
```
MKKnorr: Slight change of wording to emphasize how device code is integrated into the binary.
The last line that I added for NVIDIA can also be ommited, but would be a nice info
```suggestion
The compiled device code is embedded into the host object file. Depending on the platform,
the device code can be compiled into assembly or binary. ``nvcc`` and 
``amdclang++`` target different architectures and use different code object
formats: ``nvcc`` uses the binary ``cubin`` or the assembly ``PTX`` files, while the ``amdclang++`` path
is the binary ``hsaco`` format. On NVIDIA platforms the driver takes care of compiling the PTX files to executable code during runtime.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+host. 
+
+The device code compiled into assembly and/or binary. ``NVCC`` and 
+``amdclang++`` target different architectures and use different code object
+formats: ``NVCC`` is ``cubin`` or ``PTX`` files, while the ``amdclang++`` path
+is the ``hsaco`` format. 
+
+The ``NVCC`` at the host side only replacing the ``<<<...>>>``
+syntax with CUDA runtime functions calls and the modified host code is the input
+of the default host compiler. ``hipcc`` or ``amdclang++`` can compile the host
```
MKKnorr:
```suggestion
On the host side ``nvcc`` only replaces the ``<<<...>>>``
kernel launch syntax with the appropriate CUDA runtime function call and the modified host code is passed
to the default host compiler. ``hipcc`` or ``amdclang++`` can compile the host
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+code in one step without other C++ compilers.
```
MKKnorr: It's debateable whether it can be considered one step for hipcc, as that is a wrapper around clang/nvcc. But that's just nitpicking from my side :)

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+For example of compiling from command line, check the :ref:`SAXPY tutorial compiling <compiling_on_the_command_line>` .
```
MKKnorr:
```suggestion
An example for how to compile HIP from the command line can be found in the :ref:`SAXPY tutorial<compiling_on_the_command_line>` .
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+.. _driver_api_understand:
+
+Just-in-time compilation
```
MKKnorr:
```suggestion
Runtime compilation
```
I think runtime and just-in-time compilation are slightly different, but the main argument is: it's called (hip)rtc - *R*un *T*ime Compilation, so let's call it that to avoid confusion

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+HIP lets you compile kernels at runtime with the `hiprtc*` APIs. Kernels can be
+stored as a text string and can be passed to HIPRTC APIs alongside options to
+guide the compilation.
```
MKKnorr:
```suggestion
HIP lets you compile kernels at runtime with the `hiprtc*` API. Kernels are
stored as a text string that are then passed to HIPRTC alongside options to
guide the compilation.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+For further details, check the :doc:`../how-to/hip_rtc` .
```
MKKnorr:
```suggestion
For further details, check the :doc:`how-to section for the HIP runtime compilation<../how-to/hip_rtc>`.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+Driver API 
+===========
+
+The driver API offers developers low-level control over GPU operations, enabling
+them to manage GPU resources, load and launch kernels, and handle memory
+explicitly. In HIP the Driver API is part of the HIP runtime API, while the
+CUDA's driver API is separate from CUDA runtime API.
```
MKKnorr:
```suggestion
explicitly. In HIP the driver API is part of the HIP runtime API, while
CUDA's driver API is separate from CUDA runtime API.
```
I would actually not even mention a **HIP** driver API, because this suggests that there is a separate API within the HIP runtime API. But this would require a rewrite of almost the whole section, so feel free to ignore this
```suggestion
explicitly. Unlike CUDA, where the runtime API is separate from the driver API,
HIP provides all its functionality within the runtime API.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+One significant advantage of the driver API is its ability to dynamically load
+and manage code objects, which is particularly useful for applications that need
+to generate or modify kernels at runtime. This flexibility allows for more
+sophisticated and adaptable GPU programming.
+
+Unlike the runtime API, the driver API does not automatically handle tasks such
+as context creation and kernel loading. While the runtime API is more convenient
```
MKKnorr: Isn't this only valid for CUDA? For HIP this is always created automatically afaik

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+and easier to use for most applications, the driver API provides greater control
+and can be more efficient for complex or performance-critical applications.
```
MKKnorr: Again, this is quite specific to CUDA, as HIP doesn't differentiate between driver and runtime API

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+Using the driver API can result in longer development times due to the need for
+more detailed code and explicit management. However, the actual runtime
+performance can be similar to or even better than the runtime API, depending on
+how well the application is optimized.
```
MKKnorr: This paragraph seems a bit weird - it states that coding with the driver API takes longer and is more complicated, and then states that the runtime "can be similar" "or even better", implying that usually the performance isn't much better.

So it takes a lot of effort for hardly any benefit?

MKKnorr: This paragraph seems a bit weird - it states that coding with the driver API takes longer and is more complicated, and then states that the runtime "can be similar" "or even better", implying that usually the performance isn't much better.

So it takes a lot of effort for hardly any benefit?

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+For further details, check :ref:`porting_driver_api`, and :ref:`driver_api_reference`.
```
MKKnorr: At least the first reference would already exist. No matter whether you change the ref, the text displayed for the references should be changed to better fit into the text
```suggestion
For further details, check the :doc:`CUDA driver API porting guide<hip:how-to/hip_porting_driver_api>`, and the :ref:`driver API reference<driver_api_reference>`.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+Memory Management
+=================
+
+Memory management is important part of HIP runtime API, when creating
+high-performance applications in the HIP ecosystem. Both allocating and copying
```
MKKnorr: The HIP ecosystem is somewhat implied by the HIP runtime API
```suggestion
Memory management is an important part of the HIP runtime API, when creating
high-performance applications. Both allocating and copying
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+memory can result in bottlenecks, which can significantly impact performance.
+
+C language style ``hipMalloc`` typically used to allocate a  block of memory on 
+the device memory and ``hipFree`` to free memory. There are features like
+managed memory, virtual memory or Stream ordered memory allocator which are
+described under.
```
MKKnorr:
```suggestion
For basic device memory management, HIP uses the C-style functions :cpp:func:`hipMalloc`
for allocating and :cpp:func:`hipFree` for freeing memory. There are advanced features like
managed memory, virtual memory or stream ordered memory allocator which are
described in the following sections.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+Device memory
+-------------
+
+Device memory exists on the device (e.g. GPU) of the machine in video random
+access memory (VRAM) and kernels are operating on these. Device memory can be
```
MKKnorr: It's only called video RAM on GPUs. Also it would be nice to add an explanation what benefits it provides.

The "usually" in my suggestion can also be dropped, it's just there for APUs

```suggestion
Device memory exists on the device, e.g. on GPUs in the video random
access memory (VRAM), and is accessible by the kernels operating on the device. It is usually orders of magnitude faster than the transfers between the host and the device. Device memory can be 
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+Device memory exists on the device (e.g. GPU) of the machine in video random
+access memory (VRAM) and kernels are operating on these. Device memory can be 
+allocated as global memory, constant, texture or surface memory.
+
+Global memory
+^^^^^^^^^^^^^
+
+Read-write storage visible to all threads in a given grid. There are specialized
+versions of global memory with different usage semantics which are typically
+backed by the same hardware storing global.
```
MKKnorr: Just out of curiosity, what are the specialized versions? If they are the following types, then the cache path is the main difference, afaik.
Also, while technically correct, that it's visible to all threads in a given grid, it persists across kernel launches and its lifetime is controlled by the host
```suggestion
Read-write storage visible to all threads on a given device. There are specialized
versions of global memory with different usage semantics which are typically
backed by the same hardware, but can use different caching paths.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+Constant memory
+^^^^^^^^^^^^^
+
+Read-only storage visible to all threads in a given grid. It is a limited 
+segment of global with queryable size.
```
MKKnorr: I think this should highlight the differences to global memory, i.e. (possibly) different caching, and performance recommendations.
Also I'm not sure if it's really a segment of global memory (which is a logical partition), but rather part of device memory (as in physically located there)
```suggestion
Read-only storage visible to all threadson a given device. It is a limited 
segment backed by device memory with queryable size. It needs to be set by the host before kernel execution.
Constant memory provides the best performance benefit when all threads within a warp access the same address.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+Texture memory
+^^^^^^^^^^^^^^
+
+Read-only storage visible to all threads in a given grid and accessible
+through additional APIs.
```
MKKnorr: The use case should be highlighted (my wording might not be the best, but the 2D use case and the graphics origins should be mentioned)
```suggestion
Read-only storage visible to all threads on a given device and accessible
through additional APIs. Its origins come from graphics APIs, and provides
performance benefits when accessing memory in a pattern where the
addresses are close to each other in a 2D representation of the memory.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+A read-write version of texture memory.
+
+Managed memory (Unified memory)
+-------------------------------
+
+Unified Memory is a single memory address space accessible from any processor
+within a system. This setup simplifies memory management processes and enables
+applications to allocate data that can be read or written by code running on
+either CPUs or GPUs. The Unified memory model is shown in the following figure.
```
MKKnorr: There is no figure here :)

Also, should the performance implications be mentioned here?

MKKnorr (commented):
Good job overall, mostly just nitpicking about wording or adding details

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+Stream ordered memory allocator
+-------------------------------
+
+Stream Ordered Memory Allocator (SOMA) provides an asynchronous memory
+allocation mechanism with stream-ordering semantics. You can use SOMA to
+allocate and free memory in stream order, which ensures that all asynchronous
+accesses occur between the stream executions of allocation and deallocation.
+Compliance with stream order prevents use-before-allocation or use-after-free
+errors, which helps to avoid an undefined behavior.
```
MKKnorr:
```suggestion
errors, which helps to avoid undefined behavior.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+applications to allocate data that can be read or written by code running on
+either CPUs or GPUs. The Unified memory model is shown in the following figure.
+
+Stream ordered memory allocator
+-------------------------------
+
+Stream Ordered Memory Allocator (SOMA) provides an asynchronous memory
+allocation mechanism with stream-ordering semantics. You can use SOMA to
+allocate and free memory in stream order, which ensures that all asynchronous
+accesses occur between the stream executions of allocation and deallocation.
```
MKKnorr:
```suggestion
allocate and free memory in stream order, which ensures that all asynchronous
accesses occur between the stream executions of allocation and deallocation, without the need for device-wide synchronization.
```
It's not as if that behaviour wouldn't be possible otherwise, this is an optimization.

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+-------------------------
+
+Global memory allocation in HIP uses the C language style allocation function.
+This works fine for simple cases but can cause problems if your memory needs
+change. If you need to increase the size of your memory, you must allocate a
+second larger buffer and copy the data to it before you can free the original
+buffer. This increases overall memory usage and causes unnecessary ``memcpy``
+calls. Another solution is to allocate a larger buffer than you initially need.
+However, this isn't an efficient way to handle resources and doesn't solve the
+issue of reallocation when the extra buffer runs out.
```
MKKnorr:
```suggestion
Global memory allocations in HIP use the C-style allocation functions.
This works fine for simple cases but can cause problems if the memory needs
to be reallocated. If you need to increase the size of your memory, you must allocate a
second larger buffer and copy the data to it before you can free the original
buffer. This temporarily requires a lot more memory and causes unnecessary ``hipMemcpy``
calls. Another solution is to allocate a larger buffer than initially needed.
However, this is not an efficient way to handle resources and doesn't solve the
issue of reallocation when more memory than originally expected is needed.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+Virtual memory management solves these memory management problems. It helps to
+reduce memory usage and unnecessary ``memcpy`` calls.
```
MKKnorr: Could you add some context as for how it solves this problem? Up until here it only showcases the problems
```suggestion
Virtual memory management solves these problems. It helps to
limit memory usage to the actually needed amount and avoids unnecessary ``hipMemcpy`` calls.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+For further details, check `HIP Runtime API Reference <doxygen/html/index.html>`_.
```
MKKnorr: Please link to the virtual memory management page directly instead
```suggestion
For further details, check `HIP Runtime API Reference <../doxygen/html/group___virtual.html>`_.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+Texture memory management
+-------------------------
+
+Texture memory should be used when you need readonly random access to a 
+larger memory space that cannot be put into shared memory.
+
+For further details, check `HIP Runtime API Reference <doxygen/html/index.html>`_.
```
MKKnorr: Why is texture memory listed again here? It has already been mentioned earlier under "device memory".

Also this description sounds weird to me. Texture memory also resides in global memory, and hardly has any performance benefits over it, except for very specific use cases. It definitely does not replace shared memory.

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+Surface object
+--------------
+
+A Surface Object in GPU programming is a powerful tool that allows for
+read-write access to texture memory, making it particularly useful for
+applications that require direct manipulation of image or array data. By
+providing an API for accessing surface memory, Surface Objects enable complex
+operations like image filtering, simulation updates, and other tasks that
+involve both reading and writing to structured data stored on the GPU.
```
MKKnorr: This section makes it sound like the holy grail of accessing memory. What are the drawbacks and restrictions on these objects? Also it should be noted that the type of array here is the specific hipArray_t (that can be 1D, 2D or 3D), not a general memory buffer.

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+Understanding how to effectively use surface objects is crucial for optimizing
+performance in GPU-accelerated applications that require data modification.
```
MKKnorr: That is indeed true, but I don't see an explanation here under which circumstances it provides a performance benefit

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+For further details, check `HIP Runtime API Reference <doxygen/html/index.html>`_.
```
MKKnorr: Like the other links: directly link to the surface documentation
```suggestion
For further details, check `HIP Runtime API Reference <../doxygen/html/group___surface.html>`_.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+Execution control
+=================
+
+Stream management
+-----------------
+
+Stream management refers to the mechanisms that allow developers to control the
+order and concurrency of kernel executions and memory transfers on the GPU.
```
MKKnorr:
```suggestion
order and concurrency of kernel execution and memory transfers on the GPU.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+Stream is linear sequence of execution which belonging to a specific GPU.
```
MKKnorr:
```suggestion
Streams are associated with a specific device and operations within a stream are executed sequentially.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+Stream management refers to the mechanisms that allow developers to control the
+order and concurrency of kernel executions and memory transfers on the GPU.
+Stream is linear sequence of execution which belonging to a specific GPU. 
+Different streams can execute operations concurrently on the same GPU, which can
+lead to better utilization of the device.
+
+Stream management allows developers to optimize GPU workloads by enabling
+concurrent execution of tasks, overlapping computation with memory transfers,
+and controlling the order of operations. The priority can be also set, which 
+gives extra flexibility in the developers hand.
```
MKKnorr:
```suggestion
and controlling the order of operations. The priority of streams can also be set, which 
provides additional control over task execution.
```

docs/understand/hip_runtime_api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+The stream management concept represented on the following figure.
```
MKKnorr:
```suggestion
The stream management concept is represented in the following figure.
```

docs/.gitignore
neon60: Removed.

docs/conf.py
neon60: Removed.

docs/understand/programming_interface.rst
neon60: fixed.

docs/understand/programming_interface.rst
neon60: Completely changed these sentences.

docs/understand/programming_interface.rst
neon60: Completely changed this section.

docs/understand/programming_interface.rst
neon60: fixed.

.wordlist.txt
neon60: I am creating the issue for this. This PR not focusing on spellchecker improvements.

docs/understand/hip_runtime_api.rst
neon60: Updated.

docs/understand/hip_runtime_api.rst
neon60: Updated and moved this whole section.

docs/understand/hip_runtime_api.rst
neon60: Updated.

docs/understand/hip_runtime_api.rst
neon60: Moved this section.

docs/understand/hip_runtime_api.rst
neon60: Maybe here given grid was better. Updated. Let's discuss offline.

docs/understand/hip_runtime_api.rst
neon60: Updated.

docs/understand/hip_runtime_api.rst
neon60: Updated.

docs/understand/programming_interface.rst
neon60: Out dated comment.

docs/understand/programming_interface.rst
neon60: outdated comment.

PR #3574, opened 2024-08-14
Hardware capabilities

parbenc:
create hardware capabilities page

### DIFF
--- a/docs/reference/hardware_features.rst
+++ b/docs/reference/hardware_features.rst
@@ -1,0 +1,249 @@
+.. meta::
+  :description: This chapter describes the hardware features of the different hardware architectures.
+  :keywords: AMD, ROCm, HIP, hardware, hardware features, hardware architectures
+
+*******************************************************************************
+Hardware features
+*******************************************************************************
+
+This page gives an overview of the different hardware architectures and the
+features they implement. Hardware features do not imply performance, that
+depends on the specifications found in the :doc:`rocm:reference/gpu-arch-specs`
+page.
+
+  .. list-table::
+      :header-rows: 1
+      :name: hardware-features-table
+
+      *
+        - Hardware feature support
+        - RDNA1
+        - CDNA1
+        - RDNA2
+        - CDNA2
+        - RDNA3
+        - CDNA3
+      *
+        - :ref:`atomic functions` on 32-bit integer values in global and shared memory
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+      *
+        - Atomic functions on 64-bit integer values in global and shared memory
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+      *
+        - Atomic addition on 32-bit floating point values in global and shared memory
+        - ❌
+        - ❌
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+      *
+        - Atomic addition on 64-bit floating point values in global memory and shared memory
+        - ❌
+        - ❌
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+      *
+        - :ref:`Warp vote functions <warp_vote_functions>`
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+      *
+        - :ref:`Memory fence instructions <memory_fence_instructions>`
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+      *
+        - :ref:`Synchronization functions <synchronization_functions>`
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+      *
+        - :ref:`Surface functions <surface_object_reference>`
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+      *
+        - :ref:`float16 half precision IEEE-conformant floating-point operations<rocm:precision_support_floating_point_types>`
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+      *
+        - :ref:`bfloat16 16-bit floating-point operations<rocm:precision_support_floating_point_types>`
+        - ❌
+        - ✅
+        - ❌
+        - ✅
+        - ✅
+        - ✅
+      *
+        - Support for :ref:`8-bit floating-point types <rocm:precision_support_floating_point_types>`
+        - ❌
+        - ❌
+        - ❌
+        - ❌
+        - ❌
+        - ✅
+      *
+        - Support for :ref:`tensor float32 <rocm:precision_support_floating_point_types>`
+        - ❌
+        - ❌
+        - ❌
+        - ❌
+        - ❌
+        - ✅
+      *
+        - Packed math with 16-bit floating point values
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+      *
+        - Packed math with 32-bit floating point values
+        - ❌
+        - ❌
+        - ❌
+        - ✅
+        - ❌
+        - ✅
+      *
+        - Matrix Cores
+        - ❌
+        - ✅
+        - ❌
+        - ✅
+        - ❌
+        - ✅
+      *
+        - On-Chip Error Correcting Code (ECC)
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+      *
+        - Maximum dimensionality of grid
+        - 3
+        - 3
+        - 3
+        - 3
+        - 3
+        - 3
+      *
+        - Maximum x-, y- or z-dimension of a grid
+        - :math:`2^{32} - 1`
+        - :math:`2^{32} - 1`
+        - :math:`2^{32} - 1`
+        - :math:`2^{32} - 1`
+        - :math:`2^{32} - 1`
+        - :math:`2^{32} - 1`
+      *
+        - Maximum number of threads per grid
+        - :math:`2^{32} - 1`
+        - :math:`2^{32} - 1`
+        - :math:`2^{32} - 1`
+        - :math:`2^{32} - 1`
+        - :math:`2^{32} - 1`
+        - :math:`2^{32} - 1`
+      *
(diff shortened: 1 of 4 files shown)


neon60:
@parbenc please extend with RDNA

docs/understand/hardware_capabilities.rst
```diff
@@ -0,0 +1,206 @@
+.. meta::
+  :description: This chapter describes the hardware features of the different hardware architectures.
+  :keywords: AMD, ROCm, HIP, hardware, hardware features, hardware architectures
+
+*******************************************************************************
+Hardware features
+*******************************************************************************
+
+This page gives an overview of the different hardware architectures and the features they implement. Hardware features do not imply performance, that depends on the specifications found in the `Accelerator and GPU hardware specifications`_ page.
```
MKKnorr: This link doesn't work. You have to link to the file in the ROCm repository

docs/understand/hardware_capabilities.rst
```diff
@@ -0,0 +1,206 @@
...
+      *
+        - Hardware feature
+        - RDNA1
+        - CDNA1
+        - RDNA2
+        - CDNA2
+        - RDNA3
+        - CDNA3
+      *
+        - Atomic functions on 32-bit integer values in global memory
```
MKKnorr: Linking to the atomic functions for explanation/reference would be a nice touch
https://rocm.docs.amd.com/projects/HIP/en/latest/reference/cpp_language_extensions.html#atomic-functions

docs/understand/hardware_capabilities.rst
```diff
@@ -0,0 +1,206 @@
...
+      *
+        - Atomic addition on 64-bit floating point values in global memory and shared memory
+        - ❌
+        - ❌
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+      *
+        - Error correcting code for SRAMs IP core
```
MKKnorr: It's not really important, that it's an IP core. Can you specify which cache levels are affected by it? Can it be generalized to on-chip memory ECC? I'm not sure what devices support VRAM ECC
```suggestion
        - SRAM Error correcting code
```

docs/understand/hardware_capabilities.rst
```diff
@@ -0,0 +1,206 @@
...
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+      *
+        - Memory page migration
```
MKKnorr: What exactly does this mean? Is this supposed to be the feature needed for "unified memory"?

docs/understand/hardware_capabilities.rst
```diff
@@ -0,0 +1,206 @@
...
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+      *
+        - Thread group split
```
MKKnorr: Is this specifically a "thread group split" feature, or is it rather cooperative groups/thread groups being supported in general?
If it actually is the a thread group split feature, do we have any documentation to link to here?

docs/understand/hardware_capabilities.rst
```diff
@@ -0,0 +1,206 @@
...
+      *
+        - Preload kernel arguments
+        - ❌
+        - ❌
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+      *
+        - Packed math with 32-bit floating point values
```
MKKnorr: What about 16-bit floating point packed math?

docs/understand/hardware_capabilities.rst
```diff
@@ -0,0 +1,206 @@
...
+        - ❌
+        - ❌
+        - ✅
+        - ✅
+        - ✅
+        - ✅
+      *
+        - Support for float8 bfloat8
```
MKKnorr: Link to the corresponding precision support section for those special types
https://rocm.docs.amd.com/en/latest/compatibility/precision-support.html#floating-point-types
(We might have to rework that, as it doesn't specifically mention bfloat8)

docs/understand/hardware_capabilities.rst
```diff
@@ -0,0 +1,206 @@
...
+        - ?
+        - ?
+      *
+        - Maximum number of resident threads per compute unit
+        - ?
+        - ?
+        - ?
+        - ?
+        - ?
+        - ?
```
MKKnorr: These values for warps and threads per compute unit are known. Just depends on what you call a compute unit in the context of rdna. If it's the wgp, then:
(Actually, the blocks per cu are just a deduction from max wavefronts per cu, but I never heard about a restriction on that for amd gpus. The rest comes from the isa guides)
```suggestion
        - Maximum number of resident blocks per compute unit
        - 64
        - 32
        - 64
        - 32
        - 64
        - 32
      *
        - Maximum number of resident wavefronts per compute unit
        - 64
        - 32
        - 64
        - 32
        - 64
        - 32
      *
        - Maximum number of resident threads per compute unit
        - 2048
        - 2048
        - 2048
        - 2048
        - 2048
        - 2048
```

docs/understand/hardware_capabilities.rst
parbenc: I don't know if it specifically has anything to do with unified memory. What it does is allow for memory pages to be moved around, without virtual address change. This can allow unused pages to be moved to system RAM.

docs/understand/hardware_capabilities.rst
parbenc: I don't remember cooperative groups being mentioned. I don't think we have public facing documentation to link here, but anyone can correct me, if I'm wrong.

docs/understand/hardware_capabilities.rst
parbenc: Wasn't mentioned. I assume it's not supported

MKKnorr: It is listed in the "Packed Math" section of all of the ISA guides, seems to be supported on all architectures that are listed here

docs/understand/hardware_capabilities.rst
MKKnorr: Sounds like something that has to do with unified memory. My point is, that if there is no explanation of what this feature is, then what good is this info to a user?

docs/understand/hardware_capabilities.rst
```diff
@@ -0,0 +1,206 @@
...
+        - 3
+        - 3
+      *
+        - Maximum x-, y- or z-dimension of a block
+        - :math:`2^{31} - 1`
+        - :math:`2^{31} - 1`
+        - :math:`2^{31} - 1`
+        - :math:`2^{31} - 1`
+        - :math:`2^{31} - 1`
+        - :math:`2^{31} - 1`
```
MKKnorr: This is most certainly not the case, then one block could span a whole grid

From rocminfo, and I think these numbers haven't changed since the beginning of time:
```suggestion
        - :math:`1024`
        - :math:`1024`
        - :math:`1024`
        - :math:`1024`
        - :math:`1024`
        - :math:`1024`
```

docs/understand/hardware_capabilities.rst
```diff
@@ -0,0 +1,206 @@
...
+      *
+        - Maximum number of threads per block
+        - :math:`2^{31} - 1`
+        - :math:`2^{31} - 1`
+        - :math:`2^{31} - 1`
+        - :math:`2^{31} - 1`
+        - :math:`2^{31} - 1`
+        - :math:`2^{31} - 1`
```
MKKnorr: Same as above. I've also never heard of this limitation changing
```suggestion
        - :math:`1024`
        - :math:`1024`
        - :math:`1024`
        - :math:`1024`
        - :math:`1024`
        - :math:`1024`
```

docs/understand/hardware_capabilities.rst
```diff
@@ -0,0 +1,206 @@
...
+        - 3
+        - 3
+      *
+        - Maximum x-, y- or z-dimension of a grid of thread blocks
+        - :math:`2^{31} - 1`
+        - :math:`2^{31} - 1`
+        - :math:`2^{31} - 1`
+        - :math:`2^{31} - 1`
+        - :math:`2^{31} - 1`
+        - :math:`2^{31} - 1`
```
MKKnorr: The limit is at 32 bit, not 31
```suggestion
        - :math:`2^{32} - 1`
        - :math:`2^{32} - 1`
        - :math:`2^{32} - 1`
        - :math:`2^{32} - 1`
        - :math:`2^{32} - 1`
        - :math:`2^{32} - 1`
```

MKKnorr:
Weak dependency on https://github.com/ROCm/ROCm/pull/3899 for reference to the floating-point types

neon60:
> Weak dependency on [ROCm/ROCm#3899](https://github.com/ROCm/ROCm/pull/3899) for reference to the floating-point types

@MKKnorr I think we want to add this to conceptual. At least that was in the planning document.

neon60 (approved):

docs/reference/hardware_features.rst
```diff
@@ -0,0 +1,249 @@
...
+      :name: hardware-features-table
+
+      *
+        - Hardware feature support
+        - RDNA1
+        - CDNA1
+        - RDNA2
+        - CDNA2
+        - RDNA3
+        - CDNA3
```
randyh62, line 25: There was a recent introduction of the CDNA4 architecture. Is this something we should add to the table?

docs/reference/hardware_features.rst
```diff
@@ -0,0 +1,249 @@
...
+        - Maximum number of resident threads per compute unit
+        - 1280 [2]_
+        - 2048
+        - 1024 [2]_
+        - 2048
+        - 1024 [2]_
+        - 2048
+      *
+        - Maximum number of 32-bit vector registers per thread
+        - RDNA1: 256
```
randyh62: Since the column head indicates the RDNA or CDNA architecture, do we need to repeat it here in the cell?

randyh62 (approved):

docs/reference/hardware_features.rst
MKKnorr: As long as the GPUs are not released I would refrain from adding information here

docs/reference/hardware_features.rst
MKKnorr: Yep, overlooked that - that was for me to keep track of what column the values are for 😅

PR #3617, opened 2024-09-19
Add expected behavior to unified memory

matyas-streamhpc:


docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
```diff
@@ -63,32 +82,30 @@ Unified memory is supported on Linux by all modern AMD GPUs from the Vega
 series onward. Unified memory management can be achieved with managed memory
 allocation and, for the latest GPUs, with a system allocator.
 
-The table below lists the supported allocators. The allocators are described in
-the next section.
+The table below lists the supported allocators for the different architectures.
+The allocators are described in the next section.
 
 .. list-table:: Supported Unified Memory Allocators
     :widths: 40, 25, 25, 25
```
neon60: Now, you have three columns and not four

neon60 (approved):

docs/how-to/hip_runtime_api.rst
```diff
@@ -0,0 +1,35 @@
...
+  :keywords: AMD, ROCm, HIP, CUDA, HIP runtime API How to,
+
+.. _hip_runtime_api_how-to:
+
+********************************************************************************
+HIP runtime API
+********************************************************************************
+
+The HIP runtime API provides C and C++ functionalities to manage event, stream,
+and memory on GPUs. On AMD ROCm software, the HIP runtime uses :doc:`Common
```
randyh62:
```suggestion
and memory on GPUs. On AMD ROCm software, the HIP runtime uses :doc:`Compute
```

randyh62: Apparently CLR is supposed to be Compute Language Runtime

docs/how-to/hip_runtime_api/memory_management.rst
```diff
@@ -0,0 +1,50 @@
...
+bottlenecks, which can significantly impact performance.
+
+The programming model is based on a system with a host and a device, each having
+its own distinct memory. Kernels operate on :ref:`device_memory`, while host functions
+operate on :ref:`host_memory`.
+
+The runtime offers functions for allocating, freeing, and copying device memory,
+along with transferring data between host and device memory.
+
+Here are the various memory management techniques:
```
randyh62: It seems like these techniques could use a brief description?

docs/how-to/hip_runtime_api/memory_management/host_memory.rst
```diff
@@ -0,0 +1,217 @@
...
+* Pageable memory
+* Pinned memory
+
+The following figure explains how data is transferred in pageable and pinned
+memory.
+
+.. figure:: ../../../data/how-to/hip_runtime_api/memory_management/pageable_pinned.svg
+
+The pageable and pinned memory allow you to exercise direct control over
+memory operations, which is known as explicit memory management. When using the
```
randyh62: In _memory_management.rst_ we list four memory management techniques, none of which is "explicit memory management". Is it possible to relate this explicit memory management to one of the four techniques? Or should it be listed as one of the techniques?

docs/how-to/hip_runtime_api/memory_management/host_memory.rst
```diff
@@ -0,0 +1,217 @@
...
+      HIP_CHECK(hipMalloc((int **)&device_output, element_number * sizeof(int)));
+
+      // Device data preparation
+      HIP_CHECK(hipMemcpy(device_input, host_input, element_number * sizeof(int), hipMemcpyHostToDevice));
+      HIP_CHECK(hipMemset(device_output, 0, element_number * sizeof(int)));
+
+      // Run the kernel
+      // ...
+
+      HIP_CHECK(hipMemcpy(device_input, host_input, element_number * sizeof(int), hipMemcpyHostToDevice));
```
randyh62: This doesn't seem right to me. We have device_output and host_output, shouldn't those come into play here?

docs/how-to/hip_runtime_api/memory_management/host_memory.rst
```diff
@@ -0,0 +1,217 @@
...
+  :cpp:func:`hipMalloc` and :cpp:func:`hipFree` are blocking calls. However, HIP
+  also provides non-blocking versions :cpp:func:`hipMallocAsync` and
+  :cpp:func:`hipFreeAsync`, which require a stream as an additional argument.
+
+.. _pinned_host_memory:
+
+Pinned memory
+================================================================================
+
+Pinned memory or page-locked memory is stored in pages that are locked in specific sectors in RAM and can't be migrated. The pointer can be used on both
```
randyh62: Seems like we should note that pinned memory is allocated by hipMalloc() and hipHostMalloc().

docs/how-to/hip_runtime_api/memory_management/host_memory.rst
```diff
@@ -0,0 +1,217 @@
...
+      HIP_CHECK(hipMalloc((int **)&device_output, element_number * sizeof(int)));
+
+      // Device data preparation
+      HIP_CHECK(hipMemcpy(device_input, host_input, element_number * sizeof(int), hipMemcpyHostToDevice));
+      HIP_CHECK(hipMemset(device_output, 0, element_number * sizeof(int)));
+
+      // Run the kernel
+      // ...
+
+      HIP_CHECK(hipMemcpy(device_input, host_input, element_number * sizeof(int), hipMemcpyHostToDevice));
```
randyh62: Shouldn't this be DeviceToHost?

docs/how-to/hip_runtime_api/memory_management/device_memory.rst
```diff
@@ -0,0 +1,52 @@
...
+*******************************************************************************
+
+Device memory exists on the device, e.g. on GPUs in the video random access
+memory (VRAM), and is accessible by the kernels operating on the device. Recent
+architectures use graphics double data rate (GDDR) synchronous dynamic
+random-access memory (SDRAM) such as GDDR6, or high-bandwidth memory (HBM) such
+as HBM2e. Device memory can be allocated as global memory, constant, texture or
+surface memory.
+
+Global memory
```
randyh62: I suggest making this a bulleted list rather than four separate subheads.

docs/how-to/hip_runtime_api/memory_management/device_memory.rst
```diff
@@ -0,0 +1,52 @@
...
+                ROCm software.
+  :keywords: AMD, ROCm, HIP, device memory
+
+.. _device_memory:
+
+*******************************************************************************
+Device memory
+*******************************************************************************
+
+Device memory exists on the device, e.g. on GPUs in the video random access
```
randyh62: Device memory is allocated with hipMalloc, but is there some way to specify the type of device memory given the choices of Global, Constant, Texture, or Surface?

docs/how-to/hip_runtime_api/memory_management/coherence_control.rst
```diff
@@ -0,0 +1,180 @@
...
+      - Fine-grained
+    * - ``malloc``
+      -
+      - ``hipMemAdviseSetCoarseGrain``
+      - Coarse-grained
+
+:sup:`1` The :cpp:func:`hipHostMalloc` memory allocation coherence mode can be 
+affected by the ``HIP_HOST_COHERENT`` environment variable, if the
+``hipHostMallocCoherent``, ``hipHostMallocNonCoherent``, and
+``hipHostMallocMapped`` are unset. If neither these flags nor the
```
randyh62: Here the footnote to the table references _hipHostMallocMapped_ flag, but the table above does not show this flag. This also is mentioned in the following note.

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
```diff
@@ -0,0 +1,737 @@
...
+
+.. _unified_memory:
+
+*******************************************************************************
+Unified memory management
+*******************************************************************************
+
+In conventional architectures CPUs and attached devices have their own memory
+space and dedicated physical memory backing it up, e.g. normal RAM for CPUs and
+VRAM on GPUs. This way each device can have physical memory optimized for its
```
randyh62, line 14:
```suggestion
VRAM on GPUs. This way each device has physical memory optimized for its
```

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
```diff
@@ -0,0 +1,737 @@
...
+use case. Especially GPUs usually have specialized memory whose bandwidth is a
```
randyh62:
```suggestion
use case. GPUs usually have specialized memory whose bandwidth is a
```

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
```diff
@@ -0,0 +1,737 @@
...
+magnitude higher than the RAM attached to CPUs.
+
+While providing exceptional performance, this setup typically requires explicit
+memory management, as memory needs to be allocated, copied and freed on the used
+devices and on the host. Additionally, this makes using more than the physically
+available memory on the devices complicated.
+
+Modern GPUs circumvent the problem of having to explicitly manage the memory,
```
randyh62, line 23: Is this a feature of the "Modern GPUs" or of the runtime?

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
```diff
@@ -0,0 +1,737 @@
...
+the kernel.
+
+.. _unified memory system requirements:
+
+System requirements
+================================================================================
+
+Unified memory is supported on Linux by all modern AMD GPUs from the Vega
+series onward. Unified memory management can be achieved with managed memory
+allocation and, for the latest GPUs, with a system allocator.
```
randyh62:
```suggestion
allocation using hipMallocManaged(), and for the latest GPUs with a system allocator, as shown in the following table.
```

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
```diff
@@ -0,0 +1,737 @@
...
+
+The table below lists the supported allocators for the different architectures.
```
randyh62: I think you can eliminate this paragraph after noting it in the prior paragraph.

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
```diff
@@ -0,0 +1,737 @@
...
+The allocators are described in the next section.
```
randyh62:
```suggestion

```

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
```diff
@@ -0,0 +1,737 @@
...
+
+- **System allocated unified memory**
+
+  Starting with CDNA2, the ``new`` and `malloc()`` system allocators allow
+  you to reserve unified memory. The system allocator is more versatile and
+  offers an easy transition for code written for CPUs to HIP code as the
+  same system allocation API is used.
+
+To ensure the proper functioning of system allocated unified memory on supported
+graphics cards, it is essential to configure the environment variable
```
randyh62:
```suggestion
accelerators and graphics cards, it is essential to configure the environment variable
```

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
```diff
@@ -0,0 +1,737 @@
...
+``XNACK=1`` and use a kernel that supports Heterogeneous Memory Management
```
randyh62: Do we have a reference for Heterogeneous Memory Management?

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
```diff
@@ -0,0 +1,737 @@
...
+
+.. _using unified memory:
+
+Using unified memory
+================================================================================
+
+Unified memory can simplify the complexities of memory management in GPU
+computing, by not requiring explicit copies between the host and the devices. It
+can be particularly useful in use cases with sparse memory accesses from both
+the CPU and the GPU, as not the whole memory region needs to be transferred to
```
randyh62:
```suggestion
the CPU and the GPU, as the whole memory region does not need to be transferred to
```

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
```diff
@@ -0,0 +1,737 @@
...
+
+There are several ways, in which the developer can guide the runtime to reduce
+copies between devices, in order to improve performance.
+
+Data prefetching
+--------------------------------------------------------------------------------
+
+Data prefetching is a technique used to improve the performance of your
+application by moving data to the desired device before it's actually
+needed. ``hipCpuDeviceId`` is a special constant to specify the CPU as target.
```
randyh62, line 511: This paragraph expresses that hipCpuDeviceId is a constant to specify the CPU as a target, but is it defined by the runtime? It does not seem to be established in the code example that follows?

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
```diff
@@ -0,0 +1,737 @@
...
+        return 0;
+    }
+
+Memory advice
+--------------------------------------------------------------------------------
+
+Unified memory runtime hints can be set with :cpp:func:`hipMemAdvise()` to help
+improve the performance of your code if you know the memory usage pattern. There
+are several different types of hints as specified in the enum
+:cpp:enum:`hipMemoryAdvise`, e.g. whether a certain device mostly reads the
```
randyh62:
```suggestion
:cpp:enum:`hipMemoryAdvise`, for example, whether a certain device mostly reads the
```

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
```diff
@@ -0,0 +1,737 @@
...
+        HIP_CHECK(hipFree(b));
+        HIP_CHECK(hipFree(c));
+
+        return 0;
+    }
+
+Memory range attributes
+--------------------------------------------------------------------------------
+
+:cpp:func:`hipMemRangeGetAttribute()` allows you to query attributes of a given
```
randyh62, line 665: This is a statement of what you can do, but it might be a good idea to say why you might do it or when you might do it?

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
```diff
@@ -0,0 +1,737 @@
...
+
+Asynchronously attach memory to a stream
+--------------------------------------------------------------------------------
+
+The :cpp:func:`hipStreamAttachMemAsync()` function is able to asynchronously attach
+memory to a stream, which can help concurrent execution when using streams.
+
+Currently, this function is a no-operation (NOP) function on AMD GPUs. It simply
+returns success after the runtime memory validation passed. This function is
+necessary on Microsoft Windows, and HMM is not supported on this operating
```
randyh62:
```suggestion
required on Microsoft Windows, but HMM is not supported on this operating
```

It indicates that it is necessary on Microsoft Windows, but not needed on Linux? Therefore it doesn't do anything on Linux or isn't available on Linux? I understand HMM is not available on Windows, so using it on Windows results in nothing, but HMM is supported on Linux, so using it on Linux does nothing?

docs/how-to/hip_runtime_api/memory_management/virtual_memory.rst
```diff
@@ -0,0 +1,154 @@
...
+********************************************************************************
+Virtual memory management
+********************************************************************************
+
+Memory management is important when creating high-performance applications in
+the HIP ecosystem. Both allocating and copying memory can result in bottlenecks,
+which can significantly impact performance.
+
+Global memory allocation in HIP uses the C language style allocation function.
+This works fine for simple cases but can cause problems if your memory needs
```
randyh62:
```suggestion
This works fine for simple cases but can cause problems if your application requires additional memory.
```

docs/how-to/hip_runtime_api/memory_management/virtual_memory.rst
```diff
@@ -0,0 +1,154 @@
...
+change. If you need to increase the size of your memory, you must allocate a
```
randyh62:
```suggestion
If you need to increase the size of your memory, you must allocate a
```

docs/how-to/hip_runtime_api/memory_management/virtual_memory.rst
```diff
@@ -0,0 +1,154 @@
...
+second larger buffer and copy the data to it before you can free the original
+buffer. This increases overall memory usage and causes unnecessary ``memcpy``
+calls. Another solution is to allocate a larger buffer than you initially need.
+However, this isn't an efficient way to handle resources and doesn't solve the
+issue of reallocation when the extra buffer runs out.
+
+Virtual memory management solves these memory management problems. It helps to
```
randyh62:
```suggestion
Virtual memory management solves these problems by reducing
```

docs/how-to/hip_runtime_api/memory_management/virtual_memory.rst
```diff
@@ -0,0 +1,154 @@
...
+reduce memory usage and unnecessary ``memcpy`` calls.
```
randyh62:
```suggestion
memory usage and unnecessary ``memcpy`` calls.
```

docs/how-to/hip_runtime_api/memory_management/virtual_memory.rst
```diff
@@ -0,0 +1,154 @@
...
+
+The first step is to allocate the physical memory itself with the
+:cpp:func:`hipMemCreate` function. This function accepts the size of the buffer,
+an ``unsigned long long`` variable for the flags, and a
+:cpp:struct:`hipMemAllocationProp` variable. :cpp:struct:`hipMemAllocationProp`
+contains the properties of the memory to be allocated, such as where the memory
+is physically located and what kind of shareable handles are available. If the
+allocation is successful, the function returns a value of
+:cpp:enumerator:`hipSuccess`, with :cpp:type:`hipMemGenericAllocationHandle_t`
+representing a valid physical memory allocation. The allocated memory size must
```
randyh62, line 51:
```suggestion
representing a valid physical memory allocation. 

The allocated memory size must
```

I would at least separate these last two sentences out of the preceding paragraph. I might also suggest making them a Note perhaps? Finally, I think it needs to be more definite about who needs to align the memory size with the granularity... _You_ must align the memory size...

docs/how-to/hip_runtime_api/memory_management/virtual_memory.rst
```diff
@@ -0,0 +1,154 @@
...
+    hipMemAddressReserve(&ptr, padded_size, 0, 0, 0);
+    hipMemMap(ptr, padded_size, 0, allocHandle, 0);
+
+Set memory access
+--------------------------------------------------------------------------------
+
+Finally, use the :cpp:func:`hipMemSetAccess` function to enable memory access.
+It accepts the pointer to the virtual memory, the size, and a
+:cpp:struct:`hipMemAccessDesc` descriptor as parameters. In a multi-GPU
+environment, you can map the device memory of one GPU to another. This feature
```
randyh62, line 91:
```suggestion
environment, you can map the device memory of one GPU to another. 

This feature
```
 would start a new paragraph with "This feature..."

docs/how-to/hip_runtime_api/memory_management/virtual_memory.rst
```diff
@@ -0,0 +1,154 @@
...
+and write to it, just like you would a C style memory allocation.
+
+Free virtual memory
+--------------------------------------------------------------------------------
+
+To free the memory allocated in this manner, use the corresponding free
+functions. To unmap the memory, use :cpp:func:`hipMemUnmap`. To release the
+virtual address range, use :cpp:func:`hipMemAddressFree`.  Finally, to release
+the physical memory, use :cpp:func:`hipMemRelease`. A side effect of these
+functions is the lack of synchronization when memory is released. If you call
```
randyh62, line 121: If you call :cpp:func:`hipFree` when you have multiple streams running in parallel, it
synchronizes the device. This causes worse resource usage and performance.

This seems like it could be an important note?

docs/how-to/hip_runtime_api/memory_management/virtual_memory.rst
```diff
@@ -0,0 +1,154 @@
...
+
+Memory usage
+================================================================================
+
+Dynamically increase allocation size
+--------------------------------------------------------------------------------
+
+The :cpp:func:`hipMemAddressReserve` function allows you to increase the amount
+of pre-allocated memory. This function accepts a parameter representing the
+requested starting address of the virtual memory. This allows you to have a
```
randyh62, line 141:
```suggestion
requested starting address of the virtual memory, and the amount of additional memory to reserve. This allows you to have a
```

docs/how-to/hip_runtime_api/memory_management/stream_ordered_allocator.rst
randyh62, line 1: On line 24: - Learning curve: Requires additional effort to understand and utilize SOMA effectively.

randyh62 (commented):
added comments

docs/how-to/hip_runtime_api/memory_management/device_memory.rst
MKKnorr: No, those are different programming concepts. constant and shared can not be allocated with hipmalloc, and textures also need a different api for accessing

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
MKKnorr: "can" is actually correct here - there are "devices" that share their physical memory with the host (APUs in general, but explicitly the MI300A). But those support unified memory by default

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
MKKnorr: Added as suggested

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
MKKnorr: Adapted it a bit

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
MKKnorr: Done

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
MKKnorr: Done

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
MKKnorr: There was no distinction between GPUs and accelerators in this whole chapter, and I won't start here, especially as the differences are miniscule at best

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
MKKnorr: It's a linux feature. I added a link

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
MKKnorr: There is a logical difference between "not the whole memory region needs to be transferred" and "the whole memory region does not need to be transferred"

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
MKKnorr: I wouldn't say it is provided by the "runtime", it is a constant defined in the hip headers. I think this explanation here should also be good enough

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
MKKnorr: Done as suggested

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
MKKnorr: While I generally agree with that sentiment, this does not have a specific use case. It is really just forquerying the attributes of the memory ranges, that might have been set by the functions explained in the earlier sections. I can't think of an example that would be more than just showing how to call them.

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
MKKnorr: Good point, I didn't refactor this paragraph properly. I rewrote the description of what this function does to give a better understanding, and removed the paragraph about the support completely - HIP does not support unified memory on windows at all currently, and this function has not yet been implemented in Linux, so it is just for passing it along to CUDA based systems. That however is explained in the reference to this function, so it does not need to be repeated here.

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
randyh62: What is the meaning of the sentence?

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
randyh62: You differentiate between GPUs and graphic cards?

MKKnorr: I don't think anyone would really see the terms "GPU" and "graphics card" as differentiation, they are synonyms. Changed it either way.

docs/how-to/hip_runtime_api/memory_management/unified_memory.rst
MKKnorr: Changed the sentence to make it easier to understand.

randyh62 (approved):
Looks good to me.

g-h-c:
I think many people would be caught up by the new/delete example not working, as its ability to work is not only dependent on the device attributes but also the kernel HMM support. I would add some comments to the actual code,  e.g.

`// for this example to work the environment variable HSA_XNACK needs to be set to 1, and the kernel needs HMM support`

We can potentially indicate that a way to check for HMM support is setting the environment variable, AMD_LOG_LEVEL e.g. to 3 or 4. For instance on my instinct accelerator it prints:

`:3:rocdevice.cpp            :1790: 10080897639915 us:  HMM support: 1, XNACK: 1, Direct host access: 0`

when set to 3.
But this last thing about AMD_LOG_LEVEL does not necessarily belong to this page

MKKnorr:
Thanks for the review, I agree with your points.

> I think many people would be caught up by the new/delete example not working, as its ability to work is not only dependent on the device attributes but also the kernel HMM support. I would add some comments to the actual code, e.g.

I clarified that a bit more under the system requirements, and moved the first mention of HMM support there.

> `// for this example to work the environment variable HSA_XNACK needs to be set to 1, and the kernel needs HMM support`

Added a comment similar to this to the example

> We can potentially indicate that a way to check for HMM support is setting the environment variable, AMD_LOG_LEVEL e.g. to 3 or 4. For instance on my instinct accelerator it prints:
> 
> `:3:rocdevice.cpp :1790: 10080897639915 us: HMM support: 1, XNACK: 1, Direct host access: 0`
> 
> when set to 3. But this last thing about AMD_LOG_LEVEL does not necessarily belong to this page

Agreed, this is a bit overkill, as AMD_LOG_LEVEL=3 also prints a lot of other stuff, which might make it hard to look for the HMM and XNACK support. Also similar information is listed on https://rocm.docs.amd.com/en/latest/conceptual/gpu-memory.html#xnack, which is being linked several times in this document.

PR #3631, opened 2024-10-03
Update Kernel Launch Example

harkgill-amd:
Updated the kernel launch example to fix the issues highlighted in https://github.com/ROCm/HIP/issues/3428.

Example is essentially pseudo-code and isn't expected to compile but had a few obvious errors that are corrected.

1. hipLaunchParm was causing parameter mismatch on kernel launch 
2. `__host__ ` and `__device__ ` annotations were missing on PlusOne 
3. gid indexing updated
4. Inconsistent use of gridDim and groupDim fixed
5. wrong variable names (N vs n)

### DIFF
--- a/docs/reference/cpp_language_extensions.rst
+++ b/docs/reference/cpp_language_extensions.rst
@@ -104,7 +104,7 @@ You can include your kernel arguments after these parameters.
 .. code-block:: cpp
 
   // Example hipLaunchKernelGGL pseudocode:
-  __global__ MyKernel(float *A, float *B, float *C, size_t N)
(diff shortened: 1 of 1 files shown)


docs/reference/cpp_language_extensions.rst
```diff
@@ -104,7 +104,7 @@ You can include your kernel arguments after these parameters.
 .. code-block:: cpp
```
neon60, line 104: Please also remove this line 102 from documentation:

"You can include your kernel arguments after these parameters."

neon60 (approved):

PR #3632, opened 2024-10-04
Add initialization and error handling

matyas-streamhpc:


docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,212 @@
+.. meta::
+   :description: Error Handling
+   :keywords: AMD, ROCm, HIP, error handling, error
+
+********************************************************************************
+Error handling
+********************************************************************************
+
+The Error handling API in HIP provides the necessary tools to detect, report,
+and manage errors in GPU-accelerated applications. By checking return values,
```
MKKnorr: This should emphasize, that it is only related to HIP runtime errors (and kernels). It should also explicitly mention, that HIP runtime functions return an error code.
```suggestion
HIP provides functionality to detect, report, and manage errors that occur during the execution of HIP runtime functions or when launching kernels. Every HIP runtime function, apart from launching kernels, has `hipError_t` as return type.
```

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,212 @@
...
+using functions like :cpp:func:`hipGetErrorString()`, :cpp:func:`hipGetLastError()`,
+and :cpp:func:`hipPeekAtLastError()`, and adopting best practices like defining
```
MKKnorr: This is mixing two different aspects of error handling - GetErrorString/Name are for making it human readable, the Get/PeekAtLastError are for "gathering" errors. I know the functionality of Get/PeekAtLastError is explained later, but I think it's fundamental enough to explain it in the introduction.
```suggestion
:cpp:func:`hipGetLastError()` and :cpp:func:`hipPeekAtLastError()` can be used for catching errors from kernel launches, as kernel launches don't return an error directly. HIP maintains an internal state, that includes the last error code. hipGetLastError returns and resets that error to hipSuccess, while hipPeekAtLastError just returns the error without changing it.
To get a human readable version of the errors, :cpp:func:`hipGetErrorString()` and hipGetErrorName() can be used.
```

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,212 @@
...
+error-checking macros, developers can ensure their HIP applications are robust,
+easier to debug, and more reliable. Proper error handling is crucial for
```
MKKnorr: Error handling doesn't make an application more robust or reliable by itself. There are some errors that can't be recovered from like running out of memory for example.

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,212 @@
...
+identifying issues early in the development process and ensuring that
+applications behave as expected.
+
+Strategies
+================================================================================
```
MKKnorr: This section feels a bit out of place - it is not HIP specific, and is fairly vague in general. I would vote for removing it, as it doesn't add much to the topic at hand, or at least move it to the end of the page as an "addendum"

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,212 @@
...
+
+One of the fundamental best practices in error handling is to develop a
+consistent strategy across the entire application. This involves defining
+how errors are reported, logged, and managed. This can be achieved by using a
+centralized error handling mechanism that ensures consistency and reduces
+redundancy. For instance, using macros to simplify error checking and reduce
+code duplication is a common practice. A macro like ``HIP_CHECK`` can be defined
+to check the return value of HIP API calls and handle errors appropriately.
```
MKKnorr: HIP_CHECK is actually not used for handling errors at all, it is just reporting errors, and in most implementations of it I've seen just terminates function or even program execution.

Also referring to HIP_CHECK without giving some example code might confuse users - HIP_CHECK is not a macro that is defined by the HIP runtime, it is a custom macro that the users have to define themselves.

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,212 @@
...
+applications behave as expected.
+
+Strategies
+================================================================================
+
+One of the fundamental best practices in error handling is to develop a
+consistent strategy across the entire application. This involves defining
+how errors are reported, logged, and managed. This can be achieved by using a
+centralized error handling mechanism that ensures consistency and reduces
+redundancy. For instance, using macros to simplify error checking and reduce
```
MKKnorr: This is a whole lot of buzzwords without saying much - ensuring what kind of consistency? reducing what redundancy?

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,212 @@
...
+
+It involves detecting and handling errors as early as possible to prevent them
+from propagating and causing more significant issues. Such as validating inputs
+and preconditions before performing operations.
+
+Resource management 
+--------------------------------------------------------------------------------
+
+Ensuring that resources such as memory, file handles, and network connections
+are properly managed and released in the event of an error is essential.
```
MKKnorr: This is not really a error handling strategy, it's not like this is optional. If this is not done, it can e.g. corrupt data

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,212 @@
...
+
+Error handling usage
+================================================================================
+
+Functions like :cpp:func:`hipGetLastError` and :cpp:func:`hipPeekAtLastError`
```
MKKnorr: Those are the only two functions used for that purpose, so no "like" is needed. (When ignoring the driver API versions of those functions)
```suggestion
:cpp:func:`hipGetLastError` and :cpp:func:`hipPeekAtLastError`
```

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,212 @@
...
+are used to detect errors after HIP API calls. This ensures that any issues are
+caught early in the execution flow.
```
MKKnorr: These functions do not ensure that errors are caught early - actually the opposite. They should be used for kernel launches, which don't return an error code, or asynchronous functions, which only return a subset of possible errors when called. The device-related errors of the asynchronous functions can only be caught later, as they are... well, asynchronous.

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,212 @@
...
+
+For reporting, :cpp:func:`hipGetErrorName` and :cpp:func:`hipGetErrorString`
+provide meaningful error messages that can be logged or displayed to users.
```
MKKnorr: This depends on what you understand by "logging". An internal logging mechanism doesn't need the human readable presentation of the errors, in that case it would be easier to use the error codes.

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,212 @@
...
+This helps in understanding the nature of the error and facilitates debugging.
+
+By checking for errors and providing detailed information, these functions
+enable developers to implement appropriate error handling strategies, such as
+retry mechanisms, resource cleanup, or graceful degradation.
+
+Error handling function examples
+--------------------------------------------------------------------------------
```
MKKnorr: I think these functions are all simple enough, that they don't need code examples - it basically boils down to calling the function. And e.g. hipGetErrorString has already been used twice in the code examples before its example code is introduced.

In my opinion it would be better to replace this section with the infamous HIP_CHECK macro and highlight that and its usage a bit more. After all, that is almost the first thing useres (should) learn when starting to code with HIP, it might as well be part of the "hello world" version of HIP

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,212 @@
...
+.. code-block:: cpp
+
+  std::cerr << "Error description: " << hipGetErrorString(err) << std::endl;
+
+Best practices
+--------------------------------------------------------------------------------
+
+1. Check errors after each API call
+
+  Always check the return value of HIP API calls to catch errors early. For
```
MKKnorr: The emphasis on catching errors "early" is a bit out of place - it's as if it is an option to not catch errors early, and continue execution as if nothing has happened

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,212 @@
...
+
+  .. code-block:: cpp
+
+    hipError_t err = hipMalloc(&d_A, size);
+    if (err != hipSuccess) {
+        std::cerr << "hipMalloc failed: " << hipGetErrorString(err) << std::endl;
+        return -1;
+    }
+
+2. Use macros for error checking
```
MKKnorr: As already mentioned: this should get its own section. It is a very important concept for begginners. However it should be mentioned, that this is only reporting errors, not handling them. There actually are versions of this macro, that have a flag for exiting the program after reporting the error.

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,212 @@
...
+                    << hipGetErrorString(status) \
+                    << " at " << __FILE__ << ":" \
+                    << __LINE__ << std::endl;    \
+      }                                          \
+  }
+
+  // Addition of two values.
+  __global__ void add(int *a, int *b, int *c) {
+      *c = *a + *b;
+  }
```
MKKnorr: Even if the focus here is on the error handling, I think we shouldn't encourage people to program kernels this way. With some slight changes, this could be a vector(array)-add-like kernel. Then we also wouldn't allocate memory for "scalar" variables, and wouldn't launch a kernel with 1 thread in 1 block

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,212 @@
...
+      HIP_CHECK(hipMalloc(&d_c, sizeof(*d_c)));
+
+      // Copy input values to device.
+      HIP_CHECK(hipMemcpy(d_a, &a, sizeof(*d_a), hipMemcpyHostToDevice));
+      HIP_CHECK(hipMemcpy(d_b, &b, sizeof(*d_b), hipMemcpyHostToDevice));
+
+      // Launch add() kernel on GPU.
+      hipLaunchKernelGGL(add, dim3(1), dim3(1), 0, 0, d_a, d_b, d_c);
+      // Check the kernel launch
+      HIP_CHECK(hipPeekAtLastError());
```
MKKnorr: Unless there is a reason for not resetting the error, I think hipGetLastError should be used. This way, if there was some kind of error handling, the subsequent function might fail with the same error again.
```suggestion
      HIP_CHECK(hipGetLastError());
```

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,212 @@
...
+      // Check for kernel execution error
+      HIP_CHECK(hipDeviceSynchronize());
+      
+      //Check the kernel launch
+      HIP_CHECK(hipPeekAtLastError());
```
MKKnorr: The kernel launch has already been checked
```suggestion
```

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,212 @@
...
+
+      // Copy the result back to the host.
+      HIP_CHECK(hipMemcpy(&c, d_c, sizeof(*d_c), hipMemcpyDeviceToHost));
+
+      // Cleanup allocated memory.
+      HIP_CHECK(hipFree(d_a));
+      HIP_CHECK(hipFree(d_b));
+      HIP_CHECK(hipFree(d_c));
+
+      // Prints the result.
```
MKKnorr: Being pedantic, sorry :) same style as the other comments:
```suggestion
      // Print the result.
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
+.. meta::
+   :description: Initialization.
+   :keywords: AMD, ROCm, HIP, initialization
+
+********************************************************************************
+Initialization
+********************************************************************************
+
+Initialization involves setting up the environment and resources needed for GPU
+computation. The following steps are covered with the initialization:
```
MKKnorr:
```suggestion
The initialization involves setting up the environment and resources needed for
using GPUs. The following steps are covered with the initialization:
```

Also this should mention, that initialization is done automatically by HIP, and does not *need* to be done manually, but can be done manually for more control (timing on when HIP is initialized, what GPU to use)

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+
+- Initialize the HIP runtime
```
MKKnorr: Hm. This is kind of redundant - the initialization starts with the initialization, but "setting up" also doesn't quite fit, as the function is literally called `hipInit`
```suggestion
- Setting up the HIP runtime
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+
+  This includes loading necessary libraries and setting up internal data
+  structures.
+
+- Querying and set GPU devices
```
MKKnorr: Consistent style:
```suggestion
- Querying and setting GPUs
```
And either GPUs or devices, but not GPU device

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+
+  Identifying and querying the available GPU devices on the system.
+
+- Setting up contexts
+
+  Creating contexts for each GPU device, which are essential for managing
+  resources and executing kernels. For further details, check :ref:`context
```
MKKnorr:
```suggestion
  resources and executing kernels. For further details, check the :ref:`context
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+  section <context_driver_api>`.
+
+Initialize the HIP runtime
+================================================================================
+
+The HIP runtime is initialized automatically when the first HIP API call is
+made. However, you can explicitly initialize it using :cpp:func:`hipInit`
+function, to be able to control the timing of the initialization. The explicit
```
MKKnorr: Either "the [...] function", or just by name
```suggestion
made. However, you can explicitly initialize it using :cpp:func:`hipInit`,
to be able to control the timing of the initialization. The explicit
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+.. code-block:: cpp
+
+  #include <hip/hip_runtime.h>
+  
+  hipError_t err = hipInit(0);
+  if (err != hipSuccess)
+  {
+      // Handle error
+  }
+
```
MKKnorr: This example code doesn't really add much.

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+.. note::
+
+  Include the HIP runtime header in your source file, if you want access HIP
+  functions.
```
MKKnorr: I agree, that this needs to be mentioned somewhere, however this is quite normal for any library/framework that one is using. The only reason this needs to be highlighted, is for CUDA users, because nvcc implicitly includes the corresponding header. So maybe this should rather be added to the CUDA porting guide?

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+
+The initialization includes the following steps:
```
MKKnorr: Seems like this sentence was overlooked
```suggestion
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Querying and set GPU devices
```
MKKnorr:
```suggestion
Querying and setting GPUs
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+================================================================================
+
```
MKKnorr, line 61: Short introduction of this section - if several gpus are available, and a specific one is desired, its properties can be queried, and on that information a decision for what gpu to use can be made

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+Querying GPU devices
+--------------------------------------------------------------------------------
+
+Before using a GPU device, you might want to query its properties:
```
MKKnorr: Mention the functions and link to their reference.
Something like:
The properties of a gpu can be queried using :cpp:func:hipGetDeviceProperties, which returns/provides a struct of :cpp:struct:hipDeviceProp_t, that can be used to identify a device or give an overview of hardware characteristics that might make one gpu better suited for the task at hand than others.

Also mention hipGetDeviceCount to link to its reference

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+
+.. code-block:: cpp
+
+  #include <hip/hip_runtime.h>
+  #include <iostream>
+
+  int deviceCount;
```
MKKnorr: This code does not compile out of the box, even though the includes make it seem as if it was a stand-alone file. Consider either wrapping the example code in a function or remove the includes to make it seem as if it was a block out of some other function/main

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+  if (hipGetDeviceCount(&deviceCount) == hipSuccess){
+      for (int i = 0; i < deviceCount; ++i){
+          hipDeviceProp_t prop;
+          if ( hipGetDeviceProperties(&prop, i) == hipSuccess)
+              std::cout << "Device" << i << prop.name << std::endl;
+      }
+  }
+
+Set GPU device
```
MKKnorr:
```suggestion
Setting the GPU
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+--------------------------------------------------------------------------------
+
+Select the GPU device to be used for subsequent HIP operations:
```
MKKnorr:
```suggestion
Select the GPU to be used for subsequent HIP operations:
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+
+.. code-block:: cpp
+
+  #include <hip/hip_runtime.h>
+
+  int deviceId = 0; // Example: selecting the first device
+  hipError_t err = hipSetDevice(deviceId);
```
MKKnorr: Similar to before: Either wrap in a function or remove the includes, to not make it look like it's a stand-alone file

Also: This example code is not adding much. It is just showing how to call a function

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+
+- Resource Allocation
+
+  Prepares the device for resource allocation, such as memory allocation and
+  stream creation.
+
+- Check device availablity
+
+  Checks for errors in device selection and ensures that the specified device is
+  available and capable of executing HIP operations.
```
MKKnorr: Or returns an error. The function itself does not guarantee, that the selected device is working

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Finalize
+================================================================================
+
+The :cpp:func:`hipDeviceReset()` deletes all streams created, memory allocated,
```
MKKnorr: Either "The [...] function" or just "hipDeviceReset()"
```suggestion
:cpp:func:`hipDeviceReset()` deletes all streams created, memory allocated,
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+kernels running, events created. Make sure that no other thread is using the
```
MKKnorr: What's the scope of this operation? Device-wide? Process-wide? Thread-wide?
```suggestion
kernels running and events created (by this process?). Make sure that no other thread is using the
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+device or streams, memory, kernels, events associated with the current device.
```
MKKnorr:
```suggestion
device or streams, memory, kernels or events associated with the current device.
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Any new HIP API call initialize the HIP runtime again. The example code:
```
MKKnorr:
```suggestion
Any new HIP API call initializes the HIP runtime again.
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,124 @@
...
+
+.. code-block:: cpp
+
+  #include <hip/hip_runtime.h>
+
+  hipError_t err = hipDeviceReset();
+  if (err != hipSuccess)
+  {
+      // Handle error
+  }
```
MKKnorr: I don't think this needs its own example code block

docs/how-to/hip_runtime_api/initialization.rst
neon60: We are mentioning this later.

docs/how-to/hip_runtime_api/initialization.rst
neon60: It's already mentioned in the porting guide. Removing the note.

docs/how-to/hip_runtime_api/initialization.rst
neon60: Removed.

docs/how-to/hip_runtime_api/initialization.rst
neon60: Removed.

docs/how-to/hip_runtime_api/initialization.rst
neon60: Removed.

docs/how-to/hip_runtime_api/initialization.rst
neon60: Added text.

docs/how-to/hip_runtime_api/initialization.rst
neon60: Added updated text.

docs/how-to/hip_runtime_api/initialization.rst
neon60: Update code.

docs/how-to/hip_runtime_api/initialization.rst
neon60: Updated.

docs/how-to/hip_runtime_api/error_handling.rst
neon60: Fixed.

docs/how-to/hip_runtime_api/error_handling.rst
neon60: Removed strategies.

docs/how-to/hip_runtime_api/error_handling.rst
neon60: Removed strategies.

docs/how-to/hip_runtime_api/error_handling.rst
neon60: Removed strategies.

docs/how-to/hip_runtime_api/error_handling.rst
neon60: Removed strategies.

docs/how-to/hip_runtime_api/error_handling.rst
neon60: Rewrote.

docs/how-to/hip_runtime_api/error_handling.rst
neon60: Rewrote.

docs/how-to/hip_runtime_api/error_handling.rst
neon60: I disagree with this. I do not think the 1 thread in 1 block kernel sample code is encouraging people to a bad habit.

Obviously no one should add two values at kernel side, but for showing the error handling where the kernel is not the main focus of the example code can help to understand the error handling faster.

MKKnorr: I didn't mean the performance of this kernel, or whether it even is suited for GPUs, I wouldn't even mind if it was a only the initialization of an array, even those kernels can have their justifications on GPUs, if they in turn avoid a copy from host to device.

This code doesn't provide any parallelism, even if it was launched with more threads, as it does not use threadIds etc. Expanding this to properly use several threads wouldn't expand the code by too much, it mostly would be adding a "array_size" variable, that would be added to the malloc and memcpy calls, and would add two lines overall to the kernel - getting the threads index and checking the bounds of the array. I don't think that would distract the user from the error handling in this example.

neon60: You got your size.

MKKnorr (approved):

docs/index.md
```diff
@@ -39,7 +39,9 @@ On non-AMD platforms, like NVIDIA, HIP provides header files required to support
 :::{grid-item-card} How to
 
 * {doc}`./how-to/hip_runtime_api`
```
randyh62, line 32: I would encourage you to rename the HIP Runtime API under the how-to topics. It is not a good thing to have identically named topics that can lead to confusion. The HIP Runtime API is under the Reference section. This could be called _Managing the HIP Runtime_, or _Configuring the HIP Runtime_ or something similar.

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,101 @@
...
+
+  Identifying and querying the available GPU devices on the system.
+
+- Setting up contexts
+
+  Creating contexts for each GPU device, which are essential for managing
+  resources and executing kernels. For further details, check the :ref:`context
+  section <context_driver_api>`.
+
+Initialize the HIP runtime
```
randyh62, line 30: I think this could use a bit more exposition regarding what setting up or initializing the runtime does or provides. The text mentions that it is set up automatically, but what is the advantage of setting it up manually versus auto? If there isn't much more to say than this, then is it needed in the text?

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,101 @@
...
+================================================================================
+
+The HIP runtime is initialized automatically when the first HIP API call is
+made. However, you can explicitly initialize it using :cpp:func:`hipInit`,
+to be able to control the timing of the initialization.
+
+Querying and setting GPUs
+================================================================================
+
+If multiple GPUs are available, it's worth to query and select the desired GPU(s)
```
randyh62:
```suggestion
If multiple GPUs are available in the system, you can query and select the desired GPU(s) to use
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,101 @@
...
+based on properties.
```
randyh62:
```suggestion
based on device properties, such as ... 
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,101 @@
...
+- Setting up the HIP runtime
+
+  This includes loading necessary libraries and setting up internal data
+  structures.
+
+- Querying and setting GPUs
+
+  Identifying and querying the available GPU devices on the system.
+
+- Setting up contexts
```
randyh62, line 24: This topic appears under the Setting GPUs section from the topic above. The final step of the Initialization process seems to be Finalize?

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,142 @@
...
+return an error directly. HIP maintains an internal state, that includes the
+last error code. hipGetLastError returns and resets that error to hipSuccess,
+while :cpp:func:`hipPeekAtLastError` just returns the error without changing it.
+To get a human readable version of the errors, :cpp:func:`hipGetErrorString()`
+and :cpp:func:`hipGetErrorName()` can be used.
+
+Error handling usage
+================================================================================
+
+Error handling functions enable developers to implement appropriate error
```
randyh62:
```suggestion
Error handling functions enable developers to implement appropriate strategies
```

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,142 @@
...
+handling strategies, such as retry mechanisms, resource cleanup, or graceful
```
randyh62:
```suggestion
for handling errors, such as retry mechanisms, resource cleanup, or graceful
```

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,142 @@
...
+degradation.
+
+The descriptions of the important :ref:`error handling functions <error_handling_reference>`
```
randyh62:
```suggestion
Descriptions of the important :ref:`error handling functions <error_handling_reference>`
```

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,142 @@
...
+are the following:
```
randyh62:
```suggestion
are as follows:
```

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,142 @@
+.. meta::
+   :description: Error Handling
+   :keywords: AMD, ROCm, HIP, error handling, error
+
+********************************************************************************
+Error handling
+********************************************************************************
+
+HIP provides functionality to detect, report, and manage errors that occur
```
randyh62, line 9: I think I would combine this first paragraph with the bulleted list of Important Error Handling functions below. Maybe start with the Error Handling Usage as the Intro?

randyh62 (approved):
This feels like a good start to the topic. Are there other error catching functions or datatypes that could be included here beyond the launching of kernels?

docs/how-to/hip_runtime_api/initialization.rst
neon60: I removed the finalize section, because it was not fully covered and not part of initialization. I just wanted to add a note, that after hipDeviceReset(), first API call re-initialize the HIP API.

docs/how-to/hip_runtime_api/initialization.rst
neon60: Added the use cases, when the hipInit useful and added the hipDeviceReset note here too.

docs/index.md
neon60: I can update this on a separate PR: https://github.com/ROCm/HIP/pull/3622

neon60:
> This feels like a good start to the topic. Are there other error catching functions or datatypes that could be included here beyond the launching of kernels?

The PR mentioning all the functions and return type, which are related to HIP error handling: https://rocm.docs.amd.com/projects/HIP/en/latest/doxygen/html/group___error.html

docs/how-to/hip_runtime_api/error_handling.rst
neon60: I combined the paragraphs.

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,102 @@
...
+
+  Creating contexts for each GPU device, which are essential for managing
```
randyh62, line 26: This context section needs to be imported into this initialization topic rather than simply referenced from the CUDA driver section. It is listed as the third step in initializing the runtime, yet very little information is provided regarding setting up and managing the context. The Context management API even appears to be deprecated, meaning the context must be managed through some other mechanism? What are the elements of the context? What does it provide to the runtime?

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,102 @@
...
+  resources and executing kernels. For further details, check the :ref:`context
+  section <context_driver_api>`.
+
+Initialize the HIP runtime
+================================================================================
+
+The HIP runtime is initialized automatically when the first HIP API call is
+made. However, you can explicitly initialize it using :cpp:func:`hipInit`,
+to be able to control the timing of the initialization. The manual
+initialization can be useful, when you ensure that the GPU is initialized and
```
randyh62:
```suggestion
initialization can be useful to ensure that the GPU is initialized and
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,102 @@
...
+ready or you want to isolate the GPU initialization time from other parts of
```
randyh62:
```suggestion
ready, or to isolate GPU initialization time from other parts of
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,102 @@
...
+your program.
+
+.. note::
+
+  :cpp:func:`hipDeviceReset()` deletes all streams created, memory allocated,
```
randyh62:
```suggestion
  You can use :cpp:func:`hipDeviceReset()` to delete all streams created, memory allocated, 
```

randyh62: I think this is a good note to have, but wonder if there is some reason why the user would reset the device and need to reinitialize in the program? Maybe an example use case?

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,102 @@
...
+
+If multiple GPUs are available in the system, you can query and select the
+desired GPU(s) to use based on device properties, such as size of global memory,
+size shared memory per block, support of cooperative launch and support of
+managed memory.
+
+Querying GPUs
+--------------------------------------------------------------------------------
+
+The properties of a gpu can be queried using :cpp:func:`hipGetDeviceProperties`,
```
randyh62:
```suggestion
The properties of a GPU can be queried using :cpp:func:`hipGetDeviceProperties`,
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,102 @@
...
+which returns a struct of :cpp:struct:`hipDeviceProp_t`. This struct can be
```
randyh62:
```suggestion
which returns a struct of :cpp:struct:`hipDeviceProp_t`. The properties in the struct can be
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,102 @@
...
+used to identify a device or give an overview of hardware characteristics, that
+might make one gpu better suited for the task at hand than others.
```
randyh62:
```suggestion
might make one GPU better suited for the task than others. 
```

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,102 @@
...
+  #include <iostream>
+
+  int main() {
+
+      int deviceCount;
+      if (hipGetDeviceCount(&deviceCount) == hipSuccess){
+          for (int i = 0; i < deviceCount; ++i){
+              hipDeviceProp_t prop;
+              if ( hipGetDeviceProperties(&prop, i) == hipSuccess)
+                  std::cout << "Device" << i << prop.name << std::endl;
```
randyh62, line 79: Surely I am not choosing a GPU on the basis of the device name? Maybe capture the props in an array, and then sort through them to find what you are looking for?

randyh62 (approved):
Looks good. I think it could use a few more details though.

randyh62 (approved):
Looks good.

docs/how-to/hip_runtime_api/initialization.rst
```diff
@@ -0,0 +1,102 @@
...
+Initialize the HIP runtime
+================================================================================
+
+The HIP runtime is initialized automatically when the first HIP API call is
+made. However, you can explicitly initialize it using :cpp:func:`hipInit`,
+to be able to control the timing of the initialization. The manual
+initialization can be useful to ensure that the GPU is initialized and
+ready, or to isolate GPU initialization time from other parts of
+your program.
+
```
satyanveshd, line 39: We can add a little more detail on what is being done during HIP Initialization, such as:
- Reading the environment variables set during init, and setting up the active or visible devices etc.
- Initializes the compiler as well as HSA runtime and checks any errors due to lack of resources or no active devices etc.
- Also, sets up some internal buffers which will be used for memory copies or cooperative launches.

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -0,0 +1,126 @@
...
+during the execution of HIP runtime functions or when launching kernels. Every
+HIP runtime function, apart from launching kernels, has :cpp:type:`hipError_t`
+as return type. :cpp:func:`hipGetLastError()` and :cpp:func:`hipPeekAtLastError()`
+can be used for catching errors from kernel launches, as kernel launches don't
+return an error directly. HIP maintains an internal state, that includes the
+last error code. hipGetLastError returns and resets that error to hipSuccess,
+while :cpp:func:`hipPeekAtLastError` just returns the error without changing it.
+To get a human readable version of the errors, :cpp:func:`hipGetErrorString()`
+and :cpp:func:`hipGetErrorName()` can be used.
+
```
satyanveshd, line 27: Currently, there is a slight difference in the behavior of hipGetLastError when compared with cudaGetLastError.  
- hipGetLastError returns the returned error code of just the last API even if its hipSuccess.
- cudaGetLastError returns the error returned by any of the preceeding CUDA APIs in the same host thread. 

HIP behavior will be matched with CUDA in the next major ROCm release 7.0.  Meanwhile we are planning to introduce an environment variable in 6.4, which when enabled will match the CUDA behavior, otherwise it will continue with the current behavior by default.

Please check if we can mention this difference of the API behavior in this document.

docs/how-to/hip_runtime_api/initialization.rst
neon60: Updated the description at the previous section.

docs/how-to/hip_runtime_api/error_handling.rst
neon60: Added note.

satyanveshd (approved):
LGTM.

PR #3652, opened 2024-10-29
Add SEO metadata to the index

randyh62:


neon60:
@randyh62 We're already adding SEO metadata to index.md at the following PR:

https://github.com/ROCm/HIP/pull/3656

PR #3664, opened 2024-11-07
Add multi device management

matyas-streamhpc:


docs/how-to/hip_runtime_api/multi_device.rst
```diff
@@ -0,0 +1,219 @@
...
+    int main()
+    {
+        int deviceCount;
+        hipGetDeviceCount(&deviceCount);
+        std::cout << "Number of devices: " << deviceCount << std::endl;
+        return 0;
+    }
+
+.. _multi-device_selection:
+Device selection
```
neon60, line 69: Please add new line between the ".. _multi-device_selection:" and "Device selection"

docs/how-to/hip_runtime_api/multi_device.rst
```diff
@@ -0,0 +1,219 @@
...
+
+.. code-block:: cpp
+
+    #include <hip/hip_runtime.h>
+    #include <iostream>
+
+    int main()
+    {
+        int deviceCount;
+        hipGetDeviceCount(&deviceCount);
```
neon60, line 37: Loop over the devices and print out critical device properties, something like this:

https://github.com/ROCm/rocm-examples/blob/develop/HIP-Basic/device_query/main.cpp

neon60: Maybe reuse the example code? Maybe too much.

docs/how-to/hip_runtime_api/multi_device.rst
```diff
@@ -0,0 +1,219 @@
...
+        int deviceId0 = 0;
+        int deviceId1 = 1;
+
+        // Set device 0 and perform operations
+        hipSetDevice(deviceId0); // Set device 0 as current
+        hipMalloc(&deviceData0, size); // Allocate memory on device 0
+        simpleKernel<<<1000, 128>>>(deviceData0); // Launch kernel on device 0
+
+        // Set device 1 and perform operations
+        hipSetDevice(deviceId1); // Set device 1 as current
```
neon60: Use HIPCHECK macros

docs/how-to/hip_runtime_api/multi_device.rst
```diff
@@ -0,0 +1,219 @@
...
+        hipStreamSynchronize(stream);
+        hipStreamDestroy(stream);
+        hipFree(deviceData);
+
+        return 0;
+    }
+
+Peer-to-peer memory access
+==========================
+In multi-GPU systems, peer-to-peer memory access enables one GPU to directly read or write to the memory of another GPU. This capability reduces data transfer times by allowing GPUs to communicate directly without involving the host. Enabling peer-to-peer access can significantly improve the performance of applications that require frequent data exchange between GPUs, as it eliminates the need to transfer data through the host memory.
```
neon60: Add new line after section title

docs/how-to/hip_runtime_api/multi_device.rst
```diff
@@ -0,0 +1,219 @@
...
+
+.. _multi-device:
+
+*******************************************************************************
+Multi-device management
+*******************************************************************************
+
+Device enumeration
+==================
+Device enumeration involves identifying all the available GPUs connected to the host system. A single host machine can have multiple GPUs, each with its own unique identifier. By listing these devices, you can decide which GPU to use for computation. This process helps in managing resources efficiently and optimizing performance. The host queries the system to count and list all connected GPUs, ensuring that the application can leverage the full computational power available.
```
neon60: Remove this sentence:

"This process helps in managing resources efficiently and optimizing performance. "

matyas-streamhpc: Removed.

docs/how-to/hip_runtime_api/multi_device.rst
matyas-streamhpc: Added

docs/how-to/hip_runtime_api/multi_device.rst
matyas-streamhpc: Added

docs/how-to/hip_runtime_api/multi_device.rst
matyas-streamhpc: I added something similar as in the example code.

docs/how-to/hip_runtime_api/multi_device.rst
matyas-streamhpc: Added

neon60 (approved):

docs/how-to/hip_runtime_api/multi_device.rst
```diff
@@ -0,0 +1,287 @@
...
+*******************************************************************************
+Multi-device management
+*******************************************************************************
+
+Device enumeration
+===============================================================================
+
+Device enumeration involves identifying all the available GPUs connected to the
+host system. A single host machine can have multiple GPUs, each with its own 
+unique identifier. By listing these devices, you can decide which GPU to use
```
randyh62, line 16:
```suggestion
unique identifier. By listing these devices, you can decide which GPU to use
```

I feel like we need to comment on the fact that the runtime will choose a suitable GPU from the available GPUs if you do not assign one in the application. Is there a method whereby the runtime assigns one automatically? Doe sit choose based on some understanding of the code and the available resources, or does it pick one that is available?

docs/how-to/hip_runtime_api/multi_device.rst
```diff
@@ -0,0 +1,287 @@
...
+for computation. The host queries the system to count and list all connected
+GPUs, ensuring that the application can leverage the full computational power
```
randyh62: Is the recommendation to have an application that lists the devices and their properties so that for a given machine you can understand how to deploy the application in the future? Or is the host code intended to evaluate the code on some basis and make a selection during runtime? I am just trying to understand the recommendation to the user?

docs/how-to/hip_runtime_api/multi_device.rst
```diff
@@ -0,0 +1,287 @@
...
+        simpleKernel<<<1000, 128>>>(deviceData0); // Launch kernel on device 0
+        HIP_CHECK(hipDeviceSynchronize());
+
+        // Set device 1 and perform operations
+        HIP_CHECK(hipSetDevice(deviceId1)); // Set device 1 as current
+        HIP_CHECK(hipMalloc(&deviceData1, size)); // Allocate memory on device 1
+        simpleKernel<<<1000, 128>>>(deviceData1); // Launch kernel on device 1
+        HIP_CHECK(hipDeviceSynchronize());
+
+        // Attempt to use deviceData0 on device 1 (This will not work as deviceData0 is allocated on device 0)
```
randyh62: Is there some shared memory or global memory that would allow device 1 to access data from device 0? Or perhaps some mechanism to copy data from device 0 to device 1? Is this worth mentioning here?

docs/how-to/hip_runtime_api/multi_device.rst
```diff
@@ -0,0 +1,287 @@
...
+
+
+Stream and event behavior
+===============================================================================
+
+In a multi-device system, streams and events are essential for efficient
+parallel computation and synchronization. Streams enable asynchronous task
+execution, allowing multiple devices to process data concurrently without
+blocking one another. Events provide a mechanism for synchronizing operations
+between devices, ensuring that tasks on one device are completed before the
```
randyh62: It seems like events would ensure that kernel tasks are completed, regardless of what device they run on, before dependent tasks  (or downstream tasks) begin...

This seems like a subtle difference, but it is about the tasks being managed across multiple devices, isn't it? I don't know...

docs/how-to/hip_runtime_api/multi_device.rst
```diff
@@ -0,0 +1,287 @@
...
+        hipMalloc(&deviceData, size);
+
+        // Record the start event
+        hipEventRecord(startEvent, stream);
+
+        // Launch the kernel asynchronously
+        simpleKernel<<<1000, 128, 0, stream>>>(deviceData);
+
+        // Record the stop event
+        hipEventRecord(stopEvent, stream);
```
randyh62: Is this example showing what the paragraph above describes? It seems like it is simply a kernel running on a device, and some wait state before capturing performance data?

docs/how-to/hip_runtime_api/multi_device.rst
randyh62: Maybe mention the Peer to Peer section below?

randyh62 (approved):
Left some comments. Looks good overall.

docs/how-to/hip_runtime_api/multi_device.rst
matyas-streamhpc: Thanks, you are right! I updated the code.

docs/how-to/hip_runtime_api/multi_device.rst
matyas-streamhpc: Yes, you are right. I rephased. How does it sound now?

docs/how-to/hip_runtime_api/multi_device.rst
matyas-streamhpc: Yes, you are right. I rephased. How does it sound now?

docs/how-to/hip_runtime_api/multi_device.rst
matyas-streamhpc: Nice catch!
Typically during deployment planning. I added a sentence to clarify it.

docs/how-to/hip_runtime_api/multi_device.rst
matyas-streamhpc: New paragraph added.

docs/how-to/hip_runtime_api/multi_device.rst
```diff
@@ -0,0 +1,423 @@
...
+    }
+
+
+Stream and event behavior
+===============================================================================
+
+In a multi-device system, streams and events are essential for efficient
+parallel computation and synchronization. Streams enable asynchronous task
+execution, allowing multiple devices to process data concurrently without
+blocking one another. Events provide a mechanism for synchronizing operations
```
randyh62, line 151: I think we should indicate that events allow synchronization across streams and across devices.

docs/how-to/hip_runtime_api/multi_device.rst
```diff
@@ -0,0 +1,423 @@
...
+        hipSetDevice(0);
+        hipEventRecord(startEvent0, stream0);
+
+        // Launch the kernel asynchronously on device 0
+        simpleKernel<<<1000, 128, 0, stream0>>>(deviceData0);
+
+        // Record the stop event on device 0
+        hipEventRecord(stopEvent0, stream0);
+
+        // Wait for the stop event on device 0 to complete
```
randyh62, line 210: Is this a non-blocking operation? Or does the host code stop executing at this point to wait for stopEvent0? What lines in the code does this affect? Where does the host code stop to wait for the stopEvent to return?

docs/how-to/hip_runtime_api/multi_device.rst
```diff
@@ -0,0 +1,423 @@
...
+        hipSetDevice(1);
+        hipEventRecord(startEvent1, stream1);
+
+        // Launch the kernel asynchronously on device 1
+        simpleKernel<<<1000, 128, 0, stream1>>>(deviceData1);
+
+        // Record the stop event on device 1
+        hipEventRecord(stopEvent1, stream1);
+
+        // Wait for the stop event on device 1 to complete
```
randyh62, line 223: Is this non-blocking?

docs/how-to/hip_runtime_api/multi_device.rst
```diff
@@ -0,0 +1,423 @@
...
+        hipEventSynchronize(stopEvent1);
+
+        // Calculate elapsed time between the events for both devices
+        float milliseconds0 = 0, milliseconds1 = 0;
+        hipEventElapsedTime(&milliseconds0, startEvent0, stopEvent0);
```
randyh62, line 228: Presumably this is where the host code waits?

docs/how-to/hip_runtime_api/multi_device.rst
```diff
@@ -0,0 +1,423 @@
...
+                simpleKernel<<<1000, 128>>>(deviceData0); // Launch kernel on device 0
+                HIP_CHECK(hipDeviceSynchronize());
+
+                // Set device 1 and perform operations
+                HIP_CHECK(hipSetDevice(deviceId1)); // Set device 1 as current
+                HIP_CHECK(hipMalloc(&deviceData1, size)); // Allocate memory on device 1
+                simpleKernel<<<1000, 128>>>(deviceData1); // Launch kernel on device 1
+                HIP_CHECK(hipDeviceSynchronize());
+
+                // Attempt to use deviceData0 on device 1 (This will not work as deviceData0 is allocated on device 0)
```
randyh62, line 397: I like this demonstration of error checking, and highlighting the fact that P2P is not enabled. Nice.

docs/how-to/hip_runtime_api/multi_device.rst
```diff
@@ -0,0 +1,423 @@
...
+            {
+                double* deviceData0;
+                double* deviceData1;
+                size_t  size = 1024 * sizeof(*deviceData0);
+
+                int deviceId0 = 0;
+                int deviceId1 = 1;
+
+                // Enable peer access
+                hipSetDevice(deviceId0);
```
randyh62, line 305: I would note in the comment that the device must be the active device before enabling peer access to another device, and maybe that this provides access to all allocated and future allocated memory?

randyh62 (approved):
Added a few comments

docs/how-to/hip_runtime_api/multi_device.rst
matyas-streamhpc: Thank you!

docs/how-to/hip_runtime_api/multi_device.rst
matyas-streamhpc: hipEventElapsedTime is a non-blocking operation, so we need synchronization beforehand to ensure accurate timing measurements.

docs/how-to/hip_runtime_api/multi_device.rst
matyas-streamhpc: Yes.

docs/how-to/hip_runtime_api/multi_device.rst
matyas-streamhpc: hipEventRecord is non-blocking. Without synchronization, it would not measure the event time well, because the two events could overlap and interfere.

docs/how-to/hip_runtime_api/multi_device.rst
matyas-streamhpc: Thank! It's a good idea. I extended the comment.

docs/how-to/hip_runtime_api/multi_device.rst
matyas-streamhpc: Thanks! Updated.

randyh62 (approved):
Looks good to me

g-h-c:
The device enumeration section is perhaps slightly imprecise when it comes to systems that have both AMD and Nvidia GPUs, because the HIP runtime cannot see both at the same time, yet the page says: "The host queries the system to count and list all connected GPUs,"
A binary using HIP compiled with `HIP_PLATFORM=amd` will only see amd devices, and with `HIP_PLATFORM=nvidia` only Nvidia; or at least that is what I experience when I run it on my hybrid system.

matyas-streamhpc:
> The device enumeration section is perhaps slightly imprecise when it comes to systems that have both AMD and Nvidia GPUs, because the HIP runtime cannot see both at the same time, yet the page says: "The host queries the system to count and list all connected GPUs," A binary using HIP compiled with `HIP_PLATFORM=amd` will only see amd devices, and with `HIP_PLATFORM=nvidia` only Nvidia; or at least that is what I experience when I run it on my hybrid system.

Thank you, it's a nice catch! Updated.

PR #3730, opened 2025-01-28
Docs: Consolidate Environment Variables

adeljo-amd:
The goal of this PR is to consolidate the different environment variable tables into a single RST file, to make it easier to reference later on. A custom directive is used to select a table from the file, though it is possible to include that RST file to print out all tables contained.


neon60 (approved):

neon60:
LGTM

neon60:
There is no content change here, that's why I am merging in.

PR #3736, opened 2025-01-31
added compilation cache content

randyh62:
Had to recreate PR due to branch issues

### DIFF
--- a/docs/how-to/hip_rtc.md
+++ b/docs/how-to/hip_rtc.md
@@ -9,11 +9,13 @@
 HIP lets you compile kernels at runtime with the `hiprtc*` APIs.
 Kernels can be stored as a text string and can be passed to HIPRTC APIs alongside options to guide the compilation.
 
-NOTE:
+:::{note}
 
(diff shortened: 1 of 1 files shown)


docs/how-to/hip_rtc.md
```diff
@@ -230,6 +232,22 @@ int main() {
 }
 ```
 
+## Kernel Compilation Cache
+
+HIPRTC incorporates a cache to avoid recompiling kernels between program executions. The contents of the cache include the kernel source code (including the contents of any `#include` headers), the compilation flags, and the compiler version. After a ROCm version update, the kernels are progressively recompiled, and the new results are cached. When the cache is disabled, each kernel is recompiled every time it is requested.
+
+Use the following environment variables to manage the cache status as enabled or disabled, the location for storing the cache contents, and the cache eviction policy:
+
+* `AMD_COMGR_CACHE` By default this variable has a value of `0` and the compilation cache feature is disabled. To enable the feature set the environment variable to a value of `1` (or any value other than `0`). This behavior may change in a future release.
```
neon60, line 241: I mentioned to @adeljo-amd, to also add these variables to environment variables page on HIP.

neon60 (approved):

PR #3753, opened 2025-02-25
Docs: Add page about HIP error codes

adeljo-amd:
Add a page that catalogues all HIP error codes

### DIFF
--- a/docs/how-to/hip_runtime_api/error_handling.rst
+++ b/docs/how-to/hip_runtime_api/error_handling.rst
@@ -37,6 +37,8 @@ Best practices of HIP error handling:
 For more details on the error handling functions, see :ref:`error handling
 functions reference page <error_handling_reference>`.
 
+For a list of all error codes, see :ref:`HIP error codes <hip_error_codes>`.
+
 .. _hip_check_macros:
 
 HIP check macros
--- a/docs/reference/error_codes.rst
+++ b/docs/reference/error_codes.rst
@@ -1,0 +1,1283 @@
+.. meta::
+    :description: HIP error codes reference
+    :keywords: AMD, HIP, error codes, error, reference
+
+.. _hip_error_codes:
+
+********************************************************************************
+HIP error codes
+********************************************************************************
+
+This page lists all HIP runtime error codes and their descriptions. These error codes are
+returned by HIP API functions to indicate various runtime conditions and errors.
+
+For more details, see :ref:`Error handling functions <error_handling_reference>`.
+
+Basic Runtime Errors
+====================
+
+.. list-table::
+    :header-rows: 1
+    :widths: 30 10 60
+
+    * - Error Code
+      - Value
+      - Description
+
+    * - :term:`hipSuccess`
+      - ``0``
+      - No error
+
+    * - :term:`hipErrorUnknown`
+      - ``999``
+      - Unknown error
+
+    * - :term:`hipErrorNotReady`
+      - ``600``
+      - Device not ready
+
+    * - :term:`hipErrorIllegalState`
+      - ``401``
+      - The operation cannot be performed in the present state
+
+    * - :term:`hipErrorNotSupported`
+      - ``801``
+      - Operation not supported
+
+    * - :term:`hipErrorTbd`
+      - ``1054``
+      - To be determined/implemented
+
+.. glossary::
+
+  hipSuccess
+
+    No error. Operation completed successfully. This is returned when a HIP function completes
+    without any errors and indicates normal execution.
+
+  hipErrorUnknown
+
+    Unknown error. This is a general error code returned when no other error code is applicable or
+    when the specific error condition cannot be determined. This may indicate an unexpected
+    internal error in the HIP runtime or driver.
+
+  hipErrorNotReady
+
+    Device not ready. This error occurs when asynchronous operations have not completed.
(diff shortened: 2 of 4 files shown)


docs/reference/error_codes.rst
```diff
@@ -0,0 +1,1281 @@
...
+    When this error occurs, the affected memory contents are unreliable and the operation cannot
+    continue safely. This error generally requires system intervention, and in persistent cases,
+    may indicate hardware that needs replacement.
+
+  **hipErrorUnsupportedLimit**
+
+    Limit is not supported on this architecture. This error occurs when attempting to query or
+    set a device limit that is not supported by the current hardware. Common scenarios include:
+
+    * Using ``hipDeviceSetLimit()`` with a limit type not supported by the hardware
```
neon60:
```suggestion
    * Using :cpp:func:`hipDeviceSetLimit` with a limit type not supported by the hardware
```

neon60:
Please check all the metnioned functions and use cpp:func, where you can.

docs/reference/error_codes.rst
adeljo-amd: Done

adeljo-amd:
@neon60 Luckily all functions can be referenced, done.

docs/how-to/hip_runtime_api/error_handling.rst
```diff
@@ -37,6 +37,8 @@ Best practices of HIP error handling:
 For more details on the error handling functions, see :ref:`error handling
 functions reference page <error_handling_reference>`.
```
randyh62, line 38:
```suggestion
For more details, see :ref:`Error handling functions <error_handling_reference>`.
```

randyh62 (approved):
Nice job.

docs/reference/error_codes.rst
```diff
@@ -0,0 +1,1280 @@
...
+
+    No ROCm-capable device is detected. This error occurs when the system does not have any compatible GPU devices
+    that support the HIP runtime. Common scenarios include:
+
+    * No physical GPU is installed in the system
+    * Installed GPUs are not supported by the current HIP/ROCm version
+    * GPU drivers are missing, outdated, or corrupted
+    * GPU hardware failure or disconnection
+    * System configuration prevents GPU detection (e.g., BIOS settings, virtualization limitations)
+
```
g-h-c, line 304: On Linux, when HIP_PLATFORM=amd, if the user does not belong to both `render` and `video` you will get `hipErrorNoDevice` in HIP calls.

I think it is worth mentioning as many users will find the error code because of it.

docs/how-to/hip_runtime_api/error_handling.rst
adeljo-amd: Done

docs/reference/error_codes.rst
adeljo-amd: I've added a line, any other cases for other errors that I might have overlooked?

g-h-c: No, I think those are the main sources of the error that people might encounter

neon60 (approved):

docs/reference/error_codes.rst
g-h-c: Perhaps is worth mentioning that you might get `hipErrorInvalidDeviceFunction` if you compile using --offload-arch for a specific gfx, but then I run it on a different GPU

i.e. basically when you compile for a specific target, but then run it on a different, incompatible one

The explanation for `hipErrorInvalidDeviceFunction` already says something like that but perhaps it could be more specific

adeljo-amd: Added it as a common scenario. Anything else I should add to this error or other errors?

g-h-c: No, nothing to add from my side. Thank you.

adeljo-amd:
@g-h-c Since you're already here, do you think this is ready to be merged in?

(I'll fix the new conflicts in the meantime)

docs/reference/error_codes.rst
```diff
@@ -0,0 +1,1283 @@
...
+    Device kernel image is invalid. This error occurs when the device code image is corrupted or in an
+    unsupported format.
+
+  hipErrorLaunchFailure
+
+    Unspecified launch failure. This is a general error that occurs when a kernel launch fails.
+    Common causes include:
+
+    * Stack overflow in device code
```
g-h-c: I only have one more comment: why was "stack overflow" added as a potential cause for hipErrorLaunchFailure? On the instinct accelerator it does not cause that error code, it aborts with an HSA exception, e.g.
HSA_STATUS_ERROR_ILLEGAL_INSTRUCTION

And on NVIDIA, I don't have one to try it right now, but I believe it does not return hipErrorLaunchFailure, it segfaults.

Just checking where that information comes from.

adeljo-amd: You're right - I mixed up some of my sources and put the wrong cause there. Thanks for spotting that! I'll remove that line. Did you spot any other causes I might have misplaced?

adeljo-amd: I replaced it with "Mismatch between block size configuration and block size specified in launch bounds parameter"

g-h-c: No, LGTM

PR #3776, opened 2025-04-10
Fix CU and WGP mode

neon60:
There is no effect on warpSize

### DIFF
--- a/docs/reference/hardware_features.rst
+++ b/docs/reference/hardware_features.rst
@@ -240,10 +240,12 @@ page.
         - 106
         - 104
 
-.. [1] RDNA architectures have a configurable wavefront size. The native
-   wavefront size is 32, but they can run in "CU mode", which has an effective
-   wavefront size of 64. This affects the number of resident wavefronts and
-   blocks per compute Unit.
+.. [1] The RDNA architectures feature an experimental compiler option called 
+   ``mwavefrontsize64``, which determines the wavefront size for kernel code
+   generation. When this option is disabled, the native wavefront size of 32 is
+   used, when enabled wavefront size 64 is used. This option is not supported by
+   the HIP runtime.
+
 .. [2] RDNA architectures expand the concept of the traditional compute unit
    with the so-called work group processor, which effectively includes two
    compute units, within which all threads can cooperate.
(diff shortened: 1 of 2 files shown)


docs/reference/hardware_features.rst
```diff
@@ -240,10 +240,11 @@ page.
         - 106
         - 104
 
-.. [1] RDNA architectures have a configurable wavefront size. The native
-   wavefront size is 32, but they can run in "CU mode", which has an effective
-   wavefront size of 64. This affects the number of resident wavefronts and
-   blocks per compute Unit.
+.. [1] The RDNA architectures feature an experimental compiler option called
```
adeljo-amd, line 243: Did you confirm whether this is the case? I remember some chatter (I haven't confirmed it myself) that wavefront size of 64 doesn't really work in HIP for RDNA archs

docs/reference/hardware_features.rst
```diff
@@ -240,10 +240,11 @@ page.
         - 106
         - 104
 
-.. [1] RDNA architectures have a configurable wavefront size. The native
-   wavefront size is 32, but they can run in "CU mode", which has an effective
```
adeljo-amd, line 244: It would still be better to mention CU mode vs WGP mode I think, but using the proper wording (from the hipRTC page)

adeljo-amd:
LGTM

docs/reference/hardware_features.rst
g-h-c: The HIP runtime does not support it, it would assert on RDNA (i.e. gfx10 and above):

https://github.com/AMD-ROCm-Internal/hip/blob/cbfec76ea8354ba67840a47972942eec1c86777f/include/hip/hip_runtime.h#L40

The compiler does support it (and this, I tried in the past, the compiler does accept -mwavefront64 and the warpSize becomes 64 at runtime effectively)

So, to clarify, there is hardware and compiler support for wavefront64. But not HIP runtime support.

neon60: Thanks. Updated the PR, based on the feedback.

g-h-c (approved):

PR #3782, opened 2025-04-29
Added not about HIP runtime build on windows

neon60:


docs/install/build.rst
```diff
@@ -176,6 +176,11 @@ Set the repository branch using the variable: ``ROCM_BRANCH``. For example, for
             make -j$(nproc)
             sudo make install
 
+.. note::
+
+   HIP runtime is not buildable on Windows as it depends on closed source
```
jujiang-del: For this sentence "HIP runtime is not buildable on Windows as it depends on closed source components."
It's better like this,
The instructions are applicable on Linux only.

benvanik:
(note, Windows support is coming https://github.com/ROCm/TheRock/issues/36)

docs/install/build.rst
neon60: @jujiang-del, you are right. I should avoid negative statements, but I have to highlight that this is currently not buildable on Windows, so I went with the "The HIP runtime is only buildable on Linux."

Yes, the instructions are Linux-specific, but we have to highlight that users should not try to build on Windows.

neon60:
> (note, Windows support is coming [ROCm/TheRock#36](https://github.com/ROCm/TheRock/issues/36))

Thanks for the comment. We will add build commands, when it's available.

docs/install/build.rst
jujiang-del: Looks good. Thanks.

jujiang-del:
Looks good.

jujiang-del (approved):

ScottTodd:
> > (note, Windows support is coming [ROCm/TheRock#36](https://github.com/ROCm/TheRock/issues/36))
> 
> Thanks for the comment. We will add build commands, when it's available.

We have HIP building on Windows in (partial) open source, with some caveats. Notes are at https://github.com/ROCm/TheRock/blob/main/docs/development/windows_support.md#building-clr-from-partial-sources. I haven't tested with HIP / CLR standalone (outside of TheRock), but I suspect it should work with https://github.com/ROCm/clr/commit/a3effa16f15b0981eea2865b6d179babd9121111 and https://github.com/nod-ai/amdgpu-windows-interop. A more permanent home for that interop project or more source files (instead of precompiled .lib files) is further out.

PR #3786, opened 2025-05-09
Environment variable update

neon60:


docs/data/env_variables_hip.rst
```diff
@@ -5,32 +5,38 @@
 HIP GPU isolation variables
 --------------------------------------------------------------------------------
 
-The GPU isolation environment variables in HIP are collected in the following table.
+The GPU isolation environment variables in HIP are collected in the following
```
randyh62: I think this should be a brief introduction or explanation of what a GPU isolation variable is for.

docs/data/env_variables_hip.rst
```diff
@@ -5,32 +5,38 @@
...
-    :widths: 70,30
+    :widths: 50,30,20
 
     * - **Environment variable**
+      - **Links**
       - **Value**
 
     * - | ``ROCR_VISIBLE_DEVICES``
         | A list of device indices or UUIDs that will be exposed to applications.
+      - :doc:`GPU isolation <rocm:conceptual/gpu-isolation>`, :doc:`Setting the number of compute units <rocm:how-to/setting-cus>`, :doc:`LLM inference frameworks <rocm:how-to/rocm-for-ai/inference/llm-inference-frameworks>`
```
randyh62: The _LLM inference frameworks_ link seems unnecessary.

docs/data/env_variables_hip.rst
```diff
@@ -5,32 +5,38 @@
...
+table.
 
 .. _hip-env-isolation:
 .. list-table::
     :header-rows: 1
-    :widths: 70,30
+    :widths: 50,30,20
 
     * - **Environment variable**
+      - **Links**
```
randyh62: I see the other tables don't have the links column, and some of the links are repeated throughout the table. Perhaps the links could be added in an Introduction to GPU isolation and removed from the table?

docs/data/env_variables_hip.rst
```diff
@@ -5,32 +5,38 @@
...
     * - | ``HIP_VISIBLE_DEVICES`` or ``CUDA_VISIBLE_DEVICES``
         | Device indices exposed to HIP applications.
+      - :doc:`GPU isolation <rocm:conceptual/gpu-isolation>`, :doc:`HIP debugging <hip:how-to/debugging>`
       - Example: ``0,2``
 
 HIP profiling variables
 --------------------------------------------------------------------------------
 
-The profiling environment variables in HIP are collected in the following table.
+The profiling environment variables in HIP are collected in the following table. For
```
randyh62: It is interesting that there is no mention of the profiling variables in the rocProf documentation. There is a **cu_mask** variable that is probably related, but no mention of the environment variables. I think the _Setting the number of CUs_ topic is good, as it provides documentation for these variables, but it would be nice to see something in profiling as well?

docs/data/env_variables_hip.rst
```diff
@@ -158,11 +163,19 @@ The debugging environment variables in HIP are collected in the following table.
...
+
 HIP memory management related variables
 --------------------------------------------------------------------------------
 
 The memory management related environment variables in HIP are collected in the
-following table.
+following table. The ``HIP_HOST_COHERENT`` variable linked at the following
+pages:
+
+- :ref:`Coherence control <hip:hip-memory-coherence-table>`
```
randyh62: This link does not appear to work as written? 
Should it point to how-to/hip_runtime_api/memory_management/coherence_control.html

randyh62 (approved):
Left a few comments.

docs/data/env_variables_hip.rst
neon60: Added the ref link with PR. Link will/should work, after we merged in the PR.

docs/data/env_variables_hip.rst
neon60: Added.

docs/data/env_variables_hip.rst
neon60: This is the preferred way. We add links column, when we have multiple links per variables. We just mention the page, if all the variables are appearing on the same how-to page.

docs/data/env_variables_hip.rst
neon60: HSA_CU_MASK is ROCr runtime variable, maybe we need to move this there.

We should further expand the how-to use env variable pages.

docs/data/env_variables_hip.rst
neon60: Removed.

docs/data/env_variables_hip.rst
randyh62: This still seems like an issue in the latest update. I think it needs to be clarified for the user.

docs/data/reference/env_variables/debug_hip_env.rst
```diff
@@ -0,0 +1,103 @@
...
+        | 4: Enables debug logs next to lower-level logs.
+        | 5: Enables debug extra logs next to lower-level logs.
+
+    * - | ``AMD_LOG_LEVEL_FILE``
+        | Sets output file for ``AMD_LOG_LEVEL``.
+      - stderr output
+      -
+
+    * - | ``AMD_LOG_MASK``
+        | Specifies HIP log filters. Here is the ` complete list of log masks <https://github.com/ROCm/clr/blob/develop/rocclr/utils/debug.hpp#L40>`_.
```
randyh62, line 30: Is there some way to turn on multiple LOG masks? What is the syntax for multiple values?

randyh62: Also, just to be clear, enabling the LOG_MASK enables logging for that feature? It seems strange to use a MASK to enable something. Maybe that needs to be clarified?

randyh62 (approved):

g-h-c:
AMD_COMGR_SAVE_TEMPS, AMD_COMGR_EMIT_VERBOSE_LOGS and AMD_COMGR_REDIRECT_LOGS are useful as well, when debugging hiprtc programs. It might be worthy to add them too. 

e.g. to provide more logging for a hiprtc program you could do:

`AMD_COMGR_SAVE_TEMPS=1 AMD_COMGR_EMIT_VERBOSE_LOGS=1 AMD_COMGR_REDIRECT_LOGS=stdout ./myrtcprogram`

and it will show you exactly what the compiler is trying to compile at runtime

docs/data/reference/env_variables/miscellaneous_hip_env.rst
```diff
@@ -0,0 +1,34 @@
...
+      - **Default value**
+      - **Value**
+
+    * - | ``HIPRTC_COMPILE_OPTIONS_APPEND``
+        | Sets compile options needed for ``hiprtc`` compilation.
+      - Unset by default.
+      - ``--gpu-architecture=gfx906:sramecc+:xnack``, ``-fgpu-rdc``
+
+    * - | ``AMD_COMGR_SAVE_TEMPS``
+        | Controls the delete temporary files generated during compilation of COMGR. These files do not appear in the current working directory, but are instead left in a platform-specific temporary directory.
```
g-h-c: I think 'deletion' sounds better than 'delete' in this case.

docs/data/reference/env_variables/miscellaneous_hip_env.rst
```diff
@@ -0,0 +1,34 @@
...
+      - Unset by default.
+      - | 0: Temporary files deleted automatically.
+        | Non zero integer: Turn off temporary files delete.
```
g-h-c: Same, 'deletion'

neon60: Fixed.

docs/data/reference/env_variables/miscellaneous_hip_env.rst
neon60: Fixed.

g-h-c (approved):

PR #3817, opened 2025-07-14
Docs: Update low_fp_types

adeljo-amd:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update
- [ ] Continuous Integration

## What were the changes?

- Add CDNA4 and RDNA4 to the tables
- Correct support for some types and CDNA3
- Split float16 and bfloat16 into 2 sections

## Why are these changes needed?

Updating the page to be up-to-date.

## Updated CHANGELOG?

- [ ] Yes
- [x] No, Does not apply to this PR.

## Added/Updated documentation?

- [x] Yes
- [ ] No, Does not apply to this PR.

## Additional Checks

- [ ] I have added tests relevant to the introduced functionality, and the unit tests are passing locally.
- [ ] Any dependent changes have been merged.

### DIFF
--- a/.wordlist.txt
+++ b/.wordlist.txt
@@ -12,6 +12,7 @@ Asynchrony
 asynchrony
 backtrace
 bfloat
+BFloat
 Bitcode
 bitcode
 bitcodes
@@ -145,6 +146,7 @@ PyHIP
 queryable
 prefetching
 quad
+README
 representable
 RMW
 rocgdb
--- a/docs/reference/low_fp_types.rst
+++ b/docs/reference/low_fp_types.rst
@@ -43,27 +43,31 @@ HIP Header
 The `HIP FP4 header <https://github.com/ROCm/clr/blob/amd-staging/hipamd/include/hip/amd_detail/amd_hip_fp4.h>`_
 defines the FP4 numbers.
 
-Supported Devices
------------------
+Device Compatibility
+--------------------
 
-Different GPU models support different FP4 formats. Here's a breakdown:
+The following table shows hardware support for this precision format by GPU architecture. "Yes"
+indicates native hardware acceleration is available, while "No" indicates hardware acceleration
+is not available.
 
-.. list-table:: Supported devices for fp4 numbers
+.. list-table::
     :header-rows: 1
 
     * - Device Type
       - E2M1
-    * - Host
-      - Yes
     * - CDNA1
       - No
     * - CDNA2
       - No
     * - CDNA3
+      - No
+    * - CDNA4
       - Yes
     * - RDNA2
       - No
     * - RDNA3
+      - No
+    * - RDNA4
       - No
 
 Using FP4 Numbers in HIP Programs
@@ -130,9 +134,9 @@ The following code example demonstrates a simple roundtrip conversion using FP4 
         constexpr size_t size = 16;
         hipDeviceProp_t prop;
         hip_check(hipGetDeviceProperties(&prop, 0));
-        bool is_supported = (std::string(prop.gcnArchName).find("gfx94") != std::string::npos);
+        bool is_supported = (std::string(prop.gcnArchName).find("gfx950") != std::string::npos);
         if(!is_supported) {
-            std::cerr << "Need a gfx94x, but found: " << prop.gcnArchName << std::endl;
+            std::cerr << "Need gfx950, but found: " << prop.gcnArchName << std::endl;
             std::cerr << "Device conversions are not supported on this hardware." << std::endl;
             return -1;
         }
@@ -180,12 +184,12 @@ There are C++ style classes available as well:
 
 FP4 type has its own class:
 
-- __hip_fp4_e2m1
+- ``__hip_fp4_e2m1``
 
 There is support of vector of FP4 types:
 
-- __hip_fp4x2_e2m1: holds 2 values of FP4 e2m1 numbers
-- __hip_fp4x4_e2m1: holds 4 values of FP4 e2m1 numbers
+- ``__hip_fp4x2_e2m1``: holds 2 values of FP4 e2m1 numbers
+- ``__hip_fp4x4_e2m1``: holds 4 values of FP4 e2m1 numbers
 
 FP6 (6-bit Precision)
 ========================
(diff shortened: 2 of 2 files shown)


docs/reference/low_fp_types.rst
```diff
@@ -60,11 +60,15 @@ Different GPU models support different FP4 formats. Here's a breakdown:
...
+    * - CDNA4
       - Yes
     * - RDNA2
       - No
     * - RDNA3
       - No
+    * - RDNA4
+      - No
 
 Using FP4 Numbers in HIP Programs
```
cjatin, line 73: In the C++ sample the check is for `gfx94`, it should be `gfx950`

docs/reference/low_fp_types.rst
```diff
@@ -253,6 +260,9 @@ Different GPU models support different FP6 formats. Here's a breakdown:
     * - RDNA3
       - No
       - No
+    * - RDNA4
+      - No
+      - No
 
 Using FP6 Numbers in HIP Programs
```
cjatin, line 266: same as FP4, FP6 CPP sample checks for `gfx94` it should be `gfx950`

docs/reference/low_fp_types.rst
```diff
@@ -467,12 +477,18 @@ Different GPU models support different FP8 formats. Here's a breakdown:
...
       - No
       - No
     * - RDNA3
       - No
       - No
+    * - RDNA4
+      - No
+      - Yes
 
 Using FP8 Numbers in HIP Programs
```
cjatin, line 491: the list of devices with fp8 support is larger, can we add all the devices in C++ sample so that folks who copy and paste it to test this functionality, it works flawlessly for them

docs/reference/low_fp_types.rst
```diff
@@ -825,22 +825,193 @@ against summed float values on the CPU.
...
+      - Yes
+    * - CDNA1
+      - Yes
+    * - CDNA2
+      - Yes
+    * - CDNA3
+      - Yes
+    * - CDNA4
+      - Yes
+    * - RDNA2
```
cjatin: Does RDNA2 support bf16?
the isa manual: https://www.amd.com/content/dam/amd/en/documents/radeon-tech-docs/instruction-set-architectures/rdna2-shader-instruction-set-architecture.pdf
does not have any bf16 references.
I think the question of bf16 is bit more nuanced, dont think we can capture this with a simple yes or no

cjatin (changes requested):

docs/reference/low_fp_types.rst
adeljo-amd: Is there "emulation" on the RDNA2 for bf16? We could add an asterisk to the yes and a note that it's "emulated" if that's the case

docs/reference/low_fp_types.rst
adeljo-amd: Fixed

docs/reference/low_fp_types.rst
adeljo-amd: Fixed

docs/reference/low_fp_types.rst
adeljo-amd: Added

docs/reference/low_fp_types.rst
```diff
@@ -60,11 +60,15 @@ Different GPU models support different FP4 formats. Here's a breakdown:
     * - CDNA2
       - No
     * - CDNA3
+      - No
+    * - CDNA4
```
randyh62, line 64: I wonder if we need a table to communicate all the device architectures that are not supported? It seems like a simple note could cover the host CPU and CDNA4 architecture, and leave the table for some future release when additional architectures are added?

randyh62 (approved):

docs/reference/low_fp_types.rst
adeljo-amd: Maybe, but it might also be useful for users to see explicit information about what's supported and what's not supported. The table at least sticks out compared to a note - but I'm also torn about what would be better in this case

docs/reference/low_fp_types.rst
adeljo-amd: Based on the HIP headers, there are bf16 operators that are defined, and for a lot of math operators, they seem to default to `__h*` or even convert to float before invoking `__ocml_` functions. So RDNA2 can support bf16 AFAICT

docs/reference/low_fp_types.rst
adeljo-amd: Maybe we can remove the Host part, or rename it to "emulated" (I'd prefer removing the Host part). Shall we just keep the table and remove the Host entry?

docs/reference/low_fp_types.rst
cjatin: RDNA2 can run bf16 code. but the hardware does not support it. We get by it by either promotion to float or using help of compiler. I guess the question is what information we want to pass to the user when we say `Supported devices for float16 numbers` are we considering the promotion to float mechanism?

adeljo-amd: I've added a note to the entry (which links to a note below the table) - let me know if that's better

adeljo-amd: @cjatin I actually thought about this, and I think we need to clarify the supported tables - the tables should highlight which types can be built in HIP C++ for which GPU - so technically even if RDNA2 doesn't have bf16 instruction references, users can still build HIP C++ code with it.

Would this be a better approach? We won't need to worry about whether there are intrinsics for the type since we have plans on releasing a precision support page on the ROCm page

adeljo-amd:
Closing this for now

docs/reference/low_fp_types.rst
cjatin: Can we remove the support table all together for bf16 and fp16, we can keep them for fp4/fp6/fp8. From HIP side, we do not check for any particular arch in the classes and rely on compiler/device lib support.

adeljo-amd: Done, replaced them with a short text - let me know if this is better

.wordlist.txt
```diff
@@ -12,6 +12,7 @@ Asynchrony
 asynchrony
 backtrace
 bfloat
+BFloat
```
cjatin, line 15: bfloat16, fp16, half as well?

cjatin (approved):
Looks good

.wordlist.txt
adeljo-amd: Doesn't seem necessary, the spelling CI passes

